PR #13, opened 2015-09-06
Fixed line_col failing for positions on the last line

willcrichton:
If pos was greater than the position of the last newline in the file, then the filter would return an empty set and unwrap would fail.


nikomatsakis:
Thanks!

PR #30, opened 2015-09-27
Add file time checking for windows (re)builds

dagit:
I noticed (due to long build times) that on windows the rebuild checker was always returning true.

I discovered in this thread that windows does support file modification times: https://www.reddit.com/r/rust/comments/3dhxxv/metadata_modified/

Looking in my local documentation I found the `last_write_time()` so I added a windows specific check.

### DIFF
--- a/lalrpop-snap/src/build/mod.rs
+++ b/lalrpop-snap/src/build/mod.rs
@@ -79,7 +79,16 @@ fn needs_rebuild(lalrpop_file: &Path,
         lalrpop_metadata.mtime() >= rs_metadata.mtime()
     }
 
-    #[cfg(not(unix))]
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Ah, ok, great! Thanks!

dagit:
I think I made these changes in the wrong files. It looks like I modified lalrpop-snap, and I just discovered that is a snapshot of lalrpop. Oops.

I'm working on a better version that will hopefully also fix an issue I'm having with permission denied creating the output .rs file.

nikomatsakis:
Yeah, lalrpop-snap is just a copy of an older version. Sorry, I should have noticed that!

PR #31, opened 2015-09-28
Copy brson's changes to lalrpop-snap to lalrpop and add windows times…

dagit:
Hopefully this time I made the improvements in the right part of the codebase :)

Thanks!


nikomatsakis:
Oh, I didn't notice that he had modified lalrpop-snap directly. Yeah, that's not right. Thanks :)

PR #38, opened 2015-10-30
Drop rand from required dependencies

shahn:
The rand crate pulls in a lot of dependencies which many users of
lalrpop probably wouldn't otherwise require. It is enough to require it
to run the tests, which means using lalrpop doesn't require this
anymore.

This will come to full fruition eventually when a new lalrpop snapshot
is created, until then rand will still be pulled in.


nikomatsakis:
Thanks!

PR #44, opened 2015-12-02
Do not use (..) for empty enum variants, closes #43

### LINKED ISSUE #43: Building LALRPOP fails on latest nightly
If I try to build LALRPOP on the latest nightly, the build fails rather spectacularly by first spewing out a couple thousand warnings and then hitting an internal compiler error. Changing the machine and/or the  OS doesn't affect the result. Also, both 0.7.0 and the current master branch fail in exactly the same way. Here is a log of the failing build:

[failing_build.txt](https://github.com/nikomatsakis/lalrpop/files/47513/failing_build.txt)

fhahn:
With this PR, it should be possible to compile lalrpop with the recent nightly and it should fix  #43


nikomatsakis:
@fhahn travis is still unhappy, but the problem seems to be specific to nightly, stable/beta seem good. It might just be that we need a new snapshot. You can create one by running `./lalrpop-snap`, I think, but I'm also happy to land this and do it myself.

nikomatsakis:
@fhahn oh, and thanks a lot!

shahn:
When making a new snapshot please make sure to manually update the dependencies, as they don't get synced automatically afaict. That way the rand dependency is lifted :)

fhahn:
I've created a new snapshot (with `./snap`) and on my machine the tests work. Unfortunately I have no idea why tavis still fails.

fhahn:
I've created a snapshot and it works fine for me locally, but unfortunately some tests fail now.

shahn:
I'm not sure this is an error with the snapshot generation, for me the tests start failing when compiling with stable as soon as adf4395 is applied.

nikomatsakis:
I'll take a look locally.

nikomatsakis:
Thanks again @fhahn !

fhahn:
@nikomatsakis I've investigated the issue a little bit further. The first two commits (for which the travis build passes) do not use a new snapshot. I just updated the relevant files in `lalrpop-snap` manually.

The tests fail, if a new snapshot is created. But they also fail when creating a new snapshot from master and running the tests with stable.

Should I drop the my most recent commit (the one that creates a completely new snapshot) to fix it for the current nightly for now?

nikomatsakis:
@fhahn yes let's do it...I haven't had time to investigate.

fhahn:
@nikomatsakis great, I've dropped the commit

nikomatsakis:
Thanks! I will investigate the snap failure...

PR #46, opened 2015-12-08
Start processing at the current working directory

### LINKED ISSUE #45: process_root assumes that the source code is placed in `src`
`src` is currently the [hardcoded](https://github.com/nikomatsakis/lalrpop/blob/474d2be12debe03b03224e69a7a740737c2d096a/lalrpop/src/build/mod.rs#L23) path at which processing starts, which results in a pretty ugly "No such file or directory" error without further information when this directory doesn't exist.

Could it just start in the current directory instead, or would that cause problems? (I don't think Cargo passes the path to the crate's root file to build scripts, which could provide another good starting directory)

jonas-schievink:
Fixes #45, but I'm not sure if this is the best approach

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -11,6 +11,7 @@ use rust::RustWrite;
 use tok;
 use self::filetext::FileText;
 
+use std::env::current_dir;
 use std::fs;
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


jonas-schievink:
For example, this will process `tests`, `benches`, etc. when the build script is first invoked (even if `tests` and `benches` aren't actually compiled). However, I think this shouldn't normally cause any problems.

Travis fails because of #43

nikomatsakis:
Seems reasonable to me, thanks! I'm just waiting to hear if there is some "best practice" here.

nikomatsakis:
Seems like this is the best approach. Thanks!

PR #50, opened 2015-12-23
Move to a new snapshot

nikomatsakis:


nikomatsakis:
Huh. These tests all pass locally.

nikomatsakis:
Ah, OK, I can reproduce now. Have to start from a fresh checkout. Presumably I am not really doing my version numbers right for snapshots, so cargo thinks it has an up-to-date version. In any case, the problem seems to be due to inlining misfiring. I'll try to localize it to a smaller test case.

nikomatsakis:
Blocked on https://github.com/nikomatsakis/lalrpop/issues/55

PR #51, opened 2015-12-25
Remove term dependency

shahn:


nikomatsakis:
I do plan to give more "impressively formatted" error messages at some point. :)

nikomatsakis:
But I guess until that point might as well remove the dependency.

Issue #53, opened 2016-01-15
Build failure on Windows


mAarnos:
If I try to use the version 0.8.0 of LALRPOP I get the following build failure on both a Windows 7 and a Windows 10 machine. The Rust version was 1.5 stable. I ran into #41 which isn't fixed in 0.7.0 so I am kinda blocked by this issue. 

    Compiling lalrpop v0.8.0
    failed to run custom build command for `lalrpop v0.8.0`
    Process didn't exit successfully: `C:\Users\Mikko\Desktop\lalrpop_test\target\debug\build\lalrpop29dda9a615e8fbeb\build-script-build` (exit code: 101)
    --- stderr
    thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Error
    { repr: Os { code: 5, message: "Access is denied." } }', ../src/libcore\result.rs:738

All the other parts of LALRPOP, i.e. lalrpop-util, lalrpop-snap and lalrpop-intern compile just fine.

nikomatsakis:
Hmm. I need to setup a Windows box of my own! It sounds like this might have to do with the fact that LALRPOP tries to make its intermediate `.rs` files read-only?

mAarnos:
I noticed that if I set up my TOML file like so

    [dependencies.lalrpop-util]
    git = "https://github.com/nikomatsakis/lalrpop.git"

    [build-dependencies.lalrpop]
    git = "https://github.com/nikomatsakis/lalrpop.git"

instead of 

    [dependencies.lalrpop-util]
    version = "0.8.0"

    [build-dependencies.lalrpop]
    version = "0.8.0"

then everything works fine. So it seems that the bug has already been fixed in some commit between the version in crates.io and the latest version, or there is something pretty interesting going on. I tried going through some of the commits but didn't find anything suspicious at a quick glance.

nikomatsakis:
Huh. Well, I was going to issue a new release soon anyway -- so maybe it is fixed. However, it may also just be that this triggers some sort of make clean? Unclear.

oconnor0:
I have encountered this same error - where the .rs generated from the .lalrpop is marked read-only and so gets that "Access is denied." error - on Windows 10, Rust 1.6, and LALRPOP 0.9.0.

oconnor0:
[...]

PR #66, opened 2016-01-30
Unset read only on generated rust file before removing it to fix build failure on Windows

oconnor0:
See [#53](https://github.com/nikomatsakis/lalrpop/issues/53)

After a normal `cargo build` using LALRPOP 0.9.0, the generated rust file is set as read-only. Subsequent calls to `cargo build` fail, but if I manually remove the read-only attribute on the file, the build succeeds.

This patch unsets read-only on the generated rust file before attempting to remove it and generate a new one. It still sets read-only after the file has been generated in an attempt to preserve the external semantics of the build.

This appears to consistently resolve my build issues on Windows. I have not and cannot test it on any other platforms.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -33,10 +33,11 @@ fn process_dir<P:AsRef<Path>>(root_dir: P, force_build: bool) -> io::Result<()> 
     for lalrpop_file in lalrpop_files {
         let rs_file = lalrpop_file.with_extension("rs");
         if force_build || try!(needs_rebuild(&lalrpop_file, &rs_file)) {
+            try!(make_read_only(&rs_file, false));
             try!(remove_old_file(&rs_file));
             let grammar = try!(parse_and_normalize_grammar(lalrpop_file));
             try!(emit_recursive_ascent(&rs_file, &grammar));
-            try!(make_read_only(&rs_file));
+            try!(make_read_only(&rs_file, true));
         }
     }
     Ok(())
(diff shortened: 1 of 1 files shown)


oconnor0:
I just realized this patch only works if the generated rust file already exists.

nikomatsakis:
It seems readily adaptable though -- perhaps moving that call into the `remove_old_file` function?

oconnor0:
Yes, I tweaked `make_read_only` to check for a file's existence before attempt to modify permissions - as in https://github.com/oconnor0/lalrpop/commit/b7ce5bf4ffec4cf261ce7ee439a5f04d3cc633e3

nikomatsakis:
@oconnor0 and that works for you? Do you want to re-open the PR, or should I just pull this fix into a local branch?

oconnor0:
It does work for me.

oconnor0:
@nikomatsakis Actually, I have no idea how to update this pull request so it's probably better if you pull it to a local branch.

oconnor0:
Never mind. It updated. I didn't realize that. :)

lalrpop/src/build/mod.rs
```diff
@@ -99,11 +100,15 @@ fn needs_rebuild(lalrpop_file: &Path,
     }
 }
 
-fn make_read_only(rs_file: &Path) -> io::Result<()> {
-    let rs_metadata = try!(fs::metadata(&rs_file));
-    let mut rs_permissions = rs_metadata.permissions();
-    rs_permissions.set_readonly(true);
-    fs::set_permissions(&rs_file, rs_permissions)
+fn make_read_only(rs_file: &Path, ro: bool) -> io::Result<()> {
+    if (rs_file.is_file()) {
```
nikomatsakis: Nit: unnecessary parentheses around if condition

nikomatsakis:
@oconnor0 mind removing those parens?

oconnor0:
@nikomatsakis Not at all. I haven't got used to the syntax of Rust yet. :)

nikomatsakis:
Great, thanks! I hope this fixes the problem for everyone else too.

Issue #67, opened 2016-01-31
Add operator precedence like Yacc 


TyOverby:
I'm trying to replicate part of the rust grammar as found at https://github.com/rust-lang/rust/blob/master/src/grammar/parser-lalr.y and operator precedence is vital for not having my grammar be a huge mess of layers that encode the precedence.

nikomatsakis:
So, a couple of things:

1. I'm also working on porting that grammar, and I know that @fhahn is interested too. We should coordinate. On Monday, I plan to make a dedicated repo. That said, I've been hitting some bugs or other limitations in LALRPOP, so I was planning to try and find those first. Anyway, see discussion on https://github.com/rust-lang/rust/issues/30942.

2. I am not keen on adding YACC-style precedence declartions, and in fact the Rust grammar is a good example of why. They tend to get abused to do things that are not precedence. That said, some kind of more declarative precedence solution would be good -- I was thinking that it might be that you can declare precedence on alternatives within a nonterminal, but not on general tokens. But I've not really experimented hard in this direction, since as priorities go it falls rather low on my list.

nikomatsakis:
I updated the title to remove the reference to Yacc but keep the part I think is good. :) Specifically I have in mind something like this (though now that I write it, I find it less appealing):

```rust
Expr: Expr = {
    #[precedence(rank=1, assoc=left)]
    Expr "*" Expr => ...,
    Expr "/" Expr => ...,  // precedence of (1, left) is inherited from previous nonterminal   
    #[precedence(rank=2, assoc=left)]
    Expr "+" Expr => ...
    Expr "-" Expr => ...,
};
```

These declarations would only affect resolution of conflicts between alternatives of `Expr`. In fact, I suspect we would just normalize recursive references to `Expr` away into the layers that you write now, so that this applies equally to all algorithms and the backend doesn't have to care.

nixpulvis:
I agree that YACC style declarations get abused. I like the proposed syntax above. Could we also allow for attributes like

```rust
#[precedence(rank=1, assoc=none)]
```

nikomatsakis:
I was considering something like this, where we use an `else` groups to separate the precedence ranks, but it doesn't permit specifying associations:

```
Expr: Expr = if {
    Expr "*" Expr => ...,
    Expr "/" Expr => ...,  // precedence of (1, left) is inherited from previous nonterminal   
} else {
    Expr "+" Expr => ...
    Expr "-" Expr => ...,
} else {
[...]

PR #78, opened 2016-02-19
Alternative / Production Annotations

nixpulvis:
This change is two fold, and inspired by #67. This lays the groundwork for allowing more annotations of new forms, and in new places. Regardless of the implementation of precedence rules more flexible annotations will open doors in the future.
- Change the grammar of an `Annotation` to allow key value pairs of arguments like Rust. For example `#[precedence(rank=1, assoc=left)]`.
- Add annotations to both `parse_tree::Alternative` and `repr::Production` to allow configurations on a more fine grained level.
### Open Questions
- [ ] Should we have annotations on the `repr` data structures. I think this could make sense, as it might be useful at some point. But that's not necessarily a good enough reason to do it now.
- [x] Can and should macros be able to include annotations, and if so how to do that.
- [ ] Tests

### DIFF
--- a/lalrpop/src/grammar/parse_tree.rs
+++ b/lalrpop/src/grammar/parse_tree.rs
@@ -153,6 +153,8 @@ pub struct NonterminalData {
 pub struct Annotation {
     pub id_span: Span,
     pub id: InternedString,
+    // TODO nixpulvis: This is not exactly the correct type.
+    pub parameters: Vec<(InternedString, InternedString)>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
--- a/lalrpop/src/grammar/parse_tree.rs
+++ b/lalrpop/src/grammar/parse_tree.rs
@@ -153,8 +153,14 @@ pub struct NonterminalData {
 pub struct Annotation {
     pub id_span: Span,
     pub id: InternedString,
-    // TODO nixpulvis: This is not exactly the correct type.
-    pub parameters: Vec<(InternedString, InternedString)>,
+    pub parameters: Vec<AnnotationKeyValue>,
+}
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AnnotationKeyValue {
+    pub span: Span,
+    pub key: InternedString,
+    pub value: InternedString,  // TODO: Still not sure of this type.
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
--- a/lalrpop/src/grammar/parse_tree.rs
+++ b/lalrpop/src/grammar/parse_tree.rs
@@ -149,7 +149,7 @@ pub struct NonterminalData {
     pub alternatives: Vec<Alternative>
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Annotation {
     pub id_span: Span,
     pub id: InternedString,
@@ -160,6 +160,8 @@ pub struct Alternative {
     pub span: Span,
 
     pub expr: ExprSymbol,
+
+    pub annotations: Vec<Annotation>,
 
     // if C, only legal in macros
     pub condition: Option<Condition>,
--- a/lalrpop/src/parser/lrgrammar.lalrpop
+++ b/lalrpop/src/parser/lrgrammar.lalrpop
@@ -83,7 +83,7 @@ Annotation: Annotation = {
             parameters: vec![],
         }
     },
-    "#" "[" <lo:@L> <id:Id> "(" <l:Comma<EqPair>> ")" <hi:@R> "]" => {
+    "#" "[" <lo:@L> <id:Id> "(" <l:Comma<AnnotationKeyValue>> ")" <hi:@R> "]" => {
         Annotation {
             id_span: Span(lo, hi),
             id: id,
@@ -92,8 +92,15 @@ Annotation: Annotation = {
     },
 };
 
-EqPair: (InternedString, InternedString) =
-    <l:Id> "=" <r:Id> => (l, r);
+AnnotationKeyValue: AnnotationKeyValue = {
+    <lo:@L> <key:Id> "=" <value:Id> <hi:@R> => {
+        AnnotationKeyValue {
+            span: Span(lo, hi),
+            key: key,
+            value: value,
+        }
+    }
+};
 
 NonterminalName: (NonterminalString, Vec<NonterminalString>) = {
     <MacroId> "<" <Comma<NotMacroId>> ">",
--- a/lalrpop/src/parser/lrgrammar.lalrpop
+++ b/lalrpop/src/parser/lrgrammar.lalrpop
@@ -96,6 +96,7 @@ Alternative: Alternative = {
         Alternative {
             span: Span(lo, hi),
             expr: ExprSymbol { symbols: s },
+            annotations: vec![],
             condition: c,
             action: a
         }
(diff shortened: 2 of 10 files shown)


nikomatsakis:
I think I've decided I prefer a more explicit syntax for precedence rules vs the annotations I initially proposed, but I agree this is useful groundwork that will no doubt come in handy at some point. =) I'll try to read it carefully soon!

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -149,10 +149,12 @@ pub struct NonterminalData {
     pub alternatives: Vec<Alternative>
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Annotation {
     pub id_span: Span,
     pub id: InternedString,
+    // TODO nixpulvis: This is not exactly the correct type.
+    pub parameters: Vec<(InternedString, InternedString)>,
```
nikomatsakis: Let's add a new struct for this:

```
struct AnnotationKeyValue {
    span: Span,
    key: InternedString,
    value: InternedString,
}
```

Also, we're going to have to check now that annotations like `#[inline]` fit the right form. I wonder if it'd be worth just hard-coding that stuff into the parser, rather than checking it programmatically in `validate`? (We accept so few annotations now.)

Otherwise, we should somehow make the checking more declarative. Heh, I suppose we could make a second grammar just for annotations and then write a custom tokenizer for them that emits tokens (or just parse their Debug output). That'd be...kind of nifty actually.

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -75,10 +75,25 @@ Nonterminal: GrammarItem =
...
+    "#" "[" <lo:@L> <id:Id> "(" <l:Comma<EqPair>> ")" <hi:@R> "]" => {
+        Annotation {
+            id_span: Span(lo, hi),
+            id: id,
+            parameters: l,
+        }
+    },
+};
+
+EqPair: (InternedString, InternedString) =
```
nikomatsakis: Once this type has a name, use it here too.

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -92,18 +107,20 @@ Alternatives: Vec<Alternative> = {
 };
 
 Alternative: Alternative = {
-    <lo:@L> <s:Symbol+> <c:("if" <Cond>)?> <a:Action?> <hi:@R> => {
+    <lo:@L> <n:Annotation*> <s:Symbol+> <c:("if" <Cond>)?> <a:Action?> <hi:@R> => {
```
nikomatsakis: Pre-existing, but now that inlining works in the snapshot we should be able to remove the second arm here and change `Symbol+` into `Symbol*`.

nikomatsakis:
OK, so this looks pretty good modulo a few minor points (we could push off the validation until a later PR, though I'm into the idea of using a grammar to do it right now:). The biggest problem is lack of tests (which you also cited).

Also, we should probably add an "unstable" feature and reject `precedence` annotations unless that feature is in use. (I can help with this if you want.)

But onto your open questions:

> - Should we have annotations on the `repr` data structures. I think this could make sense, as it might be useful at some point. But that's not necessarily a good enough reason to do it now.

Easy enough to do whenever we want to do it, I wouldn't fret.

> - Can and should macros be able to include annotations, and if so how to do that.

I don't know what this means, can you give an example? It seems like macro *bodies* in your example *can* have annotations, and they get expanded.

> - Tests?

Yes, we need tests. I try to have LALRPOP really abide by unit testing, so there are tests scattered throughout targeting each submodule. That said, I don't currently seem to have any *parser* tests, I think because all of the other tests employ the parser, and since it's derived from a grammar I figured the tests wouldn't add that much. 

I think the following two tests to start:

- `lalrpop/normalize/prevalidate/test.rs` -- check for invalid/duplicate annotations, following the model of the existing tests like `unrecognized_annotation`; this will also implicitly check the parser
- maybe a test in `lalrpop/normalize/macro_expand/test.rs` checking that annotations are copied through to the newly expanded result

lalrpop/src/grammar/parse_tree.rs
nixpulvis: I'm all for a grammar for annotations, I'm not exactly sure what you mean by parsing the Debug output but it sounds hacky.

nixpulvis: I'm also not sure `InternedString` is the right choice anymore, as things like `1` used in the precedence example don't get parsed. What kinds of things would we expect for for the `value` fields of this structure.

nixpulvis:
Rebased.

nikomatsakis:
@nixpulvis 

> I'm all for a grammar for annotations, I'm not exactly sure what you mean by parsing the Debug output but it sounds hacky.

I meant parsing the result of `format!("{:?}")`. It is mildly hacky, but then that does have a declared structure. :) The alternatives would be:

0. Build all the known attributes directly into LALRPOP's grammar rather than trying to handle them generically as we do now. 
1. Write a custom tokenizer that, given a "raw" attribute, can be "serialized" into a set of tokens like `START_ATTRIBUTE` / `KEY_VALUE_PAIR` etc and then write a parser for those tokens.
2. Write a bunch of custom code to validate by hand.

I am not keen on option 2 -- this is what we (try) to do in rustc and we always fail to check for everything, meaning we permit annotations with extra key/values or just don't validate at all etc etc. This is especially true if annotations start to have a very generic structure (which I would probably support), like `#[foo(bar=baz(...))]`. Also, that code itself is just annoying to write.

I am fine with option 1. (And I can try to explain it a bit more explicitly if what I wrote is not clear.) But I do wonder if, given that we only plan to support a limit set of attributes, it's not easier to just go with option 0.

nixpulvis:
I was personally leanin towards option 1 for now, seeing as this is a parse project I figured using the parser to get annotations would be ideal. However maybe I'm still not understanding the Debug option. To me it sounds like your saying save the annotation as a String and parse it manually in Rust code.

nikomatsakis:
Sorry, my set of alternatives got renumbered by markdown. My preference for now is to build the desired attributes directly into lalrpop's grammar. I would probably go so far as to change the data structures we use to be specific to particular attributes. E.g. instead of a generic annotation, we might have 'enum Annotation { InlineAnnotation, ...

nikomatsakis:
Sorry, closed by accident. Phone screens are small.:)

nixpulvis:
Agreed, I can add that in.

On a related but tangential note, I'm trying to hack this library to allow me to get this parser I'm writing for class to parse the Tiger grammar (Appel's) compiler textbook. Do you have any recommendations for adding the shift reduce hack ala yacc to the lalr state generator? I'm currently just commenting out the check for conflicts lol.

nikomatsakis:
@nixpulvis 

>  Do you have any recommendations for adding the shift reduce hack ala yacc to the lalr state generator? 

To be clear, I remain very skeptical about actually landing such a change, however if you wanted to do it, you would modify [this code](https://github.com/nikomatsakis/lalrpop/blob/729e61d4efebbe173df8f6c56e9f562eebd093a0/lalrpop/src/lr1/build/mod.rs#L100-L112) to take no action if there is an existing shift -- you probably still want to error out if there is an existing reduce, as that indicates a reduce/reduce conflict.

nixpulvis:
@nikomatsakis Thanks, yea defiantly not saying this is the *solution* just wanted to play around with it.

nikomatsakis:
Going to close this. Been a while. Perhaps you and I should discuss whether to go forward with this or what.

nixpulvis:
Yea, sounded like there were some options for this still, and i havnt had time. Seems like something along these lines will be needed though, unless there is another tool in LALRPOP that makes it easier now.

PR #80, opened 2016-02-19
Buffer output and only write to .rs file if successful

### LINKED ISSUE #57: Partial generated .rs file prevents rebuild
I have a `.lalrpop` file in my project that results in an shift-reduce-conflict. When building, lalrpop partially generates the `.rs` file which looks like this:

```rust
#![allow(unused_imports)]
#![allow(unused_variables)]
// here are my imports from the lalrpop file
extern crate lalrpop_util as __lalrpop_util;
use self::__lalrpop_util::ParseError as __ParseError;
```

This file then prevents lalrpop from rebuilding the `lalrpop` file if it wasn't modified. When just executing `cargo build` again, `rustc` says that the function `parse_GoalSymbol` does not exist. It's even worse when you are just testing, if some lalrpop file is valid (without using the `pub` symbols in your rust code). Then the second `cargo build` works without failure, which is super confusing.

I think the best solution is to not write anything to the `.rs` file until everything was processed correctly.

nikomatsakis:
Fixes #57 

Also a few other minor things.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -7,6 +7,9 @@ scenarios and suggesting solutions.
 
 Added a standalone LALRPOP executable.
 
+We no longer generate incomplete files when grammar generation fails
+(Issue #57).
+
 # Version 0.9
 
 Miscellaneous bug fixes, mostly. Processing for a `build.rs` file now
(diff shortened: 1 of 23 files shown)


nikomatsakis:
cc @LukasKalbertodt

PR #85, opened 2016-02-21
Add support for generating files into Cargo OUT_DIR

dflemstr:
When I started using this library, I was very surprised when it generated source files next to the `.lalrpop` files, so that the generated files were checked into source code management.  I would expect any tools that generate source code to behave like [the example in the Cargo documentation](http://doc.crates.io/build-script.html#case-study-code-generation) and put them in the Cargo `OUT_DIR` directory.

This PR tries to apply the principle of least astonishment and adds a `.process()` function that behaves like you would expect; a file `src/foo/bar.lalrpop` is processed into `$OUT_DIR/foo/bar.rs`.  It also adds a function called `process_file_to` that lets the user decide where to put the generated file.

I wanted to get this to work for one of my own projects so the code might not be perfect or matching this project's style, so feel free to criticize at leisure!

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -18,6 +18,7 @@ use tls::Tls;
 use tok;
 use walkdir;
 
+use std::env;
 use std::env::current_dir;
 use std::fs;
 use std::io::{self, Write};
@@ -37,7 +38,7 @@ pub fn process() -> io::Result<()> {
 
     // The environment variable OUT_DIR is set by cargo, and specifies
     // a directory where generated code should be put.
-    process_dir(&session, &src_dir, Path::new(env!("OUT_DIR")))
+    process_dir(&session, &src_dir, Path::new(&env::var("OUT_DIR").unwrap()))
 }
 
 pub fn process_root() -> io::Result<()> {
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -16,6 +16,7 @@ use session::{ColorConfig, Session};
 use term;
 use tls::Tls;
 use tok;
+use walkdir;
 
 use std::env::current_dir;
 use std::fs;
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Heh, to be honest I never read that document. :) I agree this seems like a clearly better approach. (And is what I wanted but didn't take time to investigate.) Will review shortly!

dflemstr:
And now I broke it, trying to fix a bug with the GitHub editor :)

lalrpop/src/build/mod.rs
```diff
@@ -29,32 +31,56 @@ mod fake_term;
...
+    // The environment variable OUT_DIR is set by cargo, and specifies
+    // a directory where generated code should be put.
+    process_dir(&session, &src_dir, Path::new(&env::var("OUT_DIR").unwrap()))
+}
+
 pub fn process_root() -> io::Result<()> {
     let session = Session::new();
-    process_dir(&session, try!(current_dir()))
+    let cwd = try!(current_dir());
+    process_dir(&session, &cwd, &cwd)
```
nikomatsakis: Hmm, I feel like we should still put stuff into the OUT_DIR, or at least we should do so if it is located under `src`. Eventually I guess we might want a configuration option here.

dflemstr: Maybe; I just kept it like this for backwards compatibility.  There is the `.process()` function that does what you expect.

Note that putting things in `OUT_DIR` is *not* equivalent to putting it into the `src` dir.  You still need to `include!()` the generated file from a file that's checked in.  That's why changing `process_root()` to write its generated files to `OUT_DIR` would be a breaking change.

nikomatsakis:
Yeah so I was reading this document and now I realize that `rustc` won't find the relevant files in the `OUT_DIR` automatically.

I think what we ought to do is something like this:

1. Modify the `Settings` code to allow user to specify an out directory. The default should be the `OUT_DIR` from the environment, but it should also be possible to set things up to write in-place.
2. Add a macro used like `lalrpop! { mod foo; }` (or `lalrpop! { pub mod foo; }`) that will create the `include!` incantation (example below). We can easily make this module also allow user to specify a path if we want.

Possible macro definition might look something like this:

```
macro_rules! lalrpop {
    (mod $name:ident) => {
        mod $name {
            include!(concat!(env!("OUT_DIR"), "/", stringify!($name), ".rs"))
        }
    }
}
```

but we'd want to expand it to handle `pub` and other attributes on the `mod` declaration, and we'd probably want some syntax like `mod foo from "path/in/out_dir.rs", to allow for LALRPOP files whose names do not precisely match the `mod`.

This way, the recommended pattern becomes:

```rust
lalrpop::process();
```

and in the rs file writing:

```rust
lalrpop! { mod foo; }
```

Pretty tolerable.

dflemstr:
Note that the `include!()` macro currently does not work as expected, because of a bug (https://github.com/rust-lang/rfcs/issues/752).  I'm using [`mod_path`](https://crates.io/crates/mod_path) to work around that issue, which I can do because I target the nightly compiler.

All I'm saying is that `lalrpop` probably should stop emitting top-level inner attributes for users of this behavior to stay sane.

nikomatsakis:
Hmm. If it's just a matter of top-level inner attributes, that seems achievable.

nikomatsakis:
@dflemstr would you be up for rebasing this on top of the recent changes made in https://github.com/nikomatsakis/lalrpop/pull/102? In particular, the idea would be to remove the new top-level function and instead add a flag to the `Configuration` (and, hence, the `Session`). I would suggest adding a field to `Session` like `out_dir: Option<PathBuf>`, which defaults to `None`. The idea would be that `None` means "same directory as the lalrpop file". Then we can add two methods to `Configuration`:

```rust
fn set_out_dir<P:Into<PathBuf>>(&mut self, dir: P) -> &mut Self {
    self.session.out_dir = Some(dir.into());
    self
}

fn set_out_dir_from_env(&mut self) -> &mut Self {
    self.set_out_dir(env!("OUT_DIR"))
}
```

This seems like a good starting point. In some later PR I may try to add this `lalrpop_mod!` macro and transition to using the out directory by default.

dflemstr:
@nikomatsakis my day job means I haven't had time to look at this more :) But I've put it on my backlog to do the rebase

nikomatsakis:
@dflemstr no worries, if you'd prefer, I can do it. I don't imagine it would take all that long.

nikomatsakis:
(At the same time, my LALRPOP quota is also somewhat limited, so we could just say that whoever gets to it first will do it :)

dflemstr:
Will make new PR

PR #86, opened 2016-02-21
Remove unused module `intern` from lalrpop/src

fhahn:
This PR removes two unused source files from `lalrpop/src`


nikomatsakis:
Remnants from when I moved them into their own crate, I guess.

PR #87, opened 2016-02-22
Convert to use the `regex-syntax` crate

### LINKED ISSUE #29: use regex-syntax?
Neat project! I'm not sure if you knew this or not, but in the last few months, the `regex` crate grew a sub-crate called `regex-syntax` which exposes a full blown, well tested, regex parser/AST tree: https://doc.rust-lang.org/regex/regex_syntax/index.html

I briefly looked over yours, and it looks like it supports everything except negative matches. Not sure if that's a deal breaker, but thought I drop a note anyway!

nikomatsakis:
~~This is not quite complete, in that if people stray from the subset of regex that we currently support, then the code panics. (In particular, we don't support boundaries.) But it's quite close to being ready.~~ This should have the nice benefit of giving us access to better unicode support, case-insensitive literals, and a few other niceties (and, no doubt, fewer bugs).

Fixes #29.

cc @BurntSushi


BurntSushi:
@nikomatsakis Yay! LGTM. :-)

nikomatsakis:
@BurntSushi nice. I think I fixed up the last little bits this morning.

Issue #65, opened 2016-01-29
larger grammars generate a LOT of code


timglabisch:
Hello,

i try to parse php expressions using lalrpop:
https://github.com/timglabisch/rustphp/blob/c79060d6495a55174fc2ad5710d5774d8ec94d67/src/calculator1.lalrpop

my problem is that cargo run becomes incredible slow:

time cargo run 
1234.76s user 19.88s system 99% cpu 20:58.78 total

the file is very huge (~50mb): 

cat src/calculator1.rs | wc -l
1044621

is there something fundamentally wrong or is this expected?

nikomatsakis:
To be clear, is your concern *compilation time* (i.e., `cargo build`) or *execution time* (i.e., time to parse a PHP file). I've definitely noticed that the recursive ascent algorithm that LALRPOP uses generates quite a lot of code, which then correspondingly takes quite a lot of time to compile. You can help things by switching to the LALR(1) algorithm:

```
grammar["LALR(1)"];
```

I have been wanting to implement some other improvements as well (e.g., [generalized left corner parsing](http://dl.acm.org/citation.cfm?id=512966) or other algorithms beyond LALR(1)). Another option is to add a second table-driven mode, but that will cause the parser itself to run more slowly.

timglabisch:
sorry for not describing this clear, my concern is just `compilation time`.

nikomatsakis:
@timglabisch ok, then yes it's known -- I'm hoping to put in some effort to fix it in the near term, since I want to get a LALRPOP version of the Rust grammar up and going too (and it's going to have the same problem).

nikomatsakis:
I modified the issue title to be more precise.

sharpjs:
This was a sore point for me as well when I tried LALRPOP.  Even [my small grammar](https://github.com/sharpjs/rust-learning/blob/5155b57192eb35c1a4002959181c3584db93b4bc/src/parser.lalrpop) (C-ish expressions) produced a very large .rs file, mostly comments, that took over a minute to compile, IIRC. At the time, it seemed to me that the output size was increasing exponentially with respect to the grammar size, so I discontinued use of LALRPOP for the time being.

I'd like to see this addressed if possible, because I'm one of those crazy people who feel more comfortable with LALR(1) than LL(1).  Also, I like LALRPOP's grammar syntax and features.

nikomatsakis:
[...]

PR #89, opened 2016-02-22
table drive parser [WIP]

fhahn:
This is a work-in-progress PR for  #65. So far it does not do too much, but it already generates const arrays for interned strings and "reduced" items. 

The main intention behind this PR is that @nikomatsakis gets a rough idea how my current approach looks like and to discuss it.


nikomatsakis:
Very exciting! I'll take a look soon.

nikomatsakis:
OK, took a quick look. Maybe I'm just not thinking about this very hard, but why do we need to serialize the interned strings etc? The tradition action table looks something like this <https://en.wikipedia.org/wiki/LR_parser#Action_and_goto_tables>, right?

So I figure we need to:

1. Number the nonterminals (N) and terminals (M)
2. Make an action table with one row per state that contains a column for each nonterminal and an extra column for EOF (`$`). The cells of that table are shift/reduce actions, which could probably be represented as `i32`, where negative numbers indicate a reduction and positive numbers indicate a shift.
3. A goto table that contains a column for each terminal and a row for each state. The values of that are a `u32` indicating the next state.
4. To store the data, an enum containing a variant per terminal and nonterminal with the associated data (the recursive ascent code already [generates a similar enum][enum], except that it only contains nonterminals).
5. The stack is then a stack of u32 (state indices) and a stack of those enums. The length of the state stack is always one more than the enum.
6. The inner loop is roughly the same as how the interpreter works, except that we pull off each token, then apply the patterns to figure out what kind of token we have and convert into a token index and associated value. [This is the place][2] where the recursive ascent compiler matches the tokens to figure out what the next symbol is and extra its values.

[enum]: https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/lr1/ascent.rs#L134-L152
[2]: https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/lr1/ascent.rs#L509-L513

nikomatsakis:
@fhahn what do you think of that?

fhahn:
@nikomatsakis sounds good, thanks for taking the time to write that up! The first commits I pushed were more the result of just getting familiar with the lalrpop stuff.

I've now pushed a commit which generates a yet incomplete action table (reduce info missing so far) and the goto table. 

I've now pushed a few more commits that emit a goto and action table. I also started working on the table based parser. So far I've moved the code to lalrpop-utils, but we could also emit the code directly in the resulting .rs file later. The next thing I have to do is tokenization of the input, but I'll first have to check what the recurisve parser does there.

fhahn:
Another thing came to my mind. You initially mentioned that we probably need an array to store the terminals/nonterminals. But so far it seems enough to just create an array with the productions.

nikomatsakis:
I'll take a look at your commits in a bit but --

I suspect we will want to generate the code in the rs file. I don't think there's much logic we can share between parsers really -- maybe a little bit. But we have to generate the matching code from the patterns the user gave to identify the next terminal, and that's just a big chunk of code. That we might abstract by passing in an `Iterator<Item=u32>` (where the `u32` is the terminal id) but then we also have to dispatch to the action code, which will require popping (and downcasting) from the data stack, which is also not readily abstractable.

nikomatsakis:
@fhahn left a few more comments! Seems like you're on the right track, though I definitely think we should move away from strings and complex types and towards integers, which will be faster at runtime.

fhahn:
Yeah, I now have a better idea about what's going on with the lexer stuff. I guess in the lexer generated by lalrpop, we could just return indices to the action able for terminals (not sure if this is already done at the moment). For hand-written lexers, I am not sure. There it seems common to just use an enum for the token types. In this case it seems like using a map of tokens -> action column is necessary.

nikomatsakis:
@fhahn the code at this level of abstraction doesn't really care whether we are using the auto-generated lexer or not. It's all been normalized to matching a token of type T against a set of patterns that match against values of type T. Basically, at this level of abstraction, you just get an `Iterator<(L,T,L)>`, where `L` is the location type and `T` is the token type. 

That said, in the case of an auto-generated lexer, the token type will be `(usize, &'input str)`, and the match patterns look like `(22, input)` where `22` is the integer assigned (by the tokenizer) to a particular token. I guess it'd be more efficient, in the particular of using the auto-tokenizer, to use the same integers, and thus match `(n, input)`-- but that seems like an optimization.

Initially, I would expect the parser generator to just generate an index for each terminal in `grammar.all_terminals` and then do something like:

```
let terminal_index = match lookahead {
    Some((_, $Pattern0, _)) => 0,
    Some((_, $Pattern1, _)) => 1, 
    Some((_, $Pattern2, _)) => 2,
    ....
    None => N
}
```

fhahn:
@nikomatsakis  thanks! A function mapping token types to those indexes should be generated already. Will work on the actual parser some more today.

lalrpop/src/build/mod.rs
```diff
@@ -287,6 +287,54 @@ fn emit_uses<W:Write>(grammar: &r::Grammar,
...
+        let states = match lr1::build_states(&grammar, start_nt) {
+            Ok(states) => states,
+            Err(error) => {
+                let messages = lr1::report_error(&grammar, &error);
+                let _ = report_messages(messages);
+                exit(1) // FIXME -- propagate up instead of calling `exit`
+            }
+        };
+
+        try!(lr1::interpreter::compile(&grammar, user_nt, start_nt, &states, &mut rust));
```
nikomatsakis: Ideally I think we would integrate this with `emit_recursive_ascent`, so that it's just something like `emit_parser` and we match here, since otherwise the code ought to be the same between the two cases I think

lalrpop/src/lr1/interpreter.rs
```diff
@@ -0,0 +1,273 @@
...
+    fn write_action_table(&mut self) -> io::Result<()> {
+        let row_len = self.grammar.all_terminals.len()+1;
+
+        let productions = self.grammar.nonterminals.values().flat_map(|nt| nt.productions.iter()).cloned().collect::<Vec<_>>();
+        let production_bits: Map<_, _> = productions.iter().cloned().zip(0..).collect();
+
+        rust!(self.out, "const productions: [ReducedProduction; {}] = [", productions.len());
+
+        for p in &productions {
+            rust!(self.out, "    ReducedProduction {{ nonterminal: {}, symbol_count: {} }},", *self.nonterminal_bits.get(&p.nonterminal).expect("got nonexisting nonterminal"), p.symbols.len());
```
nikomatsakis: Nit: mega long line :)

lalrpop/src/lr1/interpreter.rs
```diff
@@ -0,0 +1,273 @@
...
+        }
+        rust!(self.out, "];");
+
+        let rows = try!(self.states.iter().enumerate().map(|(i, s)| {
+            let mut row = iter::repeat("0".to_owned()).take(row_len).collect::<Vec<_>>();
+            for (lookahead, action) in &s.tokens {
+                let target = match *action {
+                    Action::Shift(index) => format!("{}", (index.0 as i32) + 1),
+                    Action::Reduce(ref prod) => format!("-{}", production_bits.get(prod).expect("got nonexisting production")),
```
nikomatsakis: Nit: I personally prefer to use `[]` than `get/expect`. The latter doesn't give me any more information than a backtrace -- and, in fact, if we ever improve the error message for `[]` (which we can, with specialization!), would give me *less* since it wouldn't tell me the value of the key.

lalrpop/src/lr1/interpreter.rs
nikomatsakis: Nit: as belong, I prefer `[]` over `get/expect`

lalrpop/src/lr1/interpreter.rs
```diff
@@ -0,0 +1,273 @@
...
+                };
+
+                let column_index = match *lookahead {
```
nikomatsakis: The `TokenSet` code uses the same logic; we should extract this to a helper in `Grammar` I think

lalrpop/src/lr1/interpreter.rs
```diff
@@ -0,0 +1,273 @@
...
+            rust!(self.out, "let mut {}tokens = super::{}intern_token::{}Matcher::new(input);",
+                  self.prefix, self.prefix, self.prefix);
+        } else {
+            // otherwise, convert one from the `IntoIterator`
+            // supplied, using the `ToTriple` trait which inserts
+            // errors/locations etc if none are given
+            rust!(self.out, "let {}tokens = {}tokens.into_iter();", self.prefix, self.prefix);
+            rust!(self.out, "let mut {}tokens = {}tokens.map(|t| {}ToTriple::to_triple(t));",
+                  self.prefix, self.prefix, self.prefix);
+        }
```
nikomatsakis: I wonder if we can/should abstract this out; probably yes, but also probably not worth it just now.

lalrpop/src/lr1/interpreter.rs
```diff
@@ -0,0 +1,273 @@
...
+            "terminal_to_index".to_owned(),
+            vec![],
+            vec![format!("token: {}", self.grammar.types.terminal_token_type())],
+            "usize".to_owned(),
+            vec![]));
+        rust!(self.out, "{{");
+        rust!(self.out, "match token {{");
+
+        for (&k, v) in &self.grammar.terminal_bits {
+            // TODO same as in ascent.rs
```
nikomatsakis: the ascent.rs code has the notion of `match_terminal_pattern`, which is a pattern that just decides what token you are looking at, and `consume_terminal`, which also extracts out the data. The latter includes bindings. It seems like you are emitting a "consume" pattern here but not using the `pattern_names` anywhere. Nonetheless, I suspect that consuming is right: you probably want to match the lookahead and produce `(index, value)` where `value` is some variant of the "data stack" enum corresponding to this kind of terminal (and hence containing the data for the terminal).

nikomatsakis:
@fhahn just read through the code, this is indeed looking good. Seems like it's getting close!

nikomatsakis:
@fhahn just checking in...how are things going? 

As an aside, my branch that changes the LR(1) algorithm (which has gotten somewhat stalled...) refactors (in a small way) many of the LR(1) data structures here, but it shouldn't be so hard to rebase your work over top of that.

fhahn:
Unfortunately I did not have much time to work on this over the last month. But I'll try to get some work done in the coming week.

nikomatsakis:
@fhahn no worries -- I haven't had much time for LALRPOP lately either. If you do wind up feeling stalled, let me know, and I can try to pick up this project. I'd love to see it progress! :)

nikomatsakis:
Given the lack of action here, I wound up trying my hand at it. See #122. I'm going to close this PR in favor of that one, I think, since that one is fully functional. Thanks @fhahn for pushing on it though!

PR #95, opened 2016-02-25
Locations without conflicts

### LINKED ISSUE #90: Improve error report about @L @R tokens
I just confused myself by creating a stupid grammar that would read two `@L` tokens successively. A minimal example:

```rust
grammar;

Spanned<T> = @L <T> @R;
Word: String = <s:r"[a-z]+"> => <>;

pub Foo = {
    Spanned<Word> "~",
    Spanned<Spanned<Word>> "+",
};
```

When having many macros like `Spanned`, it can easily happen to next two `Spanned` macros. The grammar above results in:

```
test.lalrpop:3:14: 3:15: Local ambiguity detected

  The problem arises after having observed the following symbols in the input:
    @L
  At that point, if the next token is a `r#"[a-z]+"#`, then the parser can
  proceed in two different ways.

  First, the parser could execute the production at
  src/grammar/test.lalrpop:3:14: 3:15, which would consume the top 0 token(s)
  from the stack and produce a `@L`. This might then yield a parse tree like
    ╷      ╷ Word  @R
    ├─@L───┘        │
    └─Spanned<Word>─┘

  Alternatively, the parser could shift the `r#"[a-z]+"#` token and later use it
  to construct a `Word`. This might then yield a parse tree like
    @L r#"[a-z]+"# @R
    │  └─Word────┘  │
    └─Spanned<Word>─┘

  See the LALRPOP manual for advice on making your grammar LR(1).
```

This is not really helpful right now (my real error message with my big grammar was even more confusing...). 

(the new error reporting is awesome though :3)

nikomatsakis:
Revamp our location handling so that we are always threading location info alongside every nonterminal. This allows `@L` and `@R` actions (which just return lookahead information) to be inlined, which in turn means that you don't get any conflicts related to them. Basically this means you put `@L` and `@R` anywhere and things should just work fine.

I also did some internal refactoring which I think makes things a bit nicer overall:

- don't thread lookbehind information; we can now grab it from the top of the stack
- don't pass lookbehind/lookahead to actions if they also have symbol arguments, they can compute it themselves now

Fixes #90.


nikomatsakis:
cc @sfackler @LukasKalbertodt

LukasKalbertodt:
Oh my god, my grammar is just working now. Huge thanks! :heart:

nikomatsakis:
@LukasKalbertodt huzzah!

sfackler:
Awesome! Do you know when you're planning to push a new release to crates.io?

nikomatsakis:
@sfackler guess now is as good a time as any :)

sfackler:
Thanks!

PR #106, opened 2016-04-02
Fix bug in USAGE and add regression tests.

Nemikolh:
Fix regression introduced by d70cfac7.

### DIFF
--- a/lalrpop/src/main.rs
+++ b/lalrpop/src/main.rs
@@ -61,7 +61,7 @@ fn main1() -> io::Result<()> {
 }
 
 const USAGE: &'static str = "
-Usage: lalrpop [options] inputs...
+Usage: lalrpop [options] <inputs>...
        lalrpop --help
 
 Convert each of the given inputs (which should be a `.lalrpop` file)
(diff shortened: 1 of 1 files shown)


Nemikolh:
The `stable` and `beta` jobs have passed successfully, but the `nightly` one failed with:

```
error: internal compiler error: unexpected panic
note: the compiler unexpectedly panicked. this is a bug.
note: we would appreciate a bug report: 
      https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports
note: run with `RUST_BACKTRACE=1` for a backtrace

thread 'rustc' panicked at 'assertion failed: 
   declare::get_defined_value(ccx, &symbol).is_none()', 
   ../src/librustc_trans/monomorphize.rs:94

note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
While this is a problem with `rustc` itself, this doesn't seems to be correlated with those changes.
The error appears in the crate `lalrpop-test` when executing `cargo test`.
I hope this helps. :)

nikomatsakis:
@Nemikolh yeah that is a known bug in rustc, hopefully fixed soon.

nikomatsakis:
thanks!

PR #108, opened 2016-04-17
Input/output directory control

dflemstr:
A follow-up of #85

> When I started using this library, I was very surprised when it generated source files next to the `.lalrpop` files, so that the generated files were checked into source code management.  I would expect any tools that generate source code to behave like [the example in the Cargo documentation](http://doc.crates.io/build-script.html#case-study-code-generation) and put them in the Cargo `OUT_DIR` directory.

This PR adds API calls that allows for `build.rs` code like:

```rust
fn main() {
    lalrpop::Configuration::new()
        .use_cargo_dir_conventions()
        .process().unwrap();
}
```

...or

```rust
fn main() {
    lalrpop::Configuration::new()
        .set_in_dir("src")
        .set_out_dir("gen")
        .process().unwrap();
}
```

### DIFF
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -2,9 +2,10 @@ use build;
 use log::Level;
 use session::{ColorConfig, Session};
 use std::default::Default;
+use std::env;
 use std::env::current_dir;
 use std::error::Error;
-use std::path::Path;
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Looks great, thanks!

PR #118, opened 2016-07-06
Don't emit module-level attributes

dflemstr:
This makes it so that `lalrpop` doesn't emit top-level module-level attributes. It does so by:

  * Moving the `#[allow(unused_variables)]` to each use-site.
  * Using fully qualified paths instead of `#[allow(unused_imports)]`.

This has the additional advantage that the user will get warnings for unused manual imports.

The main purpose of this pull request is however to allow generated parser files to be `include!()`d.


nikomatsakis:
Seems good, thanks.

I haven't tested this, but using this branch, do we not see any warnings?

nikomatsakis:
Sorry for the slow response btw!

dflemstr:
No warnings last time I checked!

nikomatsakis:
@dflemstr great, thanks.

PR #123, opened 2016-08-04
upgrade snapshot to use table-driven parser

nikomatsakis:
Always good to dogfood. Ought to build faster too.


brendanzab:
Have you thought of splitting the snapshot into a separate repo? Might reduce churn on the git history.

Issue #39, opened 2015-11-05
Document custom lexers


nikomatsakis:
The current tutorial and documentation does not explain how one writes and integrates a custom lexer.

mAarnos:
I am trying to build a parser for the BNF described [here](http://www.cs.miami.edu/~tptp/TPTP/SyntaxBNF.html). If you scroll down enough you notice that the BNF also contains a custom lexer. However, without documentation it is pretty hard to implement it. It would be greatly appreciated if you could finish up the documentation, or alternatively help me in the implementation work.

nikomatsakis:
@mAarnos I should really write up the docs, yes. But in the meantime: do you have a link to what you have attempted? I can show you some examples -- it's actually fairly straightforward.

Basically, you just need to define an enum for your tokens and write up something that implements the `Iterator` trait for those tokens. It depends a bit if you want accurate span (line/column number) information and whether your iterator is fallible (i.e., if there are inputs that cannot be lexed, which is generally true). 

This is the [LALRPOP tokenizer][1], which makes maximal use of all features. [This enum][2] represents each token. Here is [the implementation of the iterator trait][3]. The item type is `Result<Spanned<Tok<'input>>, Error>`: basically, the `Err` case means a lexer error, and the `Ok` case means a successful token. `Spanned` is just an alias for `(Location, Tok<'input>, Location)`, where the first `Location` is the start of the token, and the second is the end of the token. 

Finally, you need a corresponding section in your grammar that tells LALRPOP what your location and error types are, and which variant corresponds to which terminal. Here is the [one LALRPOP uses][4].

[1]: https://github.com/nikomatsakis/lalrpop/blob/d17fdd812731e2794e0196bb21f669534f1e963e/lalrpop/src/tok/mod.rs
[2]: https://github.com/nikomatsakis/lalrpop/blob/d17fdd812731e2794e0196bb21f669534f1e963e/lalrpop/src/tok/mod.rs#L32-L86
[3]: https://github.com/nikomatsakis/lalrpop/blob/d17fdd812731e2794e0196bb21f669534f1e963e/lalrpop/src/tok/mod.rs#L606-L608
[4]: https://github.com/nikomatsakis/lalrpop/blob/d17fdd812731e2794e0196bb21f669534f1e963e/lalrpop/src/parser/lrgrammar.lalrpop#L335-L387

nikomatsakis:
[...]

PR #131, opened 2016-08-17
Rough draft of lexer tutorial

malleusinferni:
Partly addresses #39. I implemented a Whitespace parser to demonstrate custom lexers.

### DIFF
--- a/doc/lexer_tutorial.md
+++ b/doc/lexer_tutorial.md
@@ -163,7 +163,16 @@ extern {
 }
 ```
 
-From now on, the parser will take a `Lexer` as its input instead of a string slice. And any time we write a string literal in the grammar, it'll substitute a variant of our `Tok` enum. This means **we don't have to change any of the rules we already wrote!** This will work as-is:
+From now on, the parser will take a `Lexer` as its input instead of a string slice, like so:
+
+```rust
+    let lexer = lexer::Lexer::new("\n\n\n");
+    match parser::parse_Program(lexer) {
+        ...
+    }
+```
+
+And any time we write a string literal in the grammar, it'll substitute a variant of our `Tok` enum. This means **we don't have to change any of the rules we already wrote!** This will work as-is:
 
 ```rust
 FlowCtrl: ast::Stmt = {
--- a/doc/lexer_tutorial.md
+++ b/doc/lexer_tutorial.md
@@ -177,7 +177,7 @@ FlowCtrl: ast::Stmt = {
 };
 ```
 
-The complete grammar is available in `whitespace.lalrpop`.
+The complete grammar is available in `whitespace/src/parser.lalrpop`.
 
 ## Where to go from here
 
--- a/doc/lexer_tutorial.md
+++ b/doc/lexer_tutorial.md
@@ -97,14 +97,14 @@ impl<'input> Iterator for Lexer<'input> {
 
     fn next(&mut self) -> Option<Self::Item> {
         loop {
-            return Some(Ok(match self.chars.next() {
-                Some((i, ' ')) => (i, Tok::Space, i+1),
-                Some((i, '\t')) => (i, Tok::Tab, i+1),
-                Some((i, '\n')) => (i, Tok::Linefeed, i+1),
+            match self.chars.next() {
+                Some((i, ' ')) => return Some(Ok((i, Tok::Space, i+1))),
+                Some((i, '\t')) => return Some(Ok((i, Tok::Tab, i+1))),
+                Some((i, '\n')) => return Some(Ok((i, Tok::Linefeed, i+1))),
 
                 None => return None, // End of file
                 _ => continue, // Comment; skip this character
-            }))
+            }
         }
     }
 }
--- a/doc/lexer_tutorial.md
+++ b/doc/lexer_tutorial.md
@@ -114,7 +114,7 @@ That's it. That's all we need.
 
 ## Updating the parser
 
-To use this with LALRPOP, we need to expose its API to the parser. It's pretty easy to do, but also somewhat magical, so pay close attention. At the bottom of the grammar file, we need an `extern` block:
+To use this with LALRPOP, we need to expose its API to the parser. It's pretty easy to do, but also somewhat magical, so pay close attention. Pick a convenient place in the grammar file (I chose the bottom) and insert an `extern` block:
 
 ```rust
 extern {
--- a/doc/lexer_tutorial.md
+++ b/doc/lexer_tutorial.md
@@ -1,0 +1,192 @@
+# Writing a custom lexer
+
+Let's say we want to parse the Whitespace language, so we've put together a grammar like the following:
+
+```rust
+pub Program = <Statement*>;
+
+Statement: ast::Stmt = {
+    " " <StackOp>,
+    "\t" " " <MathOp>,
+    "\t" "\t" <HeapOp>,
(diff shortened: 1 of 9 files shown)


doc/lexer_tutorial.md
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+}
+```
+
+That's it. That's all we need.
+
+## Updating the parser
+
+To use this with LALRPOP, we need to expose its API to the parser. It's pretty easy to do, but also somewhat magical, so pay close attention. At the bottom of the grammar file, we need an `extern` block:
```
nikomatsakis: Nit: I think the `extern` section can appear anywhere, not just the bottom.

doc/lexer_tutorial.md
```diff
@@ -0,0 +1,192 @@
...
+    Space,
+    Tab,
+    Linefeed,
+}
+```
+
+Everything else is a comment. There are no invalid lexes, so we'll define our own error type, a void enum:
+
+```rust
+pub enum LexicalError {
```
nikomatsakis: Nit: I *think* the error can be omitted if the lexer is infallible -- but then I like including it, since most lexers *can* fail. And I might have taken away the ability to omit it, I forget. :)

doc/lexer_tutorial.md
```diff
@@ -0,0 +1,192 @@
...
+    type Location = usize;
+    type Error = lexer::LexicalError;
+
+    enum lexer::Tok {
+        // ...
+    }
+}
+```
+
+Now we have to declare each of our terminals. For each variant of `Tok`, we pick what name the parser will see, and write a pattern of the form `name => lexer::Tok::Variant`, similar to how action code works in grammar rules. The name can be an identifier, or a string literal. We'll use the latter.
```
nikomatsakis: Nit: The actual form is a bit more flexible. For example, what comes after the `=>` can be a full-on pattern, not just a variant name, and the tokens need not just be `"foo"` but can also be other identifiers like `foo`. I'm not sure how much to go into here though -- I like that this tutorial starts in a simple place; I'd expect a follow-up to show more of the full capabilities. But it seems worth mentioning that this is a simplified variant, or perhaps listing in the follow-ups.

doc/lexer_tutorial.md
```diff
@@ -0,0 +1,192 @@
...
+
+Writing a lexer for a language with multi-character tokens can get very complicated, but this is so straightforward, we can translate it directly into code without thinking very hard. Here's our `Iterator` implementation:
+
+```rust
+impl<'input> Iterator for Lexer<'input> {
+    type Item = Spanned<Tok, usize, LexicalError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            return Some(Ok(match self.chars.next() {
```
nikomatsakis: I wonder if this code would be clearer with the `return` moved into the arms? This structure is more concise but I fear it may throw people a bit. e.g.,

```rust
loop {
    match self.chars.next() {
        Some((i, ' ')) => return Some(Ok((i, Tok::Space, i+1))),
        ...
    }
}
```

nikomatsakis:
This is great! I like it quite a bit. I left a few comments; the main thing I am wondering though is the best way to introduce the remaining topics. I guess it'd be ok to just have a follow-up tutorial that parses something completely different -- the main thing I was thinking that is missing is having enum variants that carry data, and showing how that works, as well as terminals with non-quoted names.

sbstp:
I went through the tutorial and I have few questions:
1. How does the grammar know that it should use the Lexer struct?
2. Can this Lexer only parse text that contains spaces, tabs and line feeds? How are the rest of the tokens processed?

nikomatsakis:
@sbstp 

> The enum "re-declaration" is like a reverse match

Yes. Specifically, for each terminal in your grammar, it indicates what match pattern would produce it. (I wonder if I should reverse the order..)

nikomatsakis:
>  1. How does the grammar know that it should use the Lexer struct?

A good question! The grammar does *not* know that. The grammar generates a parse function that can be given any iterator that produces the appropriate `Tok` tokens. The connection to the lexer struct is made just by calling the parse function and giving an instance of the `Lexer` struct. I wonder if this can be clarified.

malleusinferni:
@sbstp 

Regarding point 1: You have to explicitly invoke the `parse_Foo()` functions with a Lexer struct as the argument, instead of a string slice:

```rust
    let input = lexer::Lexer::new("\n\n\n");
    let program = parser::parse_Program(input).expect("Oh no");
```

I just neglected to point that out in the tutorial text. I'll update it.

@nikomatsakis

> Yes. Specifically, for each terminal in your grammar, it indicates what match pattern would produce it. (I wonder if I should reverse the order..)

My single major complaint about LALRPOP's lexer support is the syntax. If you're open to changing it, that would go a long way towards making this tutorial unnecessary. I'll create a separate issue for the particulars of that discussion.

I pretty much agree with your other points and I'll push some changes later tonight.

sbstp:
How does the default tokenizer parse arbitrary terminals? Is it generated using the grammar? If a have a string "func" in my grammar, is a unique variant created for it?

sbstp:
> I wonder if I should reverse the order..

Unless there are gains to having it this way, the standard match style would probably make it easier for people understand what's going on.

malleusinferni:
@nikomatsakis @sbstp I've opened #132.

nikomatsakis:
@sbstp 

> How does the default tokenizer parse arbitrary terminals? Is it generated using the grammar? If a have a terminal "func" in my grammar, is a unique variant created for it?

I'm not sure what you mean by "arbitrary terminals". The default tokenizer effectively creates a massive regular expression from all the strings appearing in your grammar. So if you write `"foo"`, `"bar"`, and `r"[a-z]+"`, it would make a regex like `foo|bar|[a-z]+`. It then finds the longest match it can using that regex. (It's a little bit more clever than that, in that it has a notion of priorities, so that keywords like "foo" have precedence, but that's the basic idea.) The actual result is then a token of type `(usize, &'input str)`, I think -- the `usize` is just an integer identifying which string matched (`"foo"`, `"baz"`, etc) and the `&'input str` is the string that matched. 

So in terms of an extern block it would be something like:

```rust
extern {
    enum (usize, &'input str) {
        "foo" => (0, <&'input str>),
        "bar" => (1, <&'input str>),
        r"[a-z]+" => (2, <&'input str>),
    }
}
```

> EDIT: Also how do regexes work with a custom tokenizer? Should the Tok enum have an unknown string variant?

Using a regex with a custom tokenizer probably just doesn't make sense. I think they are treated like any other token -- you must have an entry that maps them to a pattern of some kind.

doc/lexer_tutorial.md
malleusinferni: I tested this, and it defaults to `type Error = ()`. Maybe there should be a section describing all the defaults?

malleusinferni:
I've pushed a couple of fixes. The last section is still a skeleton. There are so many more things it could explain how to do, but the tutorial would explode in size and it may not be worthwhile compared to documenting a lexer generator. As such, it seems to me that the high-priority goal is experimenting with the current crop of lexer libraries and working out how to make one or more of those play nice with LALRPOP. I'll dig into this for a few days if nobody objects.

nikomatsakis:
@malleusinferni great! I think I will merge this for now, and we can improve more as we go! Thanks very much. :)

Issue #156, opened 2016-10-11
Is it possible to do partial parsing, fallback and wildcard ?


gwenn:
Hello,
I am trying to parse the SQLite3 grammar with LALRPOP [here](https://github.com/gwenn/sqlpop).
And some Lemon features do not seem to be avaible with LALRPOP.
Maybe I missed them while reading the documentation.

If they are not supported, it is not really important.
Many thanks for this great parser.

## Partial parsing

SQLite parses one command at a time.
The user can choose to ignore the unparsed chunk (`pzTail `).
Or choose to parse the next command.

### Lemon grammar

```
cmdx ::= cmd.           { sqlite3FinishCoding(pParse); }
```

For example, if the input is `CREATE TABLE test (data); INSERT INTO test VALUES ('');`.
SQLite parses the first statement `CREATE TABLE test (data);` and returns the unparsed chunk ` INSERT INTO test VALUES ('');`.

## Fallback

SQLite makes possible to use some keywords without double-quoting them where an identifier is expected.
For example, `CREATE TABLE test (view TEXT);` is a valid statement.
The `view` keyword becomes a simple identifer.

### Lemon grammar

```
// An IDENTIFIER can be a generic identifier, or one of several
// keywords.  Any non-standard keyword can also be an identifier.
//
%token_class id  ID|INDEXED.

// The following directive causes tokens ABORT, AFTER, ASC, etc. to
// fallback to ID if they will not parse as their original value.
// This obviates the need for the "id" nonterminal.
//
%fallback ID
  ABORT ACTION AFTER ANALYZE ASC ATTACH BEFORE BEGIN BY CASCADE CAST COLUMNKW
  CONFLICT DATABASE DEFERRED DESC DETACH EACH END EXCLUSIVE EXPLAIN FAIL FOR
  IGNORE IMMEDIATE INITIALLY INSTEAD LIKE_KW MATCH NO PLAN
  QUERY KEY OF OFFSET PRAGMA RAISE RECURSIVE RELEASE REPLACE RESTRICT ROW
  ROLLBACK SAVEPOINT TEMP TRIGGER VACUUM VIEW VIRTUAL WITH WITHOUT
  EXCEPT INTERSECT UNION
  REINDEX RENAME CTIME_KW IF
  .
```

### Lemon grammar

```
%wildcard ANY.
```

https://www.sqlite.org/src/doc/trunk/doc/lemon.html
[...]

PR #160, opened 2016-11-08
feat: Implement error recovery

Marwes:
Work in progress branch which aims to allow parsers to recover from parse errors.

This adds the `error` keyword (bikesheddable) to LALRPOPs syntax. If `error` is encountered during parsing the parser goes into recovery mode where it will first pop states until it finds a state which it can potentially recover from. After finding a recoverable state the next step is to skip tokens until a token is found which allows the parser to continue. Recovering from an error acts as a successful parse so it is up to the user to record this error in some way.

### Unresolved questions
- [x] ~~Is it possible to have recovery for LALR grammars as well? (Need to look into this but haven't found anything concrete yet)~~ I guess his might come for free actually?

### TODO
- [x] Recover from errors at EOF
- [x] Allow the error recovered from to be inspected in its action `"+" <e: error> => { /* use `e` */ }`
- [x] ~~Implement recovery for recursive ascent parsers~~ Skipping this until later when the parse_table implementation has seen some use.
- [x] Implement `unimplemented!()` stubs (example generation)
- [x] Documentation
- [x] Clean up commits

### Reading
https://arxiv.org/pdf/1010.1234.pdf
https://www.cs.clemson.edu/course/cpsc827/material/LRk/LR%20Error%20Recovery.pdf
https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=5&ved=0ahUKEwjt9cTt6JjQAhVGjSwKHV0bAigQFgg7MAQ&url=http%3A%2F%2Fwww.cs.princeton.edu%2Fcourses%2Farchive%2Fspr05%2Fcos320%2Fnotes%2F7-parsing-error.ppt&usg=AFQjCNF0I2l3VAMVyT47DIYGv6wZOUzP2A&sig2=0-c3w6nOFevP5nTamen3dg&cad=rja

cc #156

BREAKING CHANGE

This adds the `error` keyword to grammar definitions.


nikomatsakis:
I haven't had a chance yet to look into the code, but I'm super excited to see progress here!

Marwes:
:) Still a lot of to do so there is no rush. Going to submit a separate PR removing the generated .rs files tonight which should make reviewing the changes a bit easier.

nikomatsakis:
@Marwes let me know when to take a look :)

Marwes:
Going to rebase on master to remove the generated files and add some more tests tonight after which it should be possible to review error recovery for parse tables (it largely is ready now but the huge diffs may obfuscate it I think). The recursive ascent parser might be a bit trickier to get right so I may or may not leave that until the parse tables work as expected.

lalrpop/src/grammar/repr.rs
```diff
@@ -109,6 +109,7 @@ pub struct Production {
 pub enum Symbol {
     Nonterminal(NonterminalString),
     Terminal(TerminalString),
+    Error,
```
Marwes: Is this the correct placement of `error`. Considered putting it under `Symbol::Terminal` as well but this was simpler (barring the current travis failure).

lalrpop/src/lr1/codegen/base.rs
```diff
@@ -148,7 +148,7 @@ impl<'codegen, 'grammar, W: Write, C> CodeGenerator<'codegen, 'grammar, W, C> {
         if self.grammar.intern_token.is_some() {
             // if we are generating the tokenizer, create a matcher as our input iterator
             rust!(self.out,
-                  "let mut {}tokens = {}::{}intern_token::{}Matcher::new(input);",
+                  "let mut {}tokens = {}::{}intern_token::{}Matcher::new(input).peekable();",
```
Marwes: Needed to make the tokens iterator peekable so that error recovery can check if parsing the next token would succeed. Might be able to rewrite how  `tokens.next()` is done so that `peekable` is not necessary.

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -238,6 +238,9 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
             rust!(self.out, "{}({}),", name, ty);
         }
 
+        let error_type = self.types.parse_error_type();
+        rust!(self.out, "Error({})", error_type);
```
Marwes: Is it fine to just use `Error` as the name here (no risk of conflicts)?

lalrpop/src/tok/mod.rs
```diff
@@ -83,6 +83,7 @@ pub enum Tok<'input> {
     Star,
     TildeTilde,
     Underscore,
+    ErrorT
```
Marwes: Got a conflict with the `Error` struct defined in this module so went with the ugly `ErrorT` name.

Marwes:
@nikomatsakis This should be good to review now (the remaining things to add aren't crucial to see if the parse_table implementation makes sense).

lalrpop/src/grammar/repr.rs
nikomatsakis: Huh. Yes, *probably* this is right. =)

lalrpop/src/lr1/build/mod.rs
```diff
@@ -203,7 +209,8 @@ impl<'grammar, L: LookaheadBuild> LR<'grammar, L> {
             // The `nt` will be X and the `remainder` will be `z...`.
             let (nt, remainder) = match shift_symbol {
                 None => continue, // requires a reduce
-                Some((Symbol::Terminal(_), _)) => continue, // requires a shift
+                Some((Symbol::Terminal(_), _)) |
+                Some((Symbol::Error, _)) => continue, // requires a shift
```
nikomatsakis: Nit: can you put `=> { \n continue; \n } \n` instead?

Otherwise, I find it hard to visually "see" the groups.

lalrpop/src/lr1/codegen/ascent.rs
```diff
@@ -317,6 +317,7 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent,
             for (index, &token) in tokens.iter().enumerate() {
                 let pattern = match token {
                     Token::Terminal(s) => format!("Some({})", self.match_terminal_pattern(s)),
+                    Token::Error => unimplemented!(),
```
nikomatsakis: BTW, I'm fine with landing this without support in the recursive ascent style -- but I'd prefer to give a structured error. =)

lalrpop/src/lr1/codegen/parse_table.rs
nikomatsakis: Seems like it should be fine, since we normally prefix the variant names with `Nt` or `Term` in `variant_name_for_symbol`. Maybe add a comment about this?

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -272,6 +275,28 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
 
         rust!(self.out, "];");
 
+        // Only need the ERROR table if error recovery is used
+        if self.states.iter().any(|state| state.error.is_some()) {
+            // Error transitions which are positive integers for states that can be recovered from and zero
```
nikomatsakis: Nit: Missing comma, word: "Error transitions**,** which are positive integers for states that can be recovered from and zero **otherwise**"

nikomatsakis: Can we extend the big comment at the top that explains the overall strategy? In the particular, the part that starts "We create three parse tables:" :)

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -459,12 +493,77 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
...
-              "return Err({}lalrpop_util::ParseError::UnrecognizedToken {{",
+              "let {}error = {}lalrpop_util::ParseError::UnrecognizedToken {{",
+              self.prefix,
               self.prefix);
         rust!(self.out, "token: None,");
         rust!(self.out, "expected: vec![],");
-        rust!(self.out, "}});");
+        rust!(self.out, "}};");
+
+        if self.states.iter().any(|state| state.error.is_some()) {
```
nikomatsakis: Nit: we can we, at minimum, extract this out into a separate fn, but -- better yet -- merge this with `error_recovery`? It seems like the only difference is how it tests whether to keep going at a particular state, right? (i.e., consult `EOF_ACTIONS` vs `ACTIONS`?)

lalrpop/src/normalize/prevalidate/mod.rs
```diff
@@ -185,6 +185,8 @@ impl<'grammar> Validator<'grammar> {
             SymbolKind::Nonterminal(_) => {
                 /* see resolve */
             }
+            SymbolKind::Error => {
+            }
```
nikomatsakis: If we were going to give a nicer error when combining `error` and `recursive-ascent`, this *might* be an ok place to do it. I forget if we know what kind of backend we are using at this point or what.

lalrpop/src/normalize/tyinfer/mod.rs
```diff
@@ -273,7 +273,8 @@ impl<'grammar> TypeInferencer<'grammar> {
             SymbolKind::Name(_, ref s) => self.symbol_type(&s.kind),
 
             SymbolKind::Repeat(..) | SymbolKind::Expr(..) | SymbolKind::Macro(..) |
-            SymbolKind::AmbiguousId(..) | SymbolKind::Lookahead | SymbolKind::Lookbehind => {
+            SymbolKind::AmbiguousId(..) | SymbolKind::Lookahead | SymbolKind::Lookbehind |
+            SymbolKind::Error => {
```
nikomatsakis: Wait, I don't think `SymbolKind::Error` belongs here -- it doesn't get macro-expanded away?

I think this would generate an error if you had something like

```rust
Foo = error;
```

which would cause us to try and infer the type of `Foo`.

lalrpop/src/tok/mod.rs
nikomatsakis: Ok, maybe `ErrorTok`?

lalrpop/src/build/mod.rs
```diff
@@ -348,8 +348,14 @@ fn emit_recursive_ascent(session: &Session, grammar: &r::Grammar) -> io::Result<
                                                         "super",
                                                         &mut rust)),
 
-            r::LrCodeGeneration::TestAll =>
-                try!(lr1::codegen::test_all::compile(&grammar, user_nt, start_nt, &states, &mut rust)),
+            r::LrCodeGeneration::TestAll => {
+                try!(lr1::codegen::test_all::compile(&grammar, user_nt, start_nt, &states, &mut rust));
+                rust!(rust,
```
nikomatsakis: wait, what is happening here? I never did figure out why this `pub use` is needed...

nikomatsakis: oh, is this just a hack b/c it doesn't work with `ascent` yet? 

maybe we can just have a test here that says like "FIXME() -- if the grammar uses error, just generate the parse-table variant for now"

Marwes: Yep, this is just a hack which I mean to remove after adding recovery for ascent parsers (or make it less hacky at least).

lalrpop/src/normalize/tyinfer/mod.rs
Marwes: Oops, that is a remnant from when I started and just needed something compiling after adding the `Error` variant.

Marwes:
When writing the docs for this PR I found an unfortunate interaction with repetitive patterns/reductions. In the added calculator7 test, once the parser encounters the erroneous `*` token it is in a state which reduces on `(` and `Num` (start of an expression) and errors otherwise. Unfortunately the `ERROR` table has no recovery for this state so it will proceed to pop states until it comes to the starting state (`0`), effectively removing the expression before the `,`.

If the reduction had been done first regardless of the fact that `*` it is going to result in a parse error the "GOTO" state after the reduction does have error recovery installed which would then result in the parser to recover in the way shown in the test.

Question is, can a reduction always be done eagerly, regardless of the lookahead resulting in an error? Will try and look into this but I wanted to pose the question here as well.

Marwes:
Made a rather significant rewrite of how `error` is treated. It is now placed under `TerminalString` so that it is handled the same as any other terminal when it comes to table generation. This means that the `ERROR` table is now gone with its information existing under the `error` terminal in `ACTION` instead. The reason for this change is due to the extra step that was added to recovery. It now does reductions before checking if a state can be shifted which means that it will be moved into a recoverable state (if one exists).

I tried to deduplicate the error recovery as well but only managed to do it the the generation, the generated code is still duplicated, I can try and fix this tomorrow possibly.

Added some documentation as well, I can extend it a bit if needed but I would prefer not to block on it and instead improve that later on.

nikomatsakis:
@Marwes cool! will take a look

Marwes:
Failure seems like it could be a nightly regression?

```
thread 'lr1::build::test::expr_grammar1' has overflowed its stack

fatal runtime error: stack overflow

error: process didn't exit successfully: `/home/travis/build/nikomatsakis/lalrpop/target/debug/deps/lalrpop-c430fcbb5e90f058` (signal: 6, SIGABRT: process abort signal)
```

Marwes:
Yeah, it is a bit more complicated but I do think it is necessary. From re-reading the paper in the first link I realized that `error` needs to be treated as any other terminal with respect to generating reductions and so that ambiguities can be found if one overuses it. I should perhaps add a test that ambiguity errors are generated but it should just be handled given that it is just a terminal now.

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -809,6 +965,83 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
...
+        rust!(self.out, "break;");
+        rust!(self.out, "}}");
+
+        if DEBUG_PRINT {
+            rust!(self.out, r#"println!("Error recovery reduces on action: {{}}", {}action);"#,
+                  self.prefix);
+        }
+
+        rust!(self.out,
+              "if let Some(r) = {}reduce({}{}action, None, &mut {}states, &mut {}symbols, {}) {{",
```
nikomatsakis: I am curious why supply `None` here -- from skimming the code, this parameter is the lookahead start for the next token, which it seems like we *do* have, right?

doc/tutorial.md
```diff
@@ -691,6 +691,64 @@ fn calculator5() {
 }
 ```
 
+<a id="calculator7"></a>
```
nikomatsakis: I can't tell what `calculator6` was supposed to be...seems like this could just be 6...

lalrpop-test/src/main.rs
```diff
@@ -281,6 +287,65 @@ fn error_test1() {
...
+    assert_eq!(errors.borrow()[0], ParseError::UnrecognizedToken {
+        token: Some(((), Tok::Plus,())),
+        expected: vec![],
+    });
+}
+
+#[test]
+fn error_recovery_dont_drop_unrecognized_token() {
+    let errors = RefCell::new(vec![]);
+    util::test(|v| error_recovery::parse_Item(&errors, v), "(--)", "!".to_string());
```
nikomatsakis: Hmm, I sort of expected this to yield `(!)` as the result -- i.e., we start with the `(`, then push an item state, then encounter an error on the `)`, so we push `error`, but I'd expet the `)` to be in the follow-set, so we'd produce a `!` that gets wrapped in `()`, no?

In general, it would be good to add some comments to each test demonstrating the flow. The names are great but comments are good too.

nikomatsakis:
Well, this is the branch: https://github.com/nikomatsakis/lalrpop/tree/symbol_error

BTW, regarding the `error` keyword -- I wonder if we should use a symbol, like `!`?

nikomatsakis:
> Yeah, it is a bit more complicated but I do think it is a necessary change.

It actually feels simpler to me. =) Perhaps because it's less intrusive to the overall concept.

lalrpop-test/src/main.rs
Marwes: That's just due to how lazily I implemented the parser https://github.com/nikomatsakis/lalrpop/pull/160/files#diff-a8585f30f0fedc26b330b3798c5c7febR22. I could change it to to wrap the result in parentheses.

doc/tutorial.md
Marwes: Meant to ask what it was for, I can swap calculator6 and 7 as they are now or should I remove 6?

Marwes:
> BTW, regarding the error keyword -- I wonder if we should use a symbol, like `!`?

That might make sense and would make this a non-breaking change, it may make it harder to find docs for however.

nikomatsakis:
@Marwes one last thought. The value of the `error` token currently is a `ParseError` -- I was thinking that perhaps it should be a struct:

```rust
struct ErrorRecovery {
    error: ParseError,
    dropped_tokens: Vec<T>
}
```

where we collect up the dropped tokens so that you can have access to them.

nikomatsakis:
What do you think about that?

Marwes:
That seems like a good idea (losing information is usually a bad). It should be trivial to add as well.

Which of `error` and `!` do you think is better? I could see an argument against `error` in that it just looks like any other identifier. Usually that would be obvious through syntax highlighting but as that does not exist it may be confusing in its own way.

nikomatsakis:
> Which of error and ! do you think is better? I could see an argument against error in that it just looks like any other identifier. Usually that would be obvious through syntax highlighting but as that does not exist it may be confusing in its own way.

I'm not sure. `!` in Rust is usually associated with panicking. The error recovery mechanism actually has something in common with panicking -- or at least exceptions -- since it unwinds the grammar, so to speak. Mmmm, I think let's go with `!` for now. I've tried whereever possible to leverage Rust syntax and this seems like a *reasonably* close mapping, no?

nikomatsakis:
Oh, I also have some edits I did locally. I have now learned and understood how GH's corrections feature works. I will go ahead and push a commit with various nits to your branch. Feel free to object if you don't like them. =)

Marwes:
Rebased with your commits, added the `ErrorRecovery` type and changed `error` to `!` which should hopefully tackle the last few things. Saw that I had accidentally committed with some trailing whitespace (oops), usually got rustfmt handling that automatically but had it disabled to prevent messy diffs. Was thinking I would submit a PR with rustfmt run on the source afterwards if that is alright, possibly with a rustfmt.toml to preserve formatting as much as possible?

nikomatsakis:
> Was thinking I would submit a PR with rustfmt run on the source afterwards if that is alright, possibly with a rustfmt.toml to preserve formatting as much as possible?

Yes, that's fine, I've been meaning to do that for a while. I'd prefer to just use standard settings and not mess with a rustfmt.toml.

nikomatsakis:
This is great work. Very cool!

ilsken:
>  Implement recovery for recursive ascent parsers Skipping this until later when the parse_table implementation has seen some use.

Any news on this front?

Marwes:
@ilsken Forgot about the recursive ascent part. I can't really prioritize it over other PRs right now though but perhaps in a few months I may be a bit freer.

Issue #164, opened 2016-11-15
Don't output debug information by default


amccreight:
I'm working on a parser for Firefox's IPDL language ( https://bugzilla.mozilla.org/show_bug.cgi?id=1316754 ), which we might want to eventually use as part of the Firefox build process. Building all of the frontend dependencies stuff takes a really long time (over four minutes on my machine for --release, mostly in lalrpop-snap), so we may want to commit the output of the lalrpop compiler into the tree to avoid all of that for most people.

The .rs file generated by lalrpop is quite large, at around 1.48mb. Most of the lines of the file look like this:
        0, // on "upto", error

It would be nice if by default the output of lalrpop did not generate this debugging information. I hacked up a Python script that removes line comments, leading whitespace, and squashes the table into a single line without any whitespace in it (so the entries look like "0,0,0,0,0,0,0,0,0,0,0") and that reduces the file size to 322kb, which is still large, but isn't quite as bad. I'm not sure if there are any further improvements to be had in the file size, but there at least wasn't anything obvious with that.

nikomatsakis:
@amccreight sure, I couldn't come up with a persuasive reason to care about absolute filesize. But it'd be fairly easy to omit that stuff by default.

ahmedcharles:
Is there more information that it would be beneficial to leave out, assuming it makes compile faster? Or perhaps just for smaller file size?

amccreight:
It didn't make any difference for compile time, only the file size.

nikomatsakis:
I think we already do this, write? So I guess we can close this issue (doing so). Seems sort of like a dup of https://github.com/lalrpop/lalrpop/issues/304 otherwise.

PR #165, opened 2016-11-18
fix: Shrink the amount of lines written for parse tables

Marwes:
This commit avoids writing a description of each terminal as a comment if comments are disabled. It also writes all terminals for a state on a single line which effectively reduces the size of tables from `O(terminals * states)` to `O(states)` (if comments are disabled).

cc #65, #164


Marwes:
Gist of calculator7.rs from #160 https://gist.github.com/Marwes/17489e0779fd2a8dda26c9311a6219c1

PR #169, opened 2016-11-23
Release 0.12.2

nikomatsakis:
# Version 0.12.2

Major new feature! @Marwes #[added support for error recovery](https://github.com/nikomatsakis/lalrpop/pull/160).

There have also been a number of other improvements:

- The `ParseError` type now implements `Error` and `Display` ([thanks @Marwes!](https://github.com/nikomatsakis/lalrpop/pull/167)).
- We no longer emit comments in generated code by default ([thanks @Marwes!](https://github.com/nikomatsakis/lalrpop/pull/165)).


nikomatsakis:
Hmm. There is a problem that the new error recovery code requires that tokens are clone (which didn't used to be the case). @Marwes is going to fix that. I'll yank the release for now.

brendanzab:
When will 12.3 be released? :)

PR #172, opened 2016-11-28
Release 0.12.3

nikomatsakis:
Major new feature! @Marwes [added support for error recovery](https://github.com/nikomatsakis/lalrpop/pull/160).

There have also been a number of other improvements:

- The `ParseError` type now implements `Error` and `Display` ([thanks @Marwes!](https://github.com/nikomatsakis/lalrpop/pull/167)).
- We no longer emit comments in generated code by default ([thanks @Marwes!](https://github.com/nikomatsakis/lalrpop/pull/165)).


nikomatsakis:
Argh. OK, let's try this *one more time*. =)

Marwes:
Well, at least it won't be the same error now that there is an actual test :)

PR #173, opened 2016-11-29
fix: Actually avoid all clones when not using error recovery

Marwes:


lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -499,14 +499,14 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
         // Error.
         rust!(self.out, "}} else {{");
 
-        rust!(self.out,
-            "let {}error = {}lalrpop_util::ParseError::UnrecognizedToken {{",
-            self.prefix,
-            self.prefix);
-        rust!(self.out, "token: Some({}lookahead.clone()),", self.prefix);
```
Marwes: 😓

Marwes:
Ironically its now failing because its using 0.12.3 :( (and https://github.com/rust-lang/cargo/issues/3340)

lalrpop/src/lr1/codegen/parse_table.rs
nikomatsakis: heh, d'oh!

Issue #58, opened 2016-01-23
Include expected tokens on parse error


nikomatsakis:
In the `ParseError` variant `UnrecognizedToken`, I included a spot to list the tokens that _would_ have been accepted -- but I never wrote the code to fill it in. It might also be interested to include nonterminals, though I imagine many tools would just want to screen them out. One question is how to format this list -- do we use the names from the grammar? The regular expressions? I guess so, not much else to use, unless we add a way for users to configure what gets added in the lalrpop file.

cc @LukasKalbertodt, who was asking around this.

LukasKalbertodt:
What about having a `Vec<T>`, where `T` is the token type from `ParseError`. The user could then decide how to format it for output. Formatting it into a `String` in lalrpop doesn't make too much sense to me right now. Am I missing something? 

Including non-terminals is an interesting options as well. However, I'm not sure whether passing only the name of the non-terminal as `String` is a good way to do it...

nikomatsakis:
@LukasKalbertodt the generated code cannot construct instances of the token type from `ParseError`. It just knows how to match against those tokens, not how to build them. After all, consider a token like `Identifier(&'input str)` -- how are we supposed to synthesize the string that goes inside?

I was considering something like:

```
enum GrammarSymbol {
    Terminal(&'static str),
    Nonterminal(&'static str)
}
```

where the strings are literally the strings from the grammar. It'd be your job to "pretty print" those better. (In the future, we could also maybe have some syntax in the `lalrpop` file to customize what gets returned.)

LukasKalbertodt:
After thinking about it I guess your solution is pretty neat (and of course my idea doesn't work). But: am I assuming correctly that this issue is rather low in priority? It's a shame to see 635 occurrences of `expected: vec![]` in my generated grammar :P

nikomatsakis:
I've had some further thoughts; I was hoping to fix this in the next few days. I got distracted by improving the LR1 generation algorithm to be faster and make fewer states...

LukasKalbertodt:
> I've had some further thoughts;

Thoughts worth mentioning/explaining here?

>  I got distracted by improving the LR1 generation algorithm to be faster and make fewer states...

I didn't want to sound like I'm complaining -- this is awesome!

nikomatsakis:
[...]

PR #178, opened 2016-12-28
feat: Add the expected successor tokens to UnrecognizedToken errors

Marwes:
This just adds the terminal strings defined in the grammar which is not the best possible information but it is better than no information.

Partial fix for #58


nikomatsakis:
Great =) Thanks!

PR #182, opened 2017-01-28
fix: Don't accidentally cast a reduce action to usize

Marwes:
This caused panics in certain error recovery scenarios as a reduce action represented by a negative number would get casted to `usize` which then caused an index out of bounds panic.

The added tests were extracted and minimized from [gluon's grammar](https://github.com/gluon-lang/gluon/blob/abd06a4f6efa24d5dc4800373f6dfbc066dd661d/parser/src/grammar.lalrpop) with the test case below being a minimal reproduction.

Panicking
```
type Test = a ->
```

Valid expression
```
type Test = a -> a
in 1
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,12 @@ TAGS
 Cargo.lock
 lalrpop/src/parser/lrgrammar.rs
 
+doc/calculator/src/calculator6.rs
+doc/whitespace/src/parser.rs
+
 lalrpop-test/src/error.rs
 lalrpop-test/src/error_recovery.rs
+lalrpop-test/src/error_recovery_pull_182.rs
 lalrpop-test/src/expr.rs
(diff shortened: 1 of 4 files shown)


Marwes:
Bump

nikomatsakis:
@Marwes sorry for slow response! I've been buried in notifications lately. Digging my way out.

nikomatsakis:
@Marwes shall I issue a new point release?

Marwes:
That would be appreciated! I want to do a new release of gluon pretty soon so I'd like to remove the git dependency.

Marwes:
Any chance of that point release? :)

nikomatsakis:
@Marwes yep, sorry, was busy this week with a rustc "design sprint", so have been AFK. I'll try to do it now (though I'm in an airport, so we'll see how that goes.)

PR #186, opened 2017-02-26
unify dep info

king6cong:
cargo test --all can be eventually used for
the whole workspace testing after all doctests
pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ build = "build.rs" # LALRPOP preprocessing
 # needed if you are writing your own tokenizer by
 # hand (or if you are already using the regex crate)
 [dependencies.regex]
-version = "0.2"
+version = "0.12.5"
 
 # Add a dependency on the LALRPOP runtime library:
 [dependencies.lalrpop-util]
--- a/doc/calculator/Cargo.toml
+++ b/doc/calculator/Cargo.toml
@@ -5,9 +5,13 @@ authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 build = "build.rs" # <-- We added this and everything after!
 workspace = "../.."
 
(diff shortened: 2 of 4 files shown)


nikomatsakis:
Hmm, there is a script that is supposed to be keeping these in sync. It's probably missing some entries.

king6cong:
Thanks, I looked into it.

* publish.sh regex failed to match version info like this { version = "0.12.0", path = "../../lalrpop-util" }
* doc/whitespace/Cargo.toml use lalrpop = "0.12.5" directly so duplicate dep files are generated.
* git add -f to make sure .gitignore files do not get in our way.

king6cong:
And cargo test --all fails for some doc tests, but we can leave it for another PR 😄

nikomatsakis:
@king6cong thanks for looking into that! think you can fix the script?

ahmedcharles:
I fixed the doctests, though `cargo test --all` still fails due to the reasons fixed by this change.

You'll have to rebase past one of my changes though, unfortunately, because it changes a Cargo.toml. Sorry. :)

king6cong:
@nikomatsakis PR already updated. Including changes to the script.

king6cong:
@ahmedcharles Thanks! I'am out on a trip and resolved conflict on my phone. Please help me review it

nikomatsakis:
@king6cong it seems fine to me, but can you rebase so as to eliminate the 'merge' commit?

king6cong:
@nikomatsakis Yeah，I will do it as soon as I'm back with a computer 😃

king6cong:
@nikomatsakis PR updated.

nikomatsakis:
Thanks =)

PR #187, opened 2017-02-27
remove DFA codegen and replace with regex-set

nikomatsakis:
I noticed that a lot of the long compile-times with LALRPOP generated code are due to the tokenizer generating very big functions. This can be avoided by writing your own tokenizer, of course, but since the default tokenizer is not necessarily intended to be super efficient, we can also avoid it by using regular expressions at runtime (and, honestly, this may well be faster than what we had, since the regex library is very optimized).

This branch passes LALRPOP's internal tests, but I've not had a chance to try it on anything else.

Note that this does mean that one has to add a dependency on the regex crate if you are using the default tokenizer (but not otherwise).


nikomatsakis:
As a measurement, one of my local repos, when converted to this branch, went from **4 minutes** of compilation time to **0sec**.

nikomatsakis:
I tested the runtime performance of this by checking the pascal example. I did a dozen parses of the `sched.pas`, built both ways, and obtain the following sums for their parse times:

```
lunch-box. re '\(([0-9.]+)s\): OK' '$1' typescript.regex | sum.pl
0.008863779
lunch-box. re '\(([0-9.]+)s\): OK' '$1' typescript.master | sum.pl
0.008735651
```

In short, the regex branch is *ever so slightly* slower. And this is the best case.

PR #196, opened 2017-03-16
Issue157 bracket chars are perceived as part of syntax

wieczyk:
Tokenizer handles rust's character literals in a similar way to string literals.


lalrpop/src/build/mod.rs
```diff
@@ -221,6 +221,9 @@ fn parse_and_normalize_grammar(session: &Session, file_text: &FileText) -> io::R
                 tok::ErrorCode::UnterminatedStringLiteral => {
                     "unterminated string literal; missing `\"`?"
                 }
+                tok::ErrorCode::UnterminatedCharacterLiteral => {
+                    "unterminated character literal; missing `'`?"
```
nikomatsakis: can we test a case that generates this error? I forget if we have a setup for that or not :)

wieczyk: Done.

Issue #179, opened 2016-12-28
Inconsistency when switching parser modes


pothos:
Using the type `#[recursive_ascent]` the attached examples compiles correctly, but as soon as this switch is gone errors like this occur:

```
   --> src/grammar.rs:426:101
    |
426 |                             Tok::Dot { location: __tok0, value: __tok1 } => __Symbol::Term_22_2e_22(__tok0, __tok1),
    |                                                                                                     ^^^^^^^^^^^^^^
    |
    = note: the following parameter type was expected: (std::string::String, std::string::String)

error[E0061]: this function takes 1 parameter but 2 parameters were supplied
```

Did I overlook somethink or is it a bug in the table driven mode? `#[LALR]` does not impact the behaviour.

https://tubcloud.tu-berlin.de/index.php/s/0TLepTZQDrrLp0W

nikomatsakis:
Interesting. This seems like a bug to me, I'm not sure!

pothos:
splendid, this can be closed now :) Big thanks to @wieczyk – just this evening I wanted to investigate it at the Rust meeting in my neighbourhood and realized that it was fixed already.

Now my grammar.rs is only 16 MB with LALR and needs less memory to compile, 29 MB without LALR, 23 MB LALR+recursive_ascent.

PR #197, opened 2017-03-18
parse table generator is now packing terminal parameters into tuple

wieczyk:
Fix for #179.

### DIFF
--- a/lalrpop-test/src/expr.lalrpop
+++ b/lalrpop-test/src/expr.lalrpop
@@ -11,6 +11,7 @@ extern {
         "*" => Tok::Times,
         "/" => Tok::Div,
         Num => Tok::Num(<i32>),
+        Fraction => Tok::Fraction(<i32>, <i32>),
     }
 }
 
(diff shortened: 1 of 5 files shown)


lalrpop-test/src/expr.lalrpop
```diff
@@ -11,6 +11,7 @@ extern {
         "*" => Tok::Times,
         "/" => Tok::Div,
         Num => Tok::Num(<i32>),
+        Fraction => Tok::Fraction(<i32>, <i32>),
```
nikomatsakis: Is this a test that reproduces the problem? (or used to, before your fix)

wieczyk: Yes.

The presence of this declaration is enough to provoke error.

nikomatsakis: Can you add a comment like `// Regression test for #179` or something so that we remember that?

wieczyk: Sure.

nikomatsakis:
Thanks =)

Issue #10, opened 2015-06-24
tokenizer support


nikomatsakis:
We want the ability to add your own tokenizer. This should permit very lightweight specifications but scale to really complex things. I envision the first step as generating a tokenizer based on the terminals that people use (#4) but it'd be nice to actually just permit tokenizer specifications as well, where people can write custom action code based on the strings that have been recognized.

Some things I think we'll want:

- A stack of states for the tokenizer to use
- The ability for action code to yield any number of tokens. I envision that we'll define a trait and require you to return some value that conforms to that trait. This trait will permit:
  - return `Tok` for just one token.
  - if you write a return type of `()`, we expect you to return zero tokens.
  - if you write a return type of `(Tok, Tok)`, you always return two tokens.
  - if you write a return type of `Vec<Tok>`, we expect you to return a dynamic number of tokens.
  - internally, the generated code will keep a queue of generated tokens, and as tokens are requested they are removed from the front, and we only go back to the bytes when that queue is exhausted.

nikomatsakis:
We currently have some simple tokenizer generation support -- it has no states, and you can't write any custom action code at all. There are some immediate needs for extension described in #14, but it'd be nice to tie this all together in a grand, unified design.

nikomatsakis:
My current thought is something like this. You can add `match` sections into the grammar, with an optional priority annotation. These can contain either literal strings or regular expresisons.

```
#[priority=2]
match {
    "foo" => action, // literal text
    r"bar" => action, // regular expression
    "foo" =>? action, // fallible action
}
```

The action will be to produce something that supports `IntoIterator<Item=TheToken>`. Thus you can return `None` to just skip the token, `Some(foo)` to produce one token, or `vec![...]` to produce many tokens (or other such things). If the rule is fallible, you should produce `Result<T,E>` where `E` is a `ParseError` and `T` is something supporting `IntoIterator`, as above. 

I had imagined that the internal token form would "desugar" to this. One catch is that if you are hand-writing code, you would presumably want to specify your own token *type*. The current internal tokenizer just generates tuples like `(u32, &'input str)`, where the integer identifies which token it is. This is pretty opaque and not something you could feed by hand (OTOH, the main use for this form is to specify the whitespace to skip, in which case you'd just return `None`).

[...]

PR #206, opened 2017-03-24
`match` declaration for controlling precedence in the generated tokenizer

wagenet:
See #10

## To Do:

* [x] Cleanup commit history
* [x] Validate for single match token
* [x] Validate that match and extern can't coexist (we may be able to loosen this in the future)
* [x] Support precedence
* [x] Improve comments and variable naming in token_check/mod
* [x] Review all FIXMEs
* [x] More tests


lalrpop/src/grammar/parse_tree.rs
```diff
@@ -44,12 +44,54 @@ impl Into<Box<Content>> for Span {
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MatchContents {
+    pub items: Vec<MatchItem>
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum MatchItem {
+    Unmapped(MatchSymbol, Span),
+    Mapped(MatchSymbol, MatchMapping, Span)
```
nikomatsakis: I suspect that `CatchAll` could be here. That is, I don't think that `_ => "foo"` makes any sense.

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -230,6 +231,29 @@ ExternToken: GrammarItem = {
     },
 };
 
+MatchToken: GrammarItem = {
+    <MatchToken> "else" "{" MatchContents "}"   => <>,
+    "match" "{" <c:MatchContents> "}"           => GrammarItem::MatchToken(MatchToken { contents: vec![c] }),
```
nikomatsakis: you could also do this like so:

```
"match" "{" <c:MatchContents> "}" <e:("else" "{" <MatchContents> "}")> => {
    let mut contents: Vec<_> = Some(c).into_iter().chain(e).collect();
    GrammarItem::MatchToken(MatchToken { contents: contents })
}
```
```

File: lalrpop/src/parser/lrgrammar.lalrpop [hunk 3]
```diff
@@ -230,6 +231,29 @@ ExternToken: GrammarItem = {
...
+    <lo:@L> <s:MatchSymbol> <hi:@R> => MatchItem::Unmapped(s, Span(lo, hi)),
+    <lo:@L> <from:MatchSymbol> <start:@L> <p:"=>"> <hi:@R> =>? {
+        let to = try!(super::parse_match_mapping(p, start + 2));
+        Ok(MatchItem::Mapped(from, to, Span(lo, hi)))
+    }
+};
+
+MatchSymbol: MatchSymbol = {
+    "_"        => MatchSymbol::CatchAll(),
+    <Terminal> => MatchSymbol::Terminal(<>)
```
nikomatsakis: Just FYI, the `<>` aren't needed in this case.

lalrpop/src/grammar/parse_tree.rs
wagenet: That sounds right.

lalrpop/src/parser/lrgrammar.lalrpop
wagenet: Ah, good point.

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -253,6 +310,15 @@ pub enum TerminalString {
     Error,
 }
 
+impl TerminalString {
+    pub fn as_literal(&self) -> Option<&TerminalLiteral> {
```
nikomatsakis: Just FYI, `TerminalLiteral` is `Copy`, so you don't need to return an `Option<&TerminalLiteral>` -- just `Option<TerminalLiteral>` would suffice (and then no `ref` on the pattern below)

lalrpop/src/normalize/lower/mod.rs
```diff
@@ -54,6 +54,10 @@ impl<'s> LowerState<'s> {
                     uses.push(data);
                 }
 
+                pt::GrammarItem::MatchToken(_) => {
+                    // FIXME: Do something?
```
nikomatsakis: I think you can remove this FIXME. I don't think there's anything to do here. 

I'd replace it with a comment like this:

```
// The declarations in the match token are handled
// fully by the `token_check` when it constructs the
//  `InternToken` -- there is nothing left to do here.
```

lalrpop/src/normalize/prevalidate/mod.rs
```diff
@@ -50,6 +50,26 @@ impl<'grammar> Validator<'grammar> {
         for item in &self.grammar.items {
             match *item {
                 GrammarItem::Use(..) => { }
+
+                GrammarItem::MatchToken(ref data) => {
+                    // Ensure that the catch all is final item of final block
+                    let mut contents_idx = 0;
+                    for match_contents in &data.contents {
+                        let mut item_idx = 0;
+                        for item in &match_contents.items {
```
nikomatsakis: Nit:

```
for (item_idx, item) in match_contents.items.iter().enumerate()
```

lalrpop/src/normalize/prevalidate/mod.rs
```diff
@@ -50,6 +50,26 @@ impl<'grammar> Validator<'grammar> {
...
+                            if item.is_catch_all() && (contents_idx != &data.contents.len()-1 || item_idx != &match_contents.items.len()-1) {
+                                return_err!(
+                                    item.span(),
+                                    "Catch all must be final item");
+                            } else {
+                                println!("ok");
+                            }
+                            item_idx += 1;
```
nikomatsakis: ...then you don't need this.

lalrpop/src/normalize/prevalidate/mod.rs
```diff
@@ -50,6 +50,26 @@ impl<'grammar> Validator<'grammar> {
...
+                        }
+                        contents_idx += 1;
```
nikomatsakis: ...and you can convert this variable the same way.

lalrpop/src/normalize/prevalidate/test.rs
```diff
@@ -70,3 +70,19 @@ fn pub_inline_annotation() {
         r#"grammar; #[inline] pub Term = ();"#,
         r#"           ~~~~~~            "#);
 }
+
+#[test]
+fn match_catch_all_first_of_last() {
```
nikomatsakis: yay for tests :)

lalrpop/src/normalize/resolve/mod.rs
```diff
@@ -34,8 +34,19 @@ fn resolve_in_place(grammar: &mut Grammar) -> NormResult<()> {
                        TerminalString::Bare(id) => Some((conversion.span, id, Def::Terminal)),
                    });
 
+        let match_identifiers =
```
nikomatsakis: Seems like a comment would be good here. Just because the rest of LALRPOP is appalingly commented doesn't mean we can't start now. Something like:

```
// Extract all the bare identifiers that appear in the RHS of a `match` declaration.
// Example:
//     match {
//         r"(?)begin" => "BEGIN",
//     } else {
//         r"[a-zA-Z_][a-zA-Z0-9_]*" => IDENTIFIER,
//     }
// This would result in `vec![IDENTIFIER]`.

File: lalrpop/src/normalize/token_check/mod.rs [hunk 11]
```diff
@@ -21,19 +21,50 @@ mod test;
 
 pub fn validate(mut grammar: Grammar) -> NormResult<Grammar> {
     let (has_enum_token, all_literals) = {
+        let opt_match_token = grammar.match_token();
```
nikomatsakis: Side comment: Do we ever check that there is at most one `match_token`? We ought to check in prevalidate I imagine.

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -21,19 +21,50 @@ mod test;
...
+
+        let (match_mappings, match_catch_all) = if let Some(mt) = opt_match_token {
```
nikomatsakis: I wonder if we can find a name for `match_mappings` that makes it clear what *direction* the map takes place. After all, the user writes `r"foo" => "bar"`, but this table is going in the inverse direction. Maybe we can call it `match_rev_mapping`? Or `user_name_to_match_map`? Something like that...

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -21,19 +21,50 @@ mod test;
...
+        } else {
+            (None, None)
+        };
+
         let opt_enum_token = grammar.enum_token();
         let conversions = opt_enum_token.map(|et| {
             et.conversions.iter()
                           .map(|conversion| conversion.from)
-                          .collect()
+                          .collect::<Set<TerminalString>>()
```
nikomatsakis: Is this annotation necessary? (I don't object to it, just curious.)

nikomatsakis: (Usually, I put such annotations on the local variables themselves, though in this case this might be a better choice, since the line with the local variable would get quite long.

Example:

```rust
let conversions: Option<Set<_>> = opt_enum_token.map(|et| {
    ...
};
```

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -21,19 +21,50 @@ mod test;
...
 
         let mut validator = Validator {
             grammar: &grammar,
             all_literals: map(),
             conversions: conversions,
+            match_mappings: match_mappings,
+            match_catch_all: match_catch_all
         };
 
+        // FIXME: Handle this better if possible
```
nikomatsakis: I would check this in prevalidate, and then we can just `assert!` here.

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -56,13 +87,16 @@ struct Validator<'grammar> {
     grammar: &'grammar Grammar,
     all_literals: Map<TerminalLiteral, Span>,
     conversions: Option<Set<TerminalString>>,
+    match_mappings: Option<Map<TerminalString, TerminalString>>,
```
nikomatsakis: Similarly to before, it seems like we might as well put some comments here showing what these maps contain. Bonus points for an example.

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -131,17 +165,49 @@ impl<'grammar> Validator<'grammar> {
...
-                    // terminal; otherwise it's just an unresolved
-                    // identifier.
-                    panic!("bare literal `{}` without extern token definition", c);
-                }
-                TerminalString::Literal(l) => {
-                    self.all_literals.entry(l).or_insert(span);
-                }
+                // FIMXE: Should not allow undefined literals if no CatchAll
+                TerminalString::Bare(c) => match self.match_mappings {
+                    Some(ref m) => {
```
nikomatsakis: Good place for a comment, to remind the reader what the `Some` branch means. i.e.:

```
// If the user gave us a `match` mapping, try to map this back to the actual
// regular expression.
//
// Example:
//     match { r"foo" => "bar" }
// If the user gave us `"bar"`, we would map back to `r"foo"`
```

(something like that, I actually haven't read into what this code does yet;)

lalrpop/src/normalize/prevalidate/mod.rs
wagenet: I tried that before and it was giving me some obscure message about traits. I may try again.

lalrpop/src/normalize/token_check/mod.rs
wagenet: I'm not sure we do check. I agree that we should.

lalrpop/src/normalize/token_check/mod.rs
wagenet: Surprisingly it did seem to be necessary at the time I wrote it. However, I may have simplified stuff so it might not be necessary any more.

wagenet:
@nikomatsakis any other tests you'd like to see to feel comfortable with behavior before merging? Obviously, this still needs some cleanup, but I want to know what I'm aiming for.

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -131,17 +165,49 @@ impl<'grammar> Validator<'grammar> {
...
+                        if let Some(v) = m.get(&term) {
+                            // FIXME: I don't think this span here is correct
+                            let vl = v.as_literal().expect("must map to a literal");
+                            self.all_literals.entry(*vl).or_insert(span);
+                        } else {
+                            return_err!(span, "terminal `{}` does not have a match mapping defined for it",
```
nikomatsakis: I would expect that this path is unreachable, right? Because we would have treated any bare literal that has no `match` mapping as an (undefined) nonterminal and errored out earlier.

nikomatsakis: (In other words, much like the `panic!` in the `else` case below.)

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -131,17 +165,49 @@ impl<'grammar> Validator<'grammar> {
...
+                        // defined for them that indicates they are a
+                        // terminal; otherwise it's just an unresolved
+                        // identifier.
+                        panic!("bare literal `{}` without extern token definition", c);
+                    }
+                },
+
+                TerminalString::Literal(l) => match self.match_mappings {
+                    Some(ref m) => {
+                        if let Some(v) = m.get(&term) {
```
nikomatsakis: again, a prime opportunity for a comment :)

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -131,17 +165,49 @@ impl<'grammar> Validator<'grammar> {
...
+                            // FIXME: I don't think this span here is correct
```
nikomatsakis: why do you think the span is wrong? it will be pointing into the user's code, I guess. still, I'm not 100% sure we should be initializing `all_literals` here anyway, I have to step back and think once I finish reading what you've got...

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -131,17 +165,49 @@ impl<'grammar> Validator<'grammar> {
...
+                            let vl = v.as_literal().expect("must map to a literal");
+                            self.all_literals.entry(*vl).or_insert(span);
+                        } else {
+                            // Unwrap should be safe as we shouldn't have match_catch_all without match_mappings
```
nikomatsakis: I kind of think we should package them together...I'll leave a comment on the struct itself

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -56,13 +87,16 @@ struct Validator<'grammar> {
     grammar: &'grammar Grammar,
     all_literals: Map<TerminalLiteral, Span>,
     conversions: Option<Set<TerminalString>>,
+    match_mappings: Option<Map<TerminalString, TerminalString>>,
+    match_catch_all: Option<bool>,
```
nikomatsakis: instead of having two option fields that are either both some or both none, maybe this:

```rust
/// Summarizes the user's `match` declaration, if any:
///
/// Example:
///     match {
///         "foo" => bar,
///         "baz" => "qux",
///         _
///     };
/// would result in a "rev_mapping" of `bar` to `"foo"` and `"qux"` to `"baz"`.
///
/// The `catch_all` flag would be true, since the `_` is present. 
struct MatchData {
    rev_mapping: Map<TerminalString, TerminalString>,
    catch_all: bool,
}
```

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -21,19 +21,50 @@ mod test;
...
+            let mut catch_all = false;
+
+            // FIXME: This should probably move _inside_ the Validator
+            // FIXME: Add precedences here
+            for mc in &mt.contents {
+                for item in &mc.items {
+                    // TODO: Mabye move this into MatchItem methods
+                    match *item {
+                        MatchItem::Unmapped(sym, _)        => { ret.insert(sym, sym); },
+                        MatchItem::Mapped(sym, mapping, _) => { ret.insert(mapping, sym); },
```
nikomatsakis: I think we should initialize `all_literals` with the results from the `match_data` (using my new struct I proposed elsewhere) right here. Something like this:

```rust
// If we are not using an extern tokenizer, this map contains
// every literal (`"foo"`, `r"foo"`) that we may need to tokenize,
// mapping to a span where it appears in the source. It is filled
// in as we walk the `match` decl (if any) and the source itself.
let mut all_literals = map();
...
let (match_mappings, match_catch_all) = if let Some(mt) = opt_match_token {
    ...
    match *item {
        MatchItem::Unmapped(sym, span)        => {
            ret.insert(sym, sym);
            all_literals.insert(sym, span);
        }
        MatchItem::Mapped(sym, mapping, _) => {
            ret.insert(msppin, sym);
            all_literals.insert(sym, span);
        }
        MatchItem::CatchAll(_) => .. ,
    }
}
```

nikomatsakis:
@wagenet see my [massive comment here](https://github.com/nikomatsakis/lalrpop/pull/206#pullrequestreview-29067140), that explains what I think we have to do make this actually work. 

I'd be happy to merge it in its current form if you wanted, and do some of that in a follow-up PR, but actually I think you're **really close** to having the first draft be completely done. (I had in mind some more knobs and whistles, like being able to map one regex to zero or multiple output tokens, but they're not needed for now.)

I did mention some missing tests and things, but the biggest thing would be an "end-to-end" test. These are made by editing the `lalrpop-test` project. You would add a `match_section.lalrpop`, presumably, and then modify `main.rs` to add (a) [a `mod` declaration](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop-test/src/main.rs#L25) and (b) [some `#[test]` fns that exercise the generated code a bit](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop-test/src/main.rs#L115-L118) (I linked to the `expr_intern_tok` test definitions.)

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -334,8 +336,8 @@ impl TerminalLiteral {
     /// applies when we are creating the tokenizer anyhow.
     pub fn precedence(&self) -> usize {
         match *self {
-            TerminalLiteral::Quoted(_) => 1,
-            TerminalLiteral::Regex(_) => 0,
+            TerminalLiteral::Quoted(_, p) => p,
+            TerminalLiteral::Regex(_, p) => p,
```
nikomatsakis: So I think how I would prefer to handle this is to remove this method altogether, and instead have `token_check` consult a map to compute the precedence. (As opposed to adding the precedence as a second field in `TerminalLiteral`.) I prefer `TerminalLiteral` to just reflect the bytes being matched; precedence is a concept that only applies when we are using the generated tokenizer.

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -118,6 +119,7 @@ pub struct InternToken {
     /// Set of `r"foo"` and `"foo"` literals extracted from the
     /// grammar. Sorted by order of increasing precedence.
     pub literals: Vec<TerminalLiteral>,
+    pub match_to_user_name_map: Option<Map<TerminalString, TerminalString>>,
```
nikomatsakis: In general, I think it would work out better if we merged this with the `literals` Vector into a vector of pairs. This would also force us to consider, each time we iterate, whether we want the *actual literal* that is being matched or the *user literal*, which seems likely to prevent bugs.

(Even if we kept it this way, I personally would not use an `Option<Map<..>>` but just a `Map<..>`, and I'd say that if there is no `match` then either it's an empty map, or we populate it with a `X -> X` mappings.)

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -375,11 +377,22 @@ pub struct MacroSymbol {
...
+        TerminalString::Literal(TerminalLiteral::Quoted(i, 1))
     }
 
     pub fn regex(i: InternedString) -> TerminalString {
-        TerminalString::Literal(TerminalLiteral::Regex(i))
+        TerminalString::Literal(TerminalLiteral::Regex(i, 0))
+    }
+
+    pub fn with_match_precedence(self, p: usize) -> TerminalString {
+        let base_precedence = p * 2;  // Multiply times two since we still want to distinguish between quoted and regex precedence
```
nikomatsakis: I hadn't realized that this would be necessary, but I guess with `_` it makes sense -- i.e., within a section, quoted must take precedence still (but not across sections). Very good.

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -375,11 +377,22 @@ pub struct MacroSymbol {
...
+        match self {
+            TerminalString::Literal(TerminalLiteral::Quoted(i, _)) =>
+                TerminalString::Literal(TerminalLiteral::Quoted(i, base_precedence+1)),
+            TerminalString::Literal(TerminalLiteral::Regex(i, _)) =>
+                TerminalString::Literal(TerminalLiteral::Regex(i, base_precedence+0)),
+            _ => panic!("cannot set match precedence for {:?}", self)
```
nikomatsakis: Something doesn't feel right to me here. I noticed in a few places that we were converting to/from `TerminalString` and `TerminalLiteral` more than seems necessary. I suspect there is some data structure that is defined in terms of `TerminalString` that actually should be defined in terms of `TerminalLiteral`.

lalrpop/src/lr1/build/test.rs
```diff
@@ -308,3 +308,26 @@ ty: () = {
     let _lr1_tls = Lr1Tls::install(grammar.terminals.clone());
     build_lr1_states(&grammar, nt("ForeignItem")).unwrap();
 }
+
+// Not sure if this is the right spot
```
nikomatsakis: I'd probably remove this in favor of the end-to-end tests. I tend to add things into the LR1 module if it is testing some quirky bit of the LR(1) algorithm in particular, but this is more testing that the tokenizer integrates properly with the LR(1) stuff.

lalrpop/src/lr1/example/test.rs
```diff
@@ -10,7 +10,7 @@ fn nt(t: &str) -> NonterminalString {
 }
 
 fn term(t: &str) -> TerminalString {
-    TerminalString::Literal(TerminalLiteral::Quoted(intern(t)))
+    TerminalString::quoted(intern(t))
```
nikomatsakis: 👍 to these changes

lalrpop/src/normalize/lower/mod.rs
```diff
@@ -78,16 +78,24 @@ impl<'s> LowerState<'s> {
...
+                                                                        kind: PatternKind::Usize(index),
+                                                                    },
+                                                                    Pattern {
+                                                                        span: span,
+                                                                        kind: PatternKind::Choose(input_str.clone())
+                                                                    }
+                                                                    ]),
                                                     };
+
+                                                    // FIXME: This should be cleaner
```
nikomatsakis: if we replaced `literals` with a `Vec` of structs, or something like that, then it would be nicer

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -20,29 +20,38 @@ use collections::{map, Map};
 mod test;
 
 pub fn validate(mut grammar: Grammar) -> NormResult<Grammar> {
-    let (has_enum_token, all_literals) = {
+    let (has_enum_token, all_literals, match_to_user_name_map) = {
         let opt_match_token = grammar.match_token();
 
-        let (match_mappings, match_catch_all) = if let Some(mt) = opt_match_token {
-            let mut ret = map();
+        let (match_to_user_name_map, user_name_to_match_map, match_catch_all) = if let Some(mt) = opt_match_token {
```
nikomatsakis: I still think this would be cleaner with one `Option<Struct>` then several individual variables.

nikomatsakis:
Great work @wagenet, thanks!

wagenet:
@nikomatsakis you're welcome! Thanks for your guidance :)

PR #207, opened 2017-03-28
Upgrade regex-syntax

wagenet:
This fixes a bug where some regexes are not correctly printed as string literals.


lalrpop/src/lexer/nfa/mod.rs
```diff
@@ -360,7 +360,15 @@ impl NFA {
                     self.push_edge(s0, Noop, target);
                 }
                 Ok(s0)
-            }
+            },
+
+            // FIXME: Figure out what these should do
+            Expr::AnyByte => panic!("Unsupported"),
+            Expr::AnyByteNoNL => panic!("Unsupported"),
+            Expr::WordBoundaryAscii => panic!("Unsupported"),
```
nikomatsakis: If you look up above, you'll see that there are existing candidates like `Err(Expr:WordBoundary)`, can you just move these `WordBoundary` cases up there? They return `Err(...)` to signal to the user that it is not supported.

lalrpop/src/lexer/nfa/mod.rs
```diff
@@ -360,7 +360,15 @@ impl NFA {
                     self.push_edge(s0, Noop, target);
                 }
                 Ok(s0)
-            }
+            },
+
+            // FIXME: Figure out what these should do
+            Expr::AnyByte => panic!("Unsupported"),
+            Expr::AnyByteNoNL => panic!("Unsupported"),
```
nikomatsakis: These can be merged with the `AnyChar` cases -- regex supports both unicode/non-unicode, but we only use the unicode cases right now; if we ever did support bytes, it would still be doing the right thing to merge them, I think

nikomatsakis:
Ignore my comments, I just pushed a commit that handled it.

PR #213, opened 2017-04-03
Report generation

wieczyk:
Currently lalrpop is able to produce very nice high-level errors. They are very useful especially when someone is interested in assistance with resolving conflicts. I believe that I am not only one user who would prefer to have also a brief summary of produced automaton and encountered conflicts. After many years of using bison and menhir I am used to have such summary and solve conflicts using mainly such brief report.

In future we could include a high-level description of conflicts in report file as additional section.

Usage
---------

If lalrpop is executed with `--report` option, then report will be generated for each input file. The report is written in file with `.report` extension in same directory where `.rs` file has been generated.

Example
------------

https://pastebin.com/raw/whUKg7My

Report has been generated for `calculator6.lalrpop` where `Tier` macro has been modified to produce shift/reduce conflicts. 

```diff
 Tier<Op,NextTier>: Box<Expr> = {                                                                                                                                                           
-    Tier<Op,NextTier> Op NextTier => Box::new(Expr::Op(<>)),                                                                                                                               
+    Tier<Op,NextTier> Op Tier<Op, NextTier> => Box::new(Expr::Op(<>)),                                                                                                                     
     NextTier                                                                                                                                                                               
};
```

As a result you are able to see how report looks and how states are annotated in case of conflicts.

### DIFF
--- a/lalrpop/src/lr1/report/mod.rs
+++ b/lalrpop/src/lr1/report/mod.rs
@@ -1,0 +1,337 @@
+use grammar::repr::*;
+use lr1::core::*;
+use std::io::{self, Write};
+use std::cmp::max;
+use collections::*;
+
+use super::lookahead::*;
+
+pub fn generate_report<'grammar, W : Write + 'grammar>
+    ( out:                  &'grammar mut W
+    , lr1result:            &LR1Result<'grammar>
+    ) -> io::Result<()>
+{
+    let mut generator = ReportGenerator::new(out);
+    generator.report_lr_table_construction(lr1result)
+}
+
+static INDENT_STRING : &'static str = "    ";
+
+struct ReportGenerator<'report, W>
+    where W: Write + 'report
+{
+    pub out: &'report mut W,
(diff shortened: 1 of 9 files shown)


lalrpop/src/lr1/report/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    pub out: &'report mut W,
+}
+
+type ConflictStateMap<'report, 'grammar,  L> = Map<StateIndex, Vec<&'report Conflict<'grammar, L>>>;
+
+impl <'report, W> ReportGenerator<'report, W>
+    where W: Write + 'report
+{
+    pub fn new
+        ( out:                  &'report mut W
```
nikomatsakis: can you run this file through `rustfmt`?

nikomatsakis:
never mind, ran it myself :)

wieczyk:
Oh I just come back from work and wanted to realize your request :D

PR #214, opened 2017-04-04
extend the tutorial to explain lexing vs parsing

nikomatsakis:
cc @wagenet -- sound good?


nikomatsakis:
My one concern is where this should go -- is tutorial 2b too early?

nikomatsakis:
Meh. I'll land it. I want to pull this into a book at some point and can re-visit question of lexer vs parser now.

PR #221, opened 2017-04-15
Restore Rust 1.13 compatibility

ruuda:
Lalrpop 0.13 regressed in Rust compatibility; 0.12.5 compiled fine on Rust 1.13 onwards, but 0.13 required at least Rust 1.16.

With a few small adjustments, Rust 1.13 compatibility can be restored.

It might also be a good idea to put the minimal supported Rust version in the CI configuration. That way breaking changes can be detected. If the advantages of using a newer Rust version outweigh breaking compatibility, then requiring a newer Rust version is fine, but at least breakages are explicit in this way, and they can be mentioned in the changelog.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - stable
+  - 1.16.0
   - beta
   - nightly
 script:
--- a/lalrpop-test/src/main.rs
+++ b/lalrpop-test/src/main.rs
@@ -417,7 +417,7 @@ fn generics_issue_104_test1() {
 }
 
 #[test]
-fn match_section() {
+fn test_match_section() {
     assert!(match_section::parse_Query("SELECT foo").is_ok());
     assert!(match_section::parse_Query("select foo").is_ok());
     assert!(match_section::parse_Query("INSERT foo").is_ok());
--- a/lalrpop/src/lr1/lane_table/construct/merge.rs
+++ b/lalrpop/src/lr1/lane_table/construct/merge.rs
@@ -1,6 +1,8 @@
-use super::*;
-
-use collections::Multimap;
+use collections::{Map, Multimap, Set};
+use ena::unify::UnificationTable;
+use lr1::core::{Action, LR1State, StateIndex};
+use lr1::lane_table::construct::state_set::StateSet;
+use lr1::lane_table::table::LaneTable;
 use lr1::lane_table::table::context_set::ContextSet;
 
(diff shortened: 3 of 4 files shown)


nikomatsakis:
So, in general I think that since the Rust project does not backport bug fixes, one ought to be using the latest stable release (though I do release that is a somewhat naive point-of-view). However, I'm curious how you landed on 1.13 as the version you would prefer LALRPOP to support (as opposed to, say, the current release, or 1.15, or whatever).

In particular, the fact that travis.yaml cannot use `--all` seems to me to be a reasonable reason for LALRPOP to require a higher version (since it's easy to forget to add tests there). Using `use super::*` is nice, but less important to me.

ruuda:
> However, I'm curious how you landed on 1.13 as the version you would prefer LALRPOP to support.

It is the first Rust version that supports `?`, which was a major feature that I actually use, and I was already using Rust 1.13 for a different project, so I also picked it for my project that uses LALRPOP. It is not a very strong reason to use 1.13 in particular, and I don’t have a strong reason for sticking with it, besides not wanting `cargo build` to suddenly fail.

But I am not always on the latest stable Rust, because I don’t update things arbitrarily. It is my experience that upgrades often break things, even when compatibility is promised. I only upgrade when I have the time to do manual intervention if required, and I do read changelogs beforehand. (I am not picking on Rust in particular here, stable Rust actually has never caused trouble for me.) 

> In particular, the fact that travis.yaml cannot use `--all` seems to me to be a reasonable reason for LALRPOP to require a higher version (since it's easy to forget to add tests there).

That is a valid tradeoff to make, and I am not opposed to requiring Rust 1.16. But I think it is still valuable to put *a* fixed version in the CI config to detect when the minimal required version changes, if only to explicitly mention that in the changelog. I’ve had issues in the past where a library of mine that used to compile, did not compile six months later, without the code having changed, because a dependency that did a minor version bump required a newer Rust version. (There was no `Cargo.lock`.)

ruuda:
I reverted the `cargo test --all` changes and changed the minimal version to 1.16.

nikomatsakis:
> That is a valid tradeoff to make, and I am not opposed to requiring Rust 1.16. But I think it is still valuable to put a fixed version in the CI config to detect when the minimal required version changes, if only to explicitly mention that in the changelog. 

Sounds good. I am inclined to agree that it makes sense to at least *be aware* when you now require a higher version number. Thanks for the PR. =)

nikomatsakis:
@ruuda thanks :)

PR #222, opened 2017-04-17
Make action and state functions private.

ravenexp:
Generated recursive ascent parser contains public functions
for every state and every action.
Only the top level state function(s) must be public.


nikomatsakis:
Thanks!

PR #223, opened 2017-04-17
Add user uses in generated mod intern_token

### LINKED ISSUE #113: Cannot use custom types for custom errors
Following the example from https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop-test/src/error.lalrpop I tried changing the Error declaration to type Error = SpdxParserError<'input>; where SpdxParserError is imported with "use errors::SpdxParserError;"

This generates a compilation error in the declaration for

```
    impl<'input> Iterator for __Matcher<'input> {
        type Item = Result<(usize, (usize, &'input str), usize), __ParseError<usize,(usize, &'input str),SpdxParserError<'input>>>;
```

I think in mod __intern_token the use declarations that I added to the grammar are not being included. I am very new to rust so apologies if I got something wrong.

minijackson:
This allows custom error types to be used as User ParseError

This should fix #113

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ doc/pascal/lalrpop/src/pascal.rs
 doc/whitespace/src/parser.rs
 
 lalrpop-test/src/error.rs
+lalrpop-test/src/error_issue_113.rs
 lalrpop-test/src/error_recovery.rs
 lalrpop-test/src/error_recovery_pull_182.rs
 lalrpop-test/src/expr.rs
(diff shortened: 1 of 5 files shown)


minijackson:
Oh, before I forget, because this fix add *all* the uses defined in the grammar, this might (and most the time will) add warnings of unused imports

minijackson:
The warnings should be fixed, now ;-)

nikomatsakis:
@minijackson thanks! :)

Issue #205, opened 2017-03-24
Change how lane-table uses unification variables


nikomatsakis:
Part of the lane-table algorithm is to merge sets of states based on their contexts. I implemented this by using the `ena` crate to create [unification variables](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/lr1/lane_table/construct/state_set.rs#L10-L12) and having the [value of those variables be a `ContextSet`](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/lr1/lane_table/construct/state_set.rs#L15). As a result, when two unification variables are merged, the ena crate will [invoke this `UnifyValue` routine](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/lr1/lane_table/construct/state_set.rs#L37-L44) which merges the context set -- and possibly fails to do so. This is pretty convenient, but `ena` was designed with the idea that the values of a unification variable are cheaply clonable (i.e., an integer), so I think it's fairly inefficient. 

In particular, I was thinking that we have [these special **beach-head** states](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/lr1/lane_table/construct/mod.rs#L144) identified in the algorithm, and that we always know that the "head" of any state will be one of these beach-heads. Moreover, we have a natural ordering on these beach-heads (the order in the vector) that we can use to identify the "canonical" instance. The idea of lane-table, essentially, is to merge these beach-heads into as few groups as we can.

i haven't worked out all the details, but basically:

- we only actually need unification variables for the "groups" themselves
- we start out with one group per beachhead
- their value should be an index into a vector of context sets
    - when merging, we probably want to prefer the higher index (as part of the algorithm below)

For each state we would have a vector mapping it to an associated unification variable `Group`:

```rust
groups: Vec<Option<Group>>, // one entry per state
contexts: Vec<ContextSet>, // one entry per **group**
```

When we start walking a beachhead, we would create a fresh inference variable (a `Group`) and set `contexts[beachhead] = group`. We would also push a new context set with the "row" for that beachhead. We will be mutating this context set **in place** as we go. If we find a state that has no group (ie., `group[state].is_none()`), we can merge its row-context into our context and set its group to ourselves. (If that merging fails, the grammar is not LR(1).)

[...]

PR #225, opened 2017-04-20
WIP: Issue205: change how lane table uses unification variables

wieczyk:
Solution to issue #205.

Please see comment on issue to see measurements.

### DIFF
--- a/lalrpop/src/lr1/lane_table/construct/group.rs
+++ b/lalrpop/src/lr1/lane_table/construct/group.rs
@@ -1,0 +1,103 @@
+use lr1::lane_table::table::context_set::{ContextSet, OverlappingLookahead};
+use lr1::core::*;
+use std::cmp::max;
+use ena::unify::{UnifyKey, UnifyValue, UnificationTable};
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Group {
+    index: usize
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct GroupUnifyKey {
+    index : u32
+}
+
+impl UnifyKey for GroupUnifyKey {
+    type Value = Group;
+
+    fn index(&self) -> u32 {
+        self.index
+    }
+
+    fn from_index(u: u32) -> Self {
+        GroupUnifyKey { index: u }
+    }
+
+    fn tag() -> &'static str {
+        "GroupUnifyKey"
+    }
+}
+
+impl UnifyValue for Group {
+    fn unify_values(value1: &Self, value2: &Self) -> Result<Self, (Self, Self)> {
+        // correctness of merging algorithm relies on order used here
+        Ok(max(*value1, *value2))
+    }
+}
+
+pub struct Groups<'m> {
+    // groups maps state index into group
+    groups:             Vec<Option<Group>>,
+    ///! maps group into unification key
+    context_sets:       Vec<ContextSet>,
+    unification_table:  &'m mut UnificationTable<GroupUnifyKey>,
+    ///! maps group into unification key
+    unification_keys:   Vec<GroupUnifyKey>,
+}
+
+impl<'m> Groups<'m> {
+    pub fn new<'a>(number_of_states : usize, unify: &'m mut UnificationTable<GroupUnifyKey>) -> Self {
+        let groups = vec![None; number_of_states];
+        Groups {
+            groups:             groups,
+            context_sets:       Vec::new(),
+            unification_keys:   Vec::new(),
+            unification_table:  unify,
+        }
+    }
+
+    pub fn allocate(&mut self, state: StateIndex, context_set : ContextSet) -> Group {
+        let group = Group { index : self.context_sets.len() };
+        self.context_sets.push(context_set);
+        self.groups[state.0] = Some(group);
+        self.unification_keys.push(self.unification_table.new_key(group));
+        group
+    }
+
+    pub fn has_group(&self, state : StateIndex) -> bool {
+        self.groups[state.0].is_some()
+    }
+
+    pub fn merge_state(&mut self, group: Group, state : StateIndex, context_set : &ContextSet)
+        -> Result<(), OverlappingLookahead>
+    {
+        self.groups[state.0] = Some(group);
+        self.context_set(group).inplace_union(context_set)
+    }
+
+    fn unify_key(&self, group: Group) -> GroupUnifyKey {
+        self.unification_keys[group.index]
+    }
+
+    fn context_set(&mut self, group: Group) -> &mut ContextSet {
+        let key = self.unify_key(group);
+        &mut self.context_sets[self.unification_table.probe_value(key).index]
+    }
+
+    pub fn merge_groups(&mut self, group1: Group, group2: Group) -> Result<(), OverlappingLookahead>{
+        let key1 = self.unify_key(group1);
+        let key2 = self.unify_key(group2);
+        self.unification_table.unify_var_var(key1, key2).is_ok();
+
+        let context_set = {
+            let context_set1 = &self.context_sets[group1.index];
+            let context_set2 = &self.context_sets[group2.index];
+            ContextSet::union(context_set1, context_set2)?
+        };
+
+        self.context_sets[max(group1, group2).index] = context_set;
+        Ok(())
+    }
+
+}
--- a/lalrpop/src/lr1/lane_table/construct/group.rs
+++ b/lalrpop/src/lr1/lane_table/construct/group.rs
@@ -67,7 +67,7 @@ impl Groups {
 
     pub fn allocate(&mut self, state: StateIndex, context_set : ContextSet) -> Group {
         let group = Group { index : self.context_sets.len() };
-        debug!("Allocated group={:?} for state={:?} context_set={:?}", group, state, context_set);
+        debug!("Groups::allocate: allocated group={:?} for state={:?} context_set={:?}", group, state, context_set);
         self.context_sets.push(context_set);
         self.groups[state.0] = Some(group);
         self.unification_keys.push(self.unification_table.new_key(group));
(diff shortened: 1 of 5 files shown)


lalrpop/src/lr1/lane_table/construct/group.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    pub fn add_state(&mut self, state : StateIndex) {
+        assert!(state.0 == self.groups.len());
+        self.groups.push(None)
+    }
+
+    pub fn merge_state(&mut self, group: Group, state : StateIndex, context_set_opt : Option<&ContextSet>)
+        -> bool
+    {
+        self.groups[state.0] = Some(group);
```
wieczyk: It should be later, where `Err` cannot occur.

lalrpop/src/lr1/lane_table/construct/group.rs
```diff
@@ -0,0 +1,165 @@
...
+        let key = self.unify_key(group);
+        &self.context_sets[self.unification_table.probe_value(key).index]
+    }
+
+    pub fn context_set_ref_for_state(&mut self, state: StateIndex) -> &ContextSet {
+        let group = self.groups[state.0].unwrap().clone();
+        self.context_set_ref(group)
+    }
+
+    pub fn unify_groups(&mut self, group1: Group, group2: Group) {
```
wieczyk: There is a place for micro optimization. I am always finding canonical elements for given groups. But we can skip one `UnificationTable::find` call since we know that actually the newest group will be a root group. You mention it in the issue #205 description that we can use this knowledge.

 But I believe skipping `find` on a group which is actually a root group will not give much here.

lalrpop/src/lr1/lane_table/construct/merge.rs
```diff
@@ -21,45 +20,46 @@ pub struct Merge<'m, 'grammar: 'm> {
...
 }
 
 impl<'m, 'grammar> Merge<'m, 'grammar> {
     pub fn new(table: &'m LaneTable<'grammar>,
-               unify: &'m mut UnificationTable<StateSet>,
                states: &'m mut Vec<LR1State<'grammar>>,
-               state_sets: &'m mut Map<StateIndex, StateSet>,
-               inconsistent_state: StateIndex)
+               inconsistent_state: StateIndex,
+               rows: Map<StateIndex, ContextSet>)
```
wieczyk: Converting it to normal table from `Map` looks like another place for micro optimization

lalrpop/src/lr1/lane_table/construct/group.rs
nikomatsakis: Yeah, skipping `find` isn't worth much. I *think* (but I'll have to re-read my comment) that I may have been thinking of not using union-find at all (or rather open-coding it, so to speak), in which case this fact may be of value. But let me read your patch in full to get my head in the right space.

lalrpop/src/lr1/lane_table/table/context_set/mod.rs
```diff
@@ -56,10 +56,25 @@ impl ContextSet {
 
     pub fn union(set1: &ContextSet, set2: &ContextSet) -> Result<Self, OverlappingLookahead> {
         let mut result = set1.clone();
+        result.inplace_union(set2)?;
+        Ok(result)
+    }
+
```
nikomatsakis: nit: 2 blank lines, should be just 1

lalrpop/src/lr1/lane_table/construct/merge.rs
```diff
@@ -75,78 +75,100 @@ impl<'m, 'grammar> Merge<'m, 'grammar> {
...
-                   state, successor);
+            
+            let successor_group_opt = self.groups.group(successor);
+
+            debug!("Merge::walk: state={:?} successor={:?} successor_group_opt={:?}",
+                   state, successor, successor_group_opt);
+            if let Some(successor_group) = successor_group_opt {
+                let mut successor = successor;
+                if successor_group != group {
+                    if (self.groups.merge_groups(group, successor_group)) {
```
nikomatsakis: Nit: no parens needed around if condition.

lalrpop/src/lr1/lane_table/construct/merge.rs
```diff
@@ -75,78 +75,100 @@ impl<'m, 'grammar> Merge<'m, 'grammar> {
...
+                    return Err((self.original_index(state), self.original_index(successor)))
                 }
             }
         }
-
         Ok(())
     }
 
+    fn split(&mut self, state: StateIndex, group: Group, successor: StateIndex) -> Result<StateIndex, (StateIndex, StateIndex)> {
+        // search for an existing clone with which we can merge
```
nikomatsakis: I don't know how much it matters, but I imagine it would make sense to first search for an existing clone in the current group, *then* do this logic.

lalrpop/src/lr1/lane_table/construct/group.rs
```diff
@@ -0,0 +1,165 @@
...
+
+impl UnifyValue for Group {
+    fn unify_values(value1: &Self, value2: &Self) -> Result<Self, (Self, Self)> {
+        // correctness of merging algorithm relies on order used here
+        Ok(max(*value1, *value2))
+    }
+}
+
+pub struct Groups {
+    ///! groups maps state index into group
```
nikomatsakis: Nit: `///`, not `///!` (`//!` attaches a comment the enclosing item, but that's not what you want here)

nikomatsakis: Nit: I'd probably run this file through rustfmt, also.

lalrpop/src/lr1/lane_table/construct/group.rs
nikomatsakis: Nit: should not be `pub`

lalrpop/src/lr1/lane_table/construct/group.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.context_set_ref(group)
+    }
+
+    pub fn unify_groups(&mut self, group1: Group, group2: Group) {
+        let key1 = self.unify_key(group1);
+        let key2 = self.unify_key(group2);
+        assert!(self.unification_table.unify_var_var(key1, key2).is_ok())
+    }
+
+    pub fn merge_groups(&mut self, group1: Group, group2: Group) -> bool {
```
nikomatsakis: Nit: maybe rename to `try_merge_groups` or something

lalrpop/src/lr1/lane_table/construct/group.rs
```diff
@@ -0,0 +1,165 @@
...
+        } else {
+            true
+        }
+    }
+
+    fn unify_key(&self, group: Group) -> GroupUnifyKey {
+        self.unification_keys[group.index]
+    }
+
+    fn context_set(&mut self, group: Group) -> ContextSet {
```
nikomatsakis: Nit: I'd call this `take_context_set`, to indicate that it *moves out*

lalrpop/src/lr1/lane_table/construct/group.rs
```diff
@@ -0,0 +1,165 @@
...
+                    self.context_sets[group2.index] = context_set2;
+                    debug!("Groups::merge_grops: merging groups {:?} and {:?} failed", group1, group2);
+                    return false
+                }
+            }
+        };
+
+        self.unify_groups(group1, group2);
+        debug!("Groups::merge_groups: successfull merge of groups {:?} and {:?} context_set={:?}", group1, group2, context_set);
+        self.context_sets[max(group1, group2).index] = context_set;
```
nikomatsakis: Hmm, instead of using `max`, couldn't we query to find the new root of `group1` (or `group2`)? All things being equal, it's usually better if you let the UF algorithm pick the new root on its own, so it can balance the trees.

nikomatsakis: More generally, I think I would eliminate the distinction between the `GroupUnifyKey` and `Group` and instead just have one, with this code maintaining the invariant that the context set is only populated for root keys.

(To that end, I might prefer to make the `context_sets: Vec<ContextSet>` into a `context_sets: Vec<Option<ContextSet>>`, although it'd be less efficient. In particular, storing `None` would allow us to detect cases where we *expected* a value to be present but it was not (whereas now we store an empty table, so we can't assert against that scenario).)

nikomatsakis:
OK, I did some measurements of my own. Indeed the gain in performance seems to be quite slight, or perhaps even non-existent. Similarly memory usage did not appear to be affected. What do you think @wieczyk -- is this way of structuring things cleaner than the old way, or more complex? Given that the runtime appears to be roughly the same, I guess we might as well optimize for readability.

I'll have to look through the code some more to decide, maybe apply some nits or small things. Based on line-number count, the older version was better. :)

wieczyk:
Take into account that my opinion is biased, since I spent some time here :)

I would stay with new code and go even further -- follow you suggestion about not using union-find at all -- it would allow to cut off one dependency here (ena). Let me also allow to apply all suggestions from review and do some small microoptimizations. Then with clean code you would be able to decide what do you prefer. 

I like to produce a lot of small functions -- I understand that it is not desirable everywhere and not everyone like it. If the number of lines is something what matters here then, during cleaning, I would try to do code a bit more concise.

About performance. Yeah, we did not achieve anything tangible but let's do measurements again after doing cleaning and applying micro optimizations somewhere. I did not perform any other measurements than running lalrpop on rustypop. Maybe this grammar was the best benchmark for those changes (silly guess). 

If I could suggest something then it would be to do more specific measurements. For example we could measure how many times we were cloning states. How much conflicting actions we had after LR(0) building and how much time we spent on each action. I would prefer to spent some effort on identifying bottle necks before deciding if we want to clean this code or revert it and clean the old one. I just believe that with more detailed measurements we will get better imagination which parts of those computations are heavy (or repeated often). Maybe we will get some idea for improvements which will be easier to do with current code.

What do you think?

wieczyk:
For example. Maybe if we would discover that cloning states (I mean running procedure for one particular state) is actually fast. Then we would know that it would be valuable to think about pre-computing something or sharing some results between those procedures (silly guess).

nikomatsakis:
@wieczyk I'm sort of inclined to go with the new path as well. It feels like there is more efficiency to be had, and this is clearly a step along the way, even if we're not the whole way there. Go ahead and apply the nits, I'll try to think if there are other things we can do to improve performance. I do agree it's worth doing some more measurements and profiling; I'd probably just use `perf` and `cachegrind` and try to see what's taking time. Though it must be said, overall performance is not bad -- might be interesting to focus instead on memory consumption.

I've mainly been interested in rustypop because it's the most complex grammar I have on hand. We should also test against the pascal grammar, which is non-trivial. 

PS I don't care about raw line count. =) Keep the small functions.

nikomatsakis:
OK, this has been hanging open for *far* too long. Given that @wieczyk and I agreed this seemed like a better approach, I'm inclined to merge.

nikomatsakis:
OK, I let this slip again. I will re-review to refresh my memory what the !@#! it does, then land.

nikomatsakis:
OK, it does seem to legitimately be breaking a test, will have to investigate.

PR #227, opened 2017-04-22
Remove dependencies on rustc-serialize.

withoutboats:
rustc-serialize is deprecated and is not actually used in lalrpop
right now anyway.

### DIFF
--- a/doc/pascal/lalrpop/Cargo.toml
+++ b/doc/pascal/lalrpop/Cargo.toml
@@ -9,7 +9,6 @@ path = "../../../lalrpop"
 
 [dependencies]
 docopt = "0.7"
-rustc-serialize = "0.3"
 regex = "0.2.1"
 
 [dependencies.lalrpop-util]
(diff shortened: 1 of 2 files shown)


withoutboats:
Ah damn, the main lalrpop does depend on it, I was just silly and i grepped for `rustc-serialize`.

I'm going to see about moving this to serde if there are no objections.

withoutboats:
This is actually docopt that depends on rustc-serialize, I've readded rustc-serialize to lalrpop's cargo.toml, other ones shouldn't need it though.

nikomatsakis:
@withoutboats did you forget to push? travis still seems unhappy

withoutboats:
I think I amended and force pushed and it didnt pick it up

nikomatsakis:
@withoutboats it looks like the problem was the doc/pascal library; I just pushed a fix, let's see what travis does with it.

nikomatsakis:
Thanks. =)

withoutboats:
No problem, I feel like I didn't accomplish much though since its still in your dependency graph through docopt.

PR #235, opened 2017-05-11
Added version information for lalrpop executable

jchlapinski:
this solves https://github.com/nikomatsakis/lalrpop/issues/224

### DIFF
--- a/lalrpop/src/main.rs
+++ b/lalrpop/src/main.rs
@@ -8,17 +8,25 @@ use std::env;
 use std::io::{self, Write};
 use std::process;
 
+static VERSION: &'static str = env!("CARGO_PKG_VERSION");
+
 fn main() {
     main1().unwrap();
 }
 
 fn main1() -> io::Result<()> {
     let mut stderr = std::io::stderr();
+    let mut stdout = std::io::stdout();
 
     let args: Args =
         Docopt::new(USAGE)
         .and_then(|d| d.argv(env::args()).decode())
         .unwrap_or_else(|e| e.exit());
+
+    if args.flag_version {
+        try!(writeln!(stdout, "{}", VERSION));
+        process::exit(0);
+    }
 
     let mut config = Configuration::new();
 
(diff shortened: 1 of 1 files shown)


lalrpop/src/main.rs
```diff
@@ -67,8 +75,10 @@ fn main1() -> io::Result<()> {
 const USAGE: &'static str = "
 Usage: lalrpop [options] <inputs>...
        lalrpop --help
+       lalrpop (-V | --version)
```
nikomatsakis: shouldn't this be `-v` (lowercsae)?

lalrpop/src/main.rs
```diff
@@ -8,18 +8,26 @@ use std::env;
...
+    let mut stdout = std::io::stdout();
 
     let args: Args =
         Docopt::new(USAGE)
         .and_then(|d| d.argv(env::args()).decode())
         .unwrap_or_else(|e| e.exit());
 
+    if args.flag_version {
+        try!(writeln!(stdout, "{}", VERSION));
+	process::exit(0);
```
nikomatsakis: Nit: indentation

lalrpop/src/main.rs
jchlapinski: I followed the same convention as rustc and cargo, Also it was suggested to be `-V` here https://github.com/nikomatsakis/lalrpop/issues/224#issue-222922059. 
But it can be lowercase, no matter to me.

nikomatsakis: oh, I guess `-v` is typically verbose

lalrpop/src/main.rs
```diff
@@ -67,8 +75,10 @@ fn main1() -> io::Result<()> {
 const USAGE: &'static str = "
 Usage: lalrpop [options] <inputs>...
        lalrpop --help
+       lalrpop (-v | --version)
 
 Options:
+    -v, --version        Print version.
```
nikomatsakis: ok well sorry :) let's use `-V` after all if that's the typical convention

nikomatsakis:
Sorry for letting this sit so long!

Issue #115, opened 2016-06-10
Allow #![...] in lalrpop files


blt:
It would be very useful to be able to specify `#![allow(clippy)]` ahead of the `grammar;` in a parser generation file. lalrpop generated rust is noisy when you run clippy against it. The accepted method in clippy to ignore source files is [annotation](https://github.com/Manishearth/rust-clippy/issues/702) but this manner of annotation is not valid in an lalrpop file.

polybuildr:
I think you can put a `#[allow(clippy)]` (missing `!`) or a `#[cfg_attr(feature = "cargo-clippy", allow(clippy))]` just above the `mod _;` in main.rs or wherever and that solves the problem.

nikomatsakis:
Yes, @polybuildr is right, but in general I agree it'd be nice to permit `#![...]` attributes in grammar files.

nikomatsakis:
cc @wieczyk -- maybe you would be interested in fixing this up.

wieczyk:
As rust beginner I need a small clarification here. As far as I understand lalrpop's annotation start with `#[` while rust compiler's annotation start with `#![`? 

Are we interested in having sophisticated strategy about handling rust annotations? Would it be OK to forward them as text?

polybuildr:
@wieczyk you can find more details at https://doc.rust-lang.org/book/attributes.html

From that link:
> The #[foo] attribute applies to the next item, which is the struct declaration. The #![bar] attribute applies to the item enclosing it, which is the mod declaration.

wieczyk:
I will allow to write `#![...]` annotations between `use`s and `"grammar"` annotations.

Implementation detail:  I am going to introduce special token which remembers such annotation as text. Let's allow rustc compiler to verify rust annotations.

nikomatsakis:
I think normally `#!` must precede a `use` declaration in Rust code.

ahmedcharles:
I believe this should be closed now?

PR #236, opened 2017-05-14
Issue115 allow shebang in lalrpop files

wieczyk:
Solution for #115


nikomatsakis:
Sorry for letting this sit so long!

Issue #243, opened 2017-06-27
outdated `foo.rs` file may fail to be rebuilt


nikomatsakis:
If you have a `foo.lalrpop` file that has not changed, but you bump your version of *lalrpop*, then we can often fail to rebuild `foo.rs`. We should embed the version of lalrpop in the rs file and check that it's still the same as the current executing version.

I'm marking this as good-first-bug but I'll try to leave some directions later for how to go about it!

nikomatsakis:
Fixed by #245, thanks @mchesser !

PR #245, opened 2017-07-01
Embed LALRPOP version in generated files, and check it on rebuild

mchesser:
Addresses #243

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -18,7 +18,7 @@ use tls::Tls;
 use tok;
 
 use std::fs;
-use std::io::{self, Write};
+use std::io::{self, Write, BufRead};
 use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::rc::Rc;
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Oh, hey, nice! Sorry I've been slow on the reviews! I've been traveling. (And indeed have some more travels coming up...)

lalrpop/src/build/mod.rs
```diff
@@ -157,6 +165,14 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
                                   -> bool {
         true
     }
+
+    fn compare_lalrpop_version(rs_file: &Path) -> io::Result<bool> {
+        let mut input_str = String::new();
+        let mut f = io::BufReader::new(try!(fs::File::open(&rs_file)));
+        try!(f.read_line(&mut input_str));
+
+        Ok(input_str.trim() != LALRPOP_VERSION_HEADER)
```
nikomatsakis: Very nice.

I wonder if there is some way we can test this... well, we're not testing it now anyway I guess. We should think about it for the future though!

nikomatsakis:
@mchesser I presume you've tested this locally and made sure it does indeed prevent "recompilation" most of the time? (I can't easily test myself just now...)

mchesser:
Yeah, I tested it with the CLI and with a project that uses it as a build dependency and there didn't appear to be any issues.

nikomatsakis:
Thanks! :)

PR #246, opened 2017-07-02
Ignore the possible expected token when we find an unrecognized token

fitzgen:
This will at least print an error message with the location, rather than
panicking, which is a lot more user-friendly :)

r? @nikomatsakis


lalrpop/src/build/mod.rs
```diff
@@ -209,7 +209,7 @@ fn parse_and_normalize_grammar(session: &Session, file_text: &FileText) -> io::R
         }
 
         Err(ParseError::UnrecognizedToken { token: Some((lo, _, hi)), expected }) => {
-            assert!(expected.is_empty()); // didn't implement this yet :)
+            let _ = expected; // didn't implement this yet :)
```
nikomatsakis: actually, this DID get implemented I think! You can just remove the whole line.

PR #254, opened 2017-08-09
Make ParseError Displayable

joerivanruth:
Because the types `(usize, &str)` and `()` do not implement `fmt::Display`, `lalrpop_util::ParseError<usize, (usize, &str),()>` doesn't either, which is a nuisance when displaying an error message.

I added three methods to ParseError that can be used to transform these types into more convenient types. They are of the form `fn map_loc<F,LL>(self, op: F) -> ParseError<LL, T, E> where F: Fn(L) -> LL`.

One use case is formatting, as above. Another is switching the `usize` byte offset locations into for example line/column numbers.

Although it is less pure, for user convenience it might be a good idea to use `String` rather than `()` as the default user-error type because strings have `fmt::Display`. I have not implemented this. Additionally, it might be an idea to add specific `Display` and `Error` impls for `ParseError<L,(usize,T),E>`. Together, these changes would make most ParseErrors printable, which seems desirable. I have not implemented this but will try if you'd like me to.

### DIFF
--- a/lalrpop-test/src/main.rs
+++ b/lalrpop-test/src/main.rs
@@ -159,7 +159,7 @@ fn expr_intern_tok_display_err() {
     let expr = "(1+\n(2++3))";
     let result = expr_intern_tok::parse_Expr(1, expr);
     let err : lalrpop_util::ParseError<usize, (usize, &str),()>
-            = result.expect_err("expected a syntax error");
+            = result.unwrap_err();
 
     // The problem is that (usize,&str) and () do not implement fmt::Display,
     // so neither does the ParseError.
--- a/lalrpop-test/src/main.rs
+++ b/lalrpop-test/src/main.rs
@@ -155,6 +155,26 @@ fn expr_intern_tok_test_err() {
 }
 
 #[test]
+fn expr_intern_tok_display_err() {
+    let expr = "(1+\n(2++3))";
+    let result = expr_intern_tok::parse_Expr(1, expr);
+    let err : lalrpop_util::ParseError<usize, (usize, &str),()>
+            = result.expect_err("expected a syntax error");
+
+    // The problem is that (usize,&str) and () do not implement fmt::Display,
+    // so neither does the ParseError.
+    // We can fix that by rewriting them to something else, such as a string.
+    let disp = err.map_tok(|(_,t)|t).map_err(|_| "error");
+    assert!(disp.to_string().contains("Unrecognized token +"));
+
+    // We can even convert the locations to a line number
+    let line_based = disp.map_loc(|offset| {
+        format!("line {}", expr[0..offset].lines().count())
+    });
+    assert!(line_based.to_string().contains("line 2"));
+}
+
+#[test]
 fn expr_lifetime_tok1() {
     // the problem here was that we were improperly pruning the 'input from the
     let tokens = lifetime_tok_lib::lt_tokenize("x");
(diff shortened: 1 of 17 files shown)


nikomatsakis:
> Although it is less pure, for user convenience it might be a good idea to use String rather than () as the default user-error type because strings have fmt::Display. I have not implemented this. Additionally, it might be an idea to add specific Display and Error impls for ParseError<L,(usize,T),E>. Together, these changes would make most ParseErrors printable, which seems desirable. I have not implemented this but will try if you'd like me to.

Seems like a good idea! It is annoying that errors cannot be easily printed.

lalrpop-util/src/lib.rs
```diff
@@ -36,6 +36,34 @@ pub enum ParseError<L,T,E> {
...
+            ParseError::ExtraToken { token } => ParseError::ExtraToken { token: maptok(token) },
+            ParseError::User { error } => ParseError::User { error: err_op(error) }
+        }
+    }
+
+    pub fn map_loc<F,LL>(self, op: F) -> ParseError<LL, T, E> where F: Fn(L) -> LL {
+        self.map_intern(op, |x|x, |x|x)
+    }
+
+    pub fn map_tok<F,TT>(self, op: F) -> ParseError<L, TT, E> where F: Fn(T) -> TT {
```
nikomatsakis: can we call this `map_token` instead?

lalrpop-test/src/main.rs
```diff
@@ -142,6 +142,26 @@ fn expr_intern_tok_test_err() {
...
+    let expr = "(1+\n(2++3))";
+    let result = expr_intern_tok::parse_Expr(1, expr);
+    let err : lalrpop_util::ParseError<usize, (usize, &str),()>
+            = result.unwrap_err();
+
+    // The problem is that (usize,&str) and () do not implement fmt::Display,
+    // so neither does the ParseError.
+    // We can fix that by rewriting them to something that has fmt::Display
+    let disp = err.map_tok(|(_,t)|t).map_err(|_| "error");
+    assert!(disp.to_string().contains("Unrecognized token +"));
```
nikomatsakis: It's not due to your code, but could we changed the message here to use backticks to "quote" the token? i.e.:

```
Unrecognized token `+`
```

joerivanruth:
Here's the minor nits, I'll look at the display-ability things later

joerivanruth:
I changed the error type to `&'static str`. It turns out to be impossible to add another Display impl for
`ParseError<L, (usize,T), E)>`. The compiler finds it ambiguous even though it isn't in this case as `(usize,_)` does not implement Display.

The only solution I see is to change the internal tokenizer's token type to something other than a tuple so we can add `#[derive(Display)]` to it.

joerivanruth:
Also turns out I haven't been running all the tests. I'll go and fix the errors.

joerivanruth:
Update: after `grep '\.rs$' .gitignore| xargs rm -f; cargo clean; rm Cargo.lock` the failing test started to show up locally. Fixed it. I hope.

joerivanruth:
I changed the internal lexer token from `(usize, &str)` to a new `struct InternalToken<'a>` which implements `Display`. The internal token is currently in `lalrpop-util`, I'll move it to the generated code where it will probably feel more at home. Apart from that I hope you're ok with the changes. Shall I remove the merge and rebase instead?

nikomatsakis:
@joerivanruth yes, that seems good! thanks!

joerivanruth:
Make the changes. When the the internal tokenizer is enabled, a `Token` type appears at the toplevel of the generated module. This is useful because it appears in types exported by that module, such as the ParseError.

Let me know what you think.

Joeri

Issue #247, opened 2017-07-03
Grammars using generics' associated types generate code that fails to compile


fitzgen:
# TLDR

Creating a grammar that is parameterized by a generic type that is bounded by a trait with associated types that are used as the return values of actions results in compilation failure of the generated code.

For example:

```rust
use std::str::FromStr;
use traits;

grammar<P>(callbacks: &mut P) where P: traits::ParseCallbacks;

pub Term: P::Term = {
    <n:Num> => n.into(),
    "(" <t:Term> ")" => t,
};

Num: P::Num = <s:r"[0-9]+"> => callbacks.number(i32::from_str(s).unwrap());
```

Where `traits::ParseCallbacks` is defined like this:

```rust
pub trait ParseCallbacks {
    type Term: From<Self::Num>;
    type Num;

    fn number(&mut self, i32) -> Self::Num;
}
```

## Steps to Reproduce

I've created a test case at https://github.com/fitzgen/lalrpop-generic-grammar-bug

```
$ git clone https://github.com/fitzgen/lalrpop-generic-grammar-bug.git`
$ cd lalrpop-generic-grammar-bug
$ cargo build
```

## Expected

Builds A-OK.

## Actual

Type errors in the generated parser code involving the grammar's generic type parameter's trait's associated types:

<details><pre><code>
$ cargo test
   Compiling lalrpop-generic-grammar v0.1.0 (file:///Users/fitzgen/src/lalrpop-generic-grammar)
error[E0433]: failed to resolve. Use of undeclared type or module `P`
  --> src/grammar.rs:16:15
   |
16 |         NtNum(P::Num),
   |               ^^^^^^ Use of undeclared type or module `P`

error[E0433]: failed to resolve. Use of undeclared type or module `P`
  --> src/grammar.rs:17:16
   |
17 |         NtTerm(P::Term),
   |                ^^^^^^^ Use of undeclared type or module `P`

error[E0433]: failed to resolve. Use of undeclared type or module `P`
  --> src/grammar.rs:18:20
   |
18 |         Nt____Term(P::Term),
   |                    ^^^^^^^ Use of undeclared type or module `P`

error[E0433]: failed to resolve. Use of undeclared type or module `P`
   --> src/grammar.rs:262:18
    |
262 |     ) -> (usize, P::Num, usize) {
    |                  ^^^^^^ Use of undeclared type or module `P`

error[E0433]: failed to resolve. Use of undeclared type or module `P`
[...]

PR #264, opened 2017-09-11
Create an explicit structure for `where` clauses

fitzgen:
This will make #247 easier to implement, so we can determine which type parameters are referenced in which where clauses.

r? @nikomatsakis

### DIFF
--- a/lalrpop/src/parser/test.rs
+++ b/lalrpop/src/parser/test.rs
@@ -94,3 +94,59 @@ fn match_complex() {
         _ => panic!("expected MatchToken, but was: {:?}", first_item)
     }
 }
+
+#[test]
+fn where_clauses() {
+    let clauses = vec![
+        "where T: Debug",
+        "where T: Debug + Display",
+        "where T: std::ops::Add<usize>",
+        "where T: IntoIterator<Item = usize>",
+        "where T: 'a",
+        "where 'a: 'b",
+        "where for<'a> &'a T: Debug",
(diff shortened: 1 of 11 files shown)


fitzgen:
arg I was only running the `larlpop-test` test >.<

fitzgen:
Note to self: need to support both `for<'a> &'a T: Whatever` and `T: for<'a> Fn('a)`

fitzgen:
Fixed tests; haven't added `for<'a>` support.

fitzgen:
r? @nikomatsakis 

I've got `for<'a>` stuff working now, also added a few tokenizer, parser, and integration tests.

lalrpop/src/parser/test.rs
```diff
@@ -94,3 +94,59 @@ fn match_complex() {
...
+        "where T: IntoIterator<Item = usize>",
+        "where T: 'a",
+        "where 'a: 'b",
+        "where for<'a> &'a T: Debug",
+        "where T: for<'a> Flobbles<'a>",
+        "where T: FnMut(usize)",
+        "where T: FnMut(usize, bool)",
+        "where T: FnMut() -> bool",
+        "where T: for<'a> FnMut(&'a usize)",
+        "where T: Debug, U: Display",
```
nikomatsakis: nice

nikomatsakis:
Sorry it took me a while. This is really well done.

PR #274, opened 2017-10-29
Add example with funky expression and tupels

### LINKED ISSUE #83: A rule with the "funky expression" <> doesn't work
The following grammar is accepted (with v0.9.0) and leads to a parser with a syntax error.

```
grammar;
pub Component : (&'input str,&'input str) = Number Number => <>;
Number = r"[0-9]+";
```

I was expected the same result than with
`pub Component : (&'input str,&'input str) = <a:Number> <b:Number> => (a,b);`
but maybe I misinterpreted the tutorial document... In any case I don't think the generation of a erroneous rust source code is normal.

pyfisch:
closes #83


nikomatsakis:
Thanks! ❤️

PR #279, opened 2017-11-22
Return grammar errors first

### LINKED ISSUE #278: Regression in error handling
There is a regression in the error handling which probably got introduced by 920ee76fec006cdfd8c86497d639a2bcbac34732 (#258).

With this grammar:

```rust
use lalrpop_util::ParseError;

grammar;

extern {
	type Error = &'static str;
}

pub Value = <Int> "abc";

Int: u64 = r"\d+" =>? Err(ParseError::User{ error: "Pretend there was an error" });
```

The out put of:

```rust
println!("{:?}", customerror::parse_Value("123"));
println!("{:?}", customerror::parse_Value("123 abc"));
```

is currently:

```rust
Err(User { error: "Pretend there was an error" })
Err(ExtraToken { token: (4, (1, "abc"), 7) })
```

But used to be (what I'd expect it to be):

```rust
Err(User { error: "Pretend there was an error" })
Err(User { error: "Pretend there was an error" })
```

vmx:
If some production in the grammar returns an error, give it priority
over built-in errors.

Fixes #278.

### DIFF
--- a/lalrpop-test/src/error_issue_278.lalrpop
+++ b/lalrpop-test/src/error_issue_278.lalrpop
@@ -1,0 +1,11 @@
+use lalrpop_util::ParseError;
+
+grammar;
+
+extern {
+	type Error = &'static str;
+}
+
+pub Value = <Int> "abc";
+
+Int: u64 = r"\d+" =>? Err(ParseError::User{ error: "Pretend there was an error" });
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ doc/whitespace/src/parser.rs
 
 lalrpop-test/src/error.rs
 lalrpop-test/src/error_issue_113.rs
+lalrpop-test/src/error_issue_278.rs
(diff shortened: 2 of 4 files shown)


lalrpop-test/src/error_issue_278.lalrpop
```diff
@@ -0,0 +1,11 @@
...
+
+grammar;
+
+extern {
+	type Error = &'static str;
+}
+
+pub Value = <Int> "abc";
+
+Int: u64 = r"\d+" =>? Err(ParseError::User{ error: "Pretend there was an error" });
```
brendanzab: Might want to set your editor to add new lines at the end of files - doesn't play nicely with diffs.

vmx: Fixed, thanks. I agree that it is annoying without a newline at the end of the file. The problem was that I didn't use an editor for that file.

PR #281, opened 2017-11-26
Specify `parsing` category in Cargo.toml

gwenn:
https://crates.io/categories/parsing
> Crates to help create parsers of binary and text formats.


nikomatsakis:
Thanks! Sorry for the shameful delay on my part. Rust impl period kind of dominated all my time. =)

PR #286, opened 2018-01-23
Rework error recovery mechanism

### LINKED ISSUE #240: Error recovery in 0.13 may end up in infinite loops
Tried to upgrade to 0.13.1 (from 0.12.5) but I have two tests for error recovery which now loops forever.

https://github.com/gluon-lang/gluon/blob/817d61eed462e27e196a7db6872ea57d801fd0ba/parser/tests/error_handling.rs#L64-L80

`cargo test --features test -p gluon_parser --test error_handling wrong_indent`

Looking at it in a debugger shows that after the `2` is received by lalrpop it will go into error recovery where after it will push a reduce action, evaluate it and then go right back into error recovery,

I am guessing this is to do with the lanetable implementation(?) breaking some assumption in the error recovery so I may may take a look over the weekend. I don't really have any guess as to why the lanetable would break it though so no promises that I know how to solve this!

nikomatsakis:
This change reworks LALRPOP's error recovery mechanism:

- Internally, we now generate a function (two, really) that gets invoked, rather than inlining the code twice.
- We now handle LALR and lane-table grammars correctly, in particular accounting for the fact that they can reduce several times before encountering an error (and hence we must trace through reductions when testing if we can recover from a particular state).
    - This avoids the infinite loops.
- We now use a modified algorithm that prefers to drop fewer tokens (but may drop more states as a result) -- [see description here](https://github.com/nikomatsakis/lalrpop/issues/240/#issuecomment-338483176).

I am not sure about the final bullet point. At first I thought it was clearly better, but I've become less sure. I'd be curious on feedback (cc @Marwes, do you have an opinion?). It should be relatively easy to change back, if we want to. (If you look at the new `error_recovery_span` test cases, you can see that I tried to enumerate some of the details of how the new error recovery can behave.)

To do list:

- [ ] Rework the big comment at the top of `parse_table.rs` to account for how it works now

Fixes #240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ lalrpop-test/src/error_issue_113.rs
 lalrpop-test/src/error_recovery.rs
 lalrpop-test/src/error_recovery_pull_182.rs
 lalrpop-test/src/error_recovery_issue_240.rs
+lalrpop-test/src/error_recovery_lock_in.rs
 lalrpop-test/src/expr.rs
 lalrpop-test/src/expr_arena.rs
 lalrpop-test/src/expr_generic.rs
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ lalrpop-test/src/error_recovery.rs
 lalrpop-test/src/error_recovery_pull_182.rs
 lalrpop-test/src/error_recovery_issue_240.rs
 lalrpop-test/src/error_recovery_lock_in.rs
+lalrpop-test/src/error_recovery_lalr_loop.rs
 lalrpop-test/src/expr.rs
 lalrpop-test/src/expr_arena.rs
 lalrpop-test/src/expr_generic.rs
--- a/.gitignore
+++ b/.gitignore
@@ -18,11 +18,13 @@ lalrpop-test/src/error.rs
 lalrpop-test/src/error_issue_113.rs
 lalrpop-test/src/error_recovery.rs
 lalrpop-test/src/error_recovery_pull_182.rs
+lalrpop-test/src/error_recovery_issue_240.rs
 lalrpop-test/src/expr.rs
 lalrpop-test/src/expr_arena.rs
 lalrpop-test/src/expr_generic.rs
 lalrpop-test/src/expr_intern_tok.rs
 lalrpop-test/src/expr_lalr.rs
+lalrpop-test/src/expr_module_attributes.rs
 lalrpop-test/src/generics_issue_104.rs
 lalrpop-test/src/inline.rs
 lalrpop-test/src/intern_tok.rs
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ lalrpop-test/src/error_recovery_pull_182.rs
 lalrpop-test/src/error_recovery_issue_240.rs
 lalrpop-test/src/error_recovery_lock_in.rs
 lalrpop-test/src/error_recovery_lalr_loop.rs
+lalrpop-test/src/error_recovery_span.rs
 lalrpop-test/src/expr.rs
 lalrpop-test/src/expr_arena.rs
 lalrpop-test/src/expr_generic.rs
--- a/lalrpop-test/src/associated_types.rs
+++ b/lalrpop-test/src/associated_types.rs
@@ -1,21 +1,25 @@
 // auto-generated: "lalrpop 0.13.1"
 use std::str::FromStr;
 use associated_types_lib::ParseCallbacks;
+#[allow(unused_extern_crates)]
 extern crate lalrpop_util as __lalrpop_util;
 
 mod __parse__Term {
-    #![allow(non_snake_case, non_camel_case_types, unused_mut, unused_variables, unused_imports)]
+    #![allow(non_snake_case, non_camel_case_types, unused_mut, unused_variables, unused_imports, unused_parens)]
 
     use std::str::FromStr;
     use associated_types_lib::ParseCallbacks;
+    #[allow(unused_extern_crates)]
     extern crate lalrpop_util as __lalrpop_util;
+    use super::__intern_token::Token;
+    #[allow(dead_code)]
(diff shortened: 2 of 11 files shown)


nikomatsakis:
One side note: it'd be really nice if, when we drop states, we could give the partially built things back to the user. But I'm not sure how to do that since they can have a variety of types and I don't want to expose the built-in `enum`.

nikomatsakis:
Weird. I don't see any test failures locally. Will have to investigate some more.

Marwes:
🎉 

> I am not sure about the final bullet point. At first I thought it was clearly better, but I've become less sure. I'd be curious on feedback (cc @Marwes, do you have an opinion?)

I will run through the [error handling tests](https://github.com/gluon-lang/gluon/blob/master/parser/tests/error_handling.rs) I have to see that nothing egregious happens (or at least that it can be fixed if it does). What I could see that may be a problem is that due to dropping more states causes less of the AST to constructed than the other approach.

```rust
// parse a tuple
(1, , 3)
//  ^ Oops

// Before
(1, <ERROR>, 3)
// After
<ERROR>

```
I'd have to test it to make sure though.

I will look through this PR more throughly later as well but that's my thoughts for now.

nikomatsakis:
OK, found the problem with the tests. I need to remember to `touch` the lalrpop files in the `doc` directory I guess. :/

nikomatsakis:
@Marwes 

> What I could see that may be a problem is that due to dropping more states causes less of the AST to constructed than the other approach.

Potentially, although I think I've not yet come up with good examples of this. It will also -- of course -- depend on where you put your recovery points. The newer algorithm, as it is biased to using the next token rather than dropping it, seems to be aimed at "incomplete expressions" (i.e., where you maybe have a terminator like `}`) whereas the older algorithm was aimed at "random junk inserted in the stream" (and hence it held onto state).

In any case, something like `(1, , 3)`, presuming your grammar is something like:

```
Expr = "(" Exprs ")"
    | !

Exprs = // comma sep list of Expr
```

ought to recover with `(1, !, 3)` -- i.e., it should try to keep the `,`, and hence insert a `!`. (I think both algorithms would behave the same here.)

An example where the newer algorithm would discard more would be something like `(1 + ) 2` -- well, I guess it will depend on your setup, but you can imagine something where the inner-most recovery is to keep the `1 +` and discard the `)`. The newer algorithm, in contrast, would unroll back up to the `(` (discarding the `1 + `) so that it can shift the `)`.

In general, if you are able to run this across your tests, that'd be great!

Marwes:
> Expr = "(" Exprs ")"
    | !

Yep that is basically the core of the error recovery I use and I'd hope that to handle things like `(1, , 3)` well, it was just the first example I could think.

nikomatsakis:
@Marwes OK well I'm gonna merge it but let me know how it goes =) can always change back

Marwes:
@nikomatsakis I forgot to check it yesterday but I put a reminder for it tonight.

Marwes:
Found a few changes when updating to 0.14.0.

First the error recovery appears to fire twice in one test causing duplicate errors to be generated https://github.com/Marwes/gluon/commit/e1087cb720e7418038e40b2575116161e19a1951 . This might not be a bug but I am going to take another look tomorrow to see if I can figure out the exact cause.

Second, it seems like `!` now gets the position assigned from the token before itself instead of the token after it which causes this
https://github.com/Marwes/gluon/commit/ee5f2fcf2975b40e561f5955c3cfd5023ace6bf5 case to get the `end` position from `|`. I were able to fix this by using `@R` on the right side of `!` instead so I think this new behavior makes more sense anyway.

I also needed do a small fix in layouting to prevent it from emitting an extra layout token at EOF. I believe I may have relied on LALRPOP ignoring extra tokens before which caused this to never manifest before now. It was easily fixed however https://github.com/Marwes/gluon/commit/0c3f1a15c0ffcbd4d109f7a2441a6f8835c75a58.

So best I can tell this looks great! 🎉

nikomatsakis:
@Marwes 

> This might not be a bug but I am going to take another look tomorrow to see if I can figure out the exact cause.

OK, let me know, I can also take a look if you want.

> Second, it seems like ! now gets the position assigned from the token before itself

I worked pretty hard on the spans for `!` -- you can see the `error_recovery_span` tests -- but I think it's kind of "doing the right thing now". In particular, if you do `@L ! @R` you should get a span that covers all dropped states + dropped tokens. If there are no dropped states or dropped tokens, that means there were two tokens adjacent that should have had some thing in between: e.g., `()` when it should not have been an empty list between the parens. In that case, you get a span starting at the end of `(` and before `)` (i.e., here, an empty span at a particular point).

Definitely let me know if you think it's wrong!

> So best I can tell this looks great! 🎉

Yay!

Marwes:
> I worked pretty hard on the spans for ! -- you can see the error_recovery_span tests -- but I think it's kind of "doing the right thing now".

:+1: I believe in my implementation just opted for setting the span to the token after `!` in case no tokens were dropped as that seemed simple and gave me the information I needed. But making it take the span between the token before and after is more consistent and I can still get the same information for the empty case (the end position of the token following `!` is not very interesting anyway).

> OK, let me know, I can also take a look if you want.

Will do but I want to check that it hasn't anything to do with the custom tokenizer I use first (the other bug was due to it so).

Marwes:
> First the error recovery appears to fire twice in one test causing duplicate errors to be generated Marwes/gluon@e1087cb . This might not be a bug but I am going to take another look tomorrow to see if I can figure out the exact cause.

So I am almost certain this is due to a quirk with gluon's layout engine which causes it to emit extra `CloseBlock` tokens on some parse errors. Before 0.14 this extra token were probably just dropped during recovery and so it never manifested. I opened an issue to track it in gluon but since it is such a benign error I am not going to investigate further at this time.

Nothing needs updating in LALRPOP in any case!

https://github.com/gluon-lang/gluon/issues/430

PR #293, opened 2018-01-26
Add support for pub(restricted).

### LINKED ISSUE #256: Support pub(restricted)
The lalrpop file parser allows a `fn` or a `pub fn`, but it does not allow a `pub(crate) fn`. In fact, using `pub(crate)` makes the processing of the lalrpop file panic with `thread 'main' panicked at 'assertion failed: expected.is_empty()`.

The reason I want `pub(crate)` to be supported is because I want to unit test each grammar rule independently. Rust allows unit-testing a private function only if the test function is in the same module, but with lalrpop there is no way to inject unit tests in the generated grammar module, as far as I know. So, I wrote my unit tests in a parent module, and I have to mark the rules `pub` so they can be visible. By doing so, I end up with a `dead_code` warning for each public rule when running a simple `cargo build`.

ahmedcharles:
Fixes #256.

This isn't ready yet.

@nikomatsakis Do you know where the `pub use` reexports of `parse_<nonterminal>` happen? I spent a few hours getting to this point, I figure asking might just be easier. :)

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -396,7 +396,8 @@ fn emit_recursive_ascent(session: &Session, grammar: &r::Grammar, report_file : 
         }
 
         rust!(rust,
-              "pub use self::{}parse{}::parse_{};",
+              "{}use self::{}parse{}::parse_{};",
+              grammar.nonterminals[&user_nt].visibility,
               grammar.prefix,
               start_nt,
               user_nt);
--- a/lalrpop/src/lr1/codegen/base.rs
+++ b/lalrpop/src/lr1/codegen/base.rs
@@ -135,6 +135,7 @@ impl<'codegen, 'grammar, W: Write, C> CodeGenerator<'codegen, 'grammar, W, C> {
 
         rust!(self.out, "#[allow(dead_code)]");
         try!(self.out.write_pub_fn_header(self.grammar,
+                                          &self.grammar.nonterminals[&self.start_symbol].visibility,
                                           format!("parse_{}", self.user_start_symbol),
                                           type_parameters,
                                           parameters,
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -3,6 +3,7 @@
 use collections::{Map, Set};
 use grammar::parse_tree::WhereClause;
 use grammar::repr::*;
+use intern::intern;
 use lr1::core::*;
 use lr1::lookahead::Token;
 use rust::RustWrite;
@@ -749,6 +750,7 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                               format!("_: {}", self.phantom_data_type())];
 
         try!(self.out.write_pub_fn_header(self.grammar,
+                                          &Visibility::Pub(Some(Path::from_id(intern("crate")))),
                                           format!("{}reduce", self.prefix),
                                           vec![],
                                           parameters,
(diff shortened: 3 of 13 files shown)


nikomatsakis:
> Do you know where the pub use reexports of parse_<nonterminal> happen?

Hmm, are these what you are looking for?

https://github.com/lalrpop/lalrpop/blob/9bd36041ad3e3cebcabe58b4be7cdd833d4878dd/lalrpop/src/build/mod.rs#L398-L402

ahmedcharles:
Yes, thanks.

lalrpop/src/build/mod.rs
```diff
@@ -395,11 +395,24 @@ fn emit_recursive_ascent(session: &Session, grammar: &r::Grammar, report_file :
...
+                      user_nt);
+            }
+            pt::Visibility::Pub(None) => {
+                rust!(rust,
+                      "pub  use self::{}parse{}::parse_{};",
+                      grammar.prefix,
+                      start_nt,
+                      user_nt);
+            }
+        }
```
ahmedcharles: The behavior of this section seems to just output `pub use` rather than `pub(crate) use`. Does anyone have any idea why? Did I get the pattern matching wrong?

lalrpop/src/lr1/codegen/base.rs
```diff
@@ -28,7 +28,7 @@ pub struct CodeGenerator<'codegen, 'grammar: 'codegen, W: Write + 'codegen, C> {
     /// the vector of states
     pub states: &'codegen [LR1State<'grammar>],
 
-    /// where we write output
+    /// where we wpublicrite output
```
Marwes: Spelling

lalrpop/src/build/mod.rs
Marwes: The `pub use` in question here seem to be emitted in `test_all.rs` https://github.com/lalrpop/lalrpop/blob/a527fa2ea232daf174370d4de90f3a49bfb8766c/lalrpop/src/lr1/codegen/test_all.rs#L58-L72

The reset of the `pub(crate)`s appear to be emitting correctly!

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -749,6 +749,7 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                               format!("_: {}", self.phantom_data_type())];
 
         try!(self.out.write_pub_fn_header(self.grammar,
+                                          &Visibility::Pub(None),
```
nikomatsakis: `pub(crate)` would suffice here, but I guess `pub` is fine too. In the future, it may start linting as a result though.

lalrpop/src/lr1/codegen/test_all.rs
```diff
@@ -54,11 +55,23 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TestAll> {
                                         this.states,
                                         "super::super::super",
                                         this.out));
-            rust!(this.out,
-                  "pub use self::{}parse{}::parse_{};",
-                  this.prefix,
-                  this.start_symbol,
-                  this.user_start_symbol);
+            let pub_use = match this.grammar.nonterminals[&this.user_start_symbol].visibility {
```
nikomatsakis: can we extract this into a helper function, like `emit_visibility` or something?

lalrpop/src/rust/mod.rs
```diff
@@ -106,14 +107,21 @@ impl<W:Write> RustWrite<W> {
...
                                where_clauses: Vec<String>)
                                -> io::Result<()>
     {
-        self.write_fn_header_helper(grammar, "pub ", name, type_parameters,
+        let path = match *visibility {
+            Visibility::Pub(ref path) => path.as_ref().map(|p| format!("pub({}) ", p)),
+            Visibility::Priv => panic!(),
+        };
+        self.write_fn_header_helper(grammar,
+                                    &path.as_ref().map(|p| p.as_str()).unwrap_or("pub "),
```
nikomatsakis: why don't we make `write_fn_header_helper` take the `Visibility` instead of a string?

PR #294, opened 2018-01-26
Add error when macros are marked public.

### LINKED ISSUE #262: Macro creates invalid mod name in generates Rust source
The following grammar when compiled with lalrpop 0.13.1:

```
grammar;

// Declarations
pub decs = decs_restrict<"F","T">;
pub decs_restrict<F,T>:() = {
    decs_restrict<F,T> "var",
    decs_restrict<"", T> "function" if F == "F",
    decs_restrict<F, ""> "type" if T == "T",
};
```

generates invalid code:
```
error: expected one of `;` or `{`, found `<`
   --> src/tiger.rs:509:27
    |
509 | mod __parse__decs_restrict<"", ""> {
    |                           ^ expected one of `;` or `{` here
```

ahmedcharles:
Fixes #262.

### DIFF
--- a/lalrpop/src/normalize/prevalidate/mod.rs
+++ b/lalrpop/src/normalize/prevalidate/mod.rs
@@ -124,6 +124,9 @@ impl<'grammar> Validator<'grammar> {
                     }
                 }
                 GrammarItem::Nonterminal(ref data) => {
+                    if data.public && !data.args.is_empty() {
+                        return_err!(data.span, "macros cannot be marked public");
+                    }
                     let inline_annotation = intern(INLINE);
                     let known_annotations = vec![inline_annotation];
                     let mut found_annotations = set();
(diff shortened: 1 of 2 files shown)


lalrpop/src/normalize/prevalidate/mod.rs
```diff
@@ -124,6 +124,9 @@ impl<'grammar> Validator<'grammar> {
                     }
                 }
                 GrammarItem::Nonterminal(ref data) => {
+                    if data.public && !data.args.is_empty() {
+                        return_err!(data.span, "macros cannot be marked public");
+                    }
```
nikomatsakis: Can we add a test for this?

Something like this

https://github.com/lalrpop/lalrpop/blob/9bd36041ad3e3cebcabe58b4be7cdd833d4878dd/lalrpop/src/normalize/prevalidate/test.rs#L10-L16

ahmedcharles: Sure. I wasn't sure how to add a negative test, but now I do.

Issue #280, opened 2017-11-26
Consider generating .rs files into OUT_DIR by default


Marwes:
This is coming from two angles.

First, cargocomb can't compile crates which generate files into the source directory https://github.com/rust-lang-nursery/crater/issues/157. While this is perhaps not really LALRPOPs problem, it may be something to consider. gluon couldn't compile with cargobomb which meant a recent improvement to rustc broke an assumption that I had (erroneously) which caused tests to segfault on nightlies with that change.

Second, with the grammar.rs file in the source tree it can be off-putting for new contributors looking for the parser to find an enormous, impenetrable file.

(While the first may not be LALRPOPs problem and the second more of a hypotetical, both of these problems were actually discovered when explaining how to contribute to gluon on a meetup so I wanted to take some steps to protect against this in the future).

Both of these are solved by hiding the file away in `OUT_DIR` instead which I now do explicitly with https://github.com/gluon-lang/gluon/pull/401 but it might be a good idea to make the default.

brendanzab:
For my meta language I'm definitely going to be output into `OUT_DIR` - it's more how cargo is meant to be used! There is a [`use_cargo_dir_conventions`](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/api/mod.rs#L64-L70) option in the `Configuration` struct, but perhaps that should be the default. 🤔

vmx:
I would expect that the generated file is checked into the repository. Else `lalrpop` is no longer a dev-dependency, but a normal dependency.

brendanzab:
I normally do:

```

[dependencies]
lalrpop-util = "0.13.1"

[build-dependencies]
lalrpop = "0.13.1"
```

build-dependency, not a dev-dependency

vmx:
@brendanzab You are right, I mixed that up (I do the same). This means that even if i check in the generated file, there would still be a dependency on lalrpop.

brendanzab:
That's correct, but it's not included in the binary. I'm not a fan of checking in generated code, so that's what I do.

vmx:
In the current default setup, even if you check in the generated file, you'd still have a dependency on lalrpop. This makes me change my mind, so I'm now also in favour of putting the generated files into the OUT_DIR by default (together with some documentation on how to deal with it).

brendanzab:
[...]

PR #296, opened 2018-01-26
Add rustfmt_skip attribute.

### LINKED ISSUE #233: Warn about rustfmt
To quote myself from rust-lang-nursery/rustfmt#1399 —

> The generated file is large and, relative to the style used by
> rustfmt, densely formatted, and has many long lines, at which rustfmt
> always emits error messages about being unable to shorten them.
> rustfmt does, however, manage to format much of the file — expanding
> it from 3010 lines to 8265.

I suggest that LALRPOP's documentation advise users to instruct
rustfmt to ignore LALRPOP's generated file, by putting
`#[cfg_attr(rustfmt, rustfmt_skip)]` on the `mod` item that references
the generated file. (Adding `#![cfg_attr(rustfmt, rustfmt_skip)]`
inside the module stops rustfmt from formatting the file, but not from
complaining about the long lines.)

(I'd offer a patch, but I don't know where it would be appropriate to
insert such a warning.)

ahmedcharles:
Fixes #233.

### DIFF
--- a/lalrpop/src/rust/mod.rs
+++ b/lalrpop/src/rust/mod.rs
@@ -178,6 +178,7 @@ impl<W:Write> RustWrite<W> {
     }
 
     pub fn write_module_attributes(&mut self, grammar: &Grammar) -> io::Result<()> {
+        rust!(self, "#![cfg_attr(rustfmt, rustfmt_skip)]");
         for attribute in grammar.module_attributes.iter() {
             rust!(self, "{}", attribute);
         }


lalrpop/src/rust/mod.rs
```diff
@@ -178,6 +178,7 @@ impl<W:Write> RustWrite<W> {
     }
 
     pub fn write_module_attributes(&mut self, grammar: &Grammar) -> io::Result<()> {
+        rust!(self, "#![cfg_attr(rustfmt, rustfmt_skip)]");
```
nikomatsakis: Hmm, it feels a bit odd to have this here... but reading the code I'm not sure where else it would go. There's always this tension about rust/mod.rs being sort of "general purpose" but also kinda' tailored to the job it's asked to do. Still, I might rather have this in the `emit_module_attributes` in `build/mod.rs` versus here.

ahmedcharles: Most of my time working on this was spent trying to find a place that would output the attribute in the correct location. :)

So, I don't really know the code base well enough to have a deeply informed opinion. There are two reasons to keep it here:

1. It's already tested. (It probably doesn't mean much given how much testing I did, but people often ignore this factor.)
2. I think every code generator for rust (assuming the rust module isn't for helping format rust code) would want to ignore rustfmt. That's not specific to lalrpop in any way.

nikomatsakis: OK, I don't feel that strongly either way.

nikomatsakis:
(It *would* be nice to test this, but it feels more annoying than it's worth really.)

ahmedcharles:
It is tested, in the sense that it's added unconditionally to every output file. Whether or not it stops rustfmt from formatting the module isn't tested, but that would indeed be annoying to test.

Marwes:
Unfortunately, this seems to have broken the usage of `include!` on the parser https://github.com/rust-lang/rust/issues/47995

cc #280

PR #297, opened 2018-01-26
Fix cargo build for lalrpop-test.

ahmedcharles:


Marwes:
I guess `lalrpop-test` isn't really mean to be `cargo build` since it only contains tests. Is there any reason why you want to `cargo build` it?

@nikomatsakis Why is `lalrpop-test` a separate crate in the first place and not just placed under `tests/` in the `lalrpop` crate?

ahmedcharles:
I don't want to build 'it' but `cargo build` in the root directory builds all crates, so all crates should build.

nikomatsakis:
@Marwes 

> Why is lalrpop-test a separate crate in the first place and not just placed under tests/ in the lalrpop crate?

To be 100% honest, when I started I don't think I knew about the `tests` directory. =)

Can you put tests in there that use `build.rs`? If so, I'd be in favor of refactoring things.

Marwes:
> Can you put tests in there that use build.rs? If so, I'd be in favor of refactoring things.

Bah, of course that won't work since we don't want to pollute the `lalrpop` crate's `build.rs` by having it compile test files. We could move these tests to `lalrpop-test` `tests/` directory though and just keep an empty `lib.rs/main.rs` in `lalrpop-test`. That way we don't need to suppress the warnings as this PR does and the tests for each `.lalrpop` file could get their own file perhaps which may reduce compile times a bit (when running with `--test <testname>`).

ahmedcharles:
That seems like a reasonable idea.

Issue #292, opened 2018-01-26
create a "LALRPOP manual"


nikomatsakis:
I think we should move LALRPOP documents into an [mdbook](https://github.com/rust-lang-nursery/mdBook).

Currently we have the tutorials etc living in `doc`. It'd be nicer to have an mdbook that auto-renders  to a github pages. Then we can also start adding LALRPOP internals in there in another chapter. I envision a table-of-contents sort of like:

- How to setup your project
    - roughly what the README has
- Tutorial
- Custom lexers
- Getting an error?
    - the error messages can give links in here
- Contributing and implementation details
    - high-level overview of how the code works and other stuff
    - the role of lalrpop-snap
    - maybe collect some of the material from my blog posts in some cases

nikomatsakis:
To get started, I'd say let's just rearrange the existing things into this structure.

udoprog:
Hey. I would like to do the initial breakout. If I don't get anything done until tonight I'd be happy if someone else took it over.

This is the first time I use mdbook (looks cool!). The example is easy enough to follow.

I'm going to try to break the existing tutorial sections into distinct pages. At the top there is a TOC, which fills the same purpose as a `SUMMARY.md`.
I'm also going to add a contributors section based of the existing contributors list on Github. It's nice to give folks some exposure on the manual page for doing good work :). I'll keep this as a separate commit if we don't want to do include this.

PR #298, opened 2018-01-26
Convert documentation to mdbook format

udoprog:
Fixes: #292

**Caveats**

Cross-linking is strange, as evidenced by the link to `lexer_tutorial/index.html` in `doc/src/tutorial/004_controlling_lexer.md`. Linking to `.md` does not translate the link to `.html`, and this project has one or two such cross links. These links also must be relative to the generated path. This has the unfortunate side-effect that links do not work as expected on GitHub. This also seems to be the case [on the Rust Book]. I'll check if a bug report exists on mdbook, otherwise I'll create one.
Edit: [#588: Expand links properly when cross linking to other `.md` files](https://github.com/rust-lang-nursery/mdBook/issues/588).

[on the Rust Book]: https://github.com/rust-lang/book/blame/master/second-edition/src/ch02-00-guessing-game-tutorial.md#L262

Titles were simplified to provide a better summary, e.g. "Parsing parenthesized numbers like `22` and `(22)` and `((22))`" -> "Parsing parenthesized numbers". This is to fit them into the sidebar better.

`Crash course on parsers` is given a dedicated section, the tutorial starts with "Adding LALRPOP to your project".

Tutorial sections are numbered to preserve fs ordering. This might be bad for direct linking in the future, so might be a bad idea. But when outlining the documentation in an editor it helps.

**Notes**

Build with:

```bash
cd doc
mdbook build
open book/index.html
```

When generating the documentation page, all other folders under `doc` must be copied separately to permit the links to work. Alternatively we could link to the master branch on GitHub to get better syntax highlighting.

Contributors added as a separate commit to make it easier to exclude if needed. Not everyone on that list has given permission to be listed on the documentation site, so it might be better to leave it empty initially. Please provide feedback.


Marwes:
Opened it up locally and clicked around a bit. Looks good though links do seem to be broken, not only the `.md` links but also links like `calculator/src/calculator1.lalrpop`. Not sure what is the best way to render those files now that direct links to the git repo won't work 🤔

udoprog:
Yeah, this was mentioned in the original description:
> When generating the documentation page, all other folders under doc must be copied separately to permit the links to work. Alternatively we could link to the master branch on GitHub to get better syntax highlighting.

Marwes:
@udoprog Ah, sorry it didn't connect for me.

I suppose the rust book do not use external sources like this but keep everything inline so there might not be a good solution. Moving everything here inline would mean the files couldn't be built and tested instead.

nikomatsakis:
Maybe we can add a script that wraps `mdbook build` which copies everything into the right place? It could also patch links to `.md` files to link to `.html` -- alternatively, we can just link directly to `.html` in the sources. We can then setup a github pages deployment thing.

udoprog:
@nikomatsakis mid term I'm hoping that https://github.com/rust-lang-nursery/mdBook/pull/589 can be converted into an acceptable state. I'm currently waiting for another round of feedback from the maintainer. This use-case is also pointing to that mdbook probably should be able to handle path translations for static links.

Until then, having paths relative and already translated from the output directory for links seems to be the way to go (e.g. `foo/bar.html`). If this looks otherwise good, I can introduce a build script that handles the rest.

nikomatsakis:
@udoprog 

> If this looks otherwise good, I can introduce a build script that handles the rest.

That would be awesome, yeah.

udoprog:
I adjusted the links I found; relative links and links to anchors (e.g. '#calculator2b`).

There's also a very simple build script now in the PR:
```bash
tools/build-doc
```

nikomatsakis:
@udoprog nice!

PR #299, opened 2018-01-31
Migrate from lalrpop-intern to string-cache.

### LINKED ISSUE #158: port to use `string_cache` instead of `lalrpop_intern`
It seems way better:

https://crates.io/crates/string_cache

ahmedcharles:
Fixes #158.

This probably isn't ready the way it is, but I'd like to get some initial feedback on it.

### DIFF
--- a/lalrpop/src/grammar/repr.rs
+++ b/lalrpop/src/grammar/repr.rs
@@ -4,7 +4,7 @@
  * representation incrementally.
  */
 
-use intern::{intern, InternedString};
+use string_cache::DefaultAtom as Atom;
 use grammar::pattern::{Pattern};
 use message::Content;
 use std::fmt::{Debug, Display, Formatter, Error};
@@ -100,7 +100,7 @@ pub enum LrCodeGeneration {
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Parameter {
-    pub name: InternedString,
+    pub name: Atom,
     pub ty: TypeRepr,
 }
 
@@ -114,7 +114,7 @@ pub struct Production {
     pub span: Span,
 }
 
-#[derive(Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub enum Symbol {
     Nonterminal(NonterminalString),
     Terminal(TerminalString),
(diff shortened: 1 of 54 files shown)


nikomatsakis:
@ahmedcharles nice! I've been wanting to do this for a while.

ahmedcharles:
I'm still trying to make the tests pass. And going from a `Copy` type to non-`Copy` is really annoying.

lalrpop/src/grammar/repr.rs
```diff
@@ -222,13 +222,13 @@ impl TypeRepr {
                               None => vec![],
                           })
                           .collect(),
-            TypeRepr::Associated { type_parameter, .. } =>
-                vec![TypeParameter::Id(type_parameter)],
-            TypeRepr::Lifetime(l) =>
-                vec![TypeParameter::Lifetime(l)],
+            TypeRepr::Associated { ref type_parameter, .. } =>
+                vec![TypeParameter::Id(type_parameter.clone())],
```
nikomatsakis: sad that `Atom` is not copy; but what you gonna' do

lalrpop/src/lr1/build/mod.rs
```diff
@@ -336,7 +336,7 @@ impl LookaheadBuild for Nil {
 
 impl LookaheadBuild for TokenSet {
     fn epsilon_moves<'grammar>(lr: &LR<'grammar, Self>,
-                               nt: NonterminalString,
+                               nt: &NonterminalString,
```
nikomatsakis: I definitely approve of this pattern; pass atoms by ref on the stack, clone to move into a struct

ahmedcharles: My approach so far is that if the function ever passes the type by reference, then it accepts a reference. If it only passes the type by value, then it accepts by value.

lalrpop/src/grammar/repr.rs
ahmedcharles: I think I agree with it not being copy. I'd probably have more conviction if I hadn't just changed all that code. :)

nikomatsakis: Well, the reason is that they maintain a reference count. Seems fine.

lalrpop/src/lr1/build/mod.rs
nikomatsakis: OK, that works. For things that are cheap to clone (e.g., Rc and Arc), I typically take a stronger rule:

- All variables on the stack are by reference
- All things in structs are by value

nikomatsakis: (But either one is fine.)

Issue #304, opened 2018-02-04
Further reduce the size of the parse tables


Marwes:
At a whim, I looked at how much memory the generated parser were taking up in the binary and it turns out it is pretty sizeable.

Gluon's grammar currently has ~320 states, ~150 symbols and ~30 terminals, which with some quick calculations gives a resulting size of ~0.25Mb.

```
__ACTION = 320 * 30
__GOTO = 320 * 150
BYTES = (__ACTION + __GOTO) * 4 bytes = 243200
```

On its own, this is not a huge part of gluons total ~3.5 Mb size but it is not a trivial part either.

So far I can think of three improvements towards this.

The easiest solution would be to just a use a smaller representation than `i32` as the state representation. With just 328 states it would easily fit in a `i16` instead. For really small grammars even an `i8` would be enough. This should be relatively easily fixed when generating the table and simply select the smallest integer which fits all states.

Another improvement which can help (if it is possible) some cases is to avoid generating unique tables for each `pub` item. Gluon exports [two](https://github.com/gluon-lang/gluon/blob/d01783e37e32a2a3315c410b79b17025b80771a2/parser/src/grammar.lalrpop#L589-L593) entrypoints and these each get their own unique tables despite being almost the same. I *think* it is possible to generate a single parser for both of these and just generate a small initializer for each which starts the parser of in the correct state after which everything should work as normal (please correct me if I am wrong about this @nikomatsakis ).

A third possibility is to use a different, more compact representation for the tables than the current straight array. This would likely incur a performance penalty so it might not be worth it.

nikomatsakis:
There is a lot of prior art on this. A very typical thing also is to compress the tables in some way. 

> I think it is possible to generate a single parser for both of these and just generate a small initializer for each which starts the parser of in the correct state after which everything should work as normal 

Quite possibly =) I've not given it much thought.

Marwes:
> There is a lot of prior art on this. A very typical thing also is to compress the tables in some way.

I tried to look into some prior art but I must be searching for the wrong terms since as I didn't find anything about it :/ . Should try a bit harder I guess.

I thought about mentioning compression as well but I feel that is cheating since it just serves to hide how much memory you actually need to use the parser. (+ it does not help to reduce size when distributing the library either since that will be compressed regardless).

ahmedcharles:
Technically, compression here could (probably does) mean a data layout in memory that is decompressed for every state.

[...]

PR #306, opened 2018-02-04
 perf: Use the minimum integer size necessary to represent states 

Marwes:
Part of #304 . We detect how many reductions and states that is needed before generating any code and then use the minimum necessary integer size to hold all of these. Most grammars likely fit within `i16` but smaller ones may fit within `i8`
.
The first commit simplifies the generated code a bit but did not help reduce the binary code in any way. Might make it a bit less work for LLVM at least.

### DIFF
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -1264,7 +1264,7 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
               "for {p}top in (0..{p}states_len).rev() {{",
               p = self.prefix);
         rust!(self.out,
-              "let {p}state = {p}states[{p}top];",
+              "let {p}state = {p}states[{p}top] as usize;",
               p = self.prefix);
         if DEBUG_PRINT {
             rust!(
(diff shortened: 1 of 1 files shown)


lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -332,6 +334,16 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                                                                       .zip(0..)
                                                                       .collect();
 
+        let state_type = {
+            if states.len() <= ::std::i8::MAX as usize && reduce_indices.len() <= (::std::i8::MAX as usize + 1) {
```
nikomatsakis: Nit: can we do:

```rust
let max_value = ::std::cmp::max(states.len(), reduce_indices.len() - 1);
if max_value <= ::std::i8::MAX { ... }
```

also, worth commenting where the `-1` comes from =)

ahmedcharles:
My only nit is that making the travis changes independently may be beneficial.

nikomatsakis:
> My only nit is that making the travis changes independently may be beneficial.

true...

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -332,6 +334,19 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                                                                       .zip(0..)
                                                                       .collect();
 
+        let state_type = {
+            // `reduce_indices` are allowed to be +1 since the negative maximum of any integer type
+            // is one larger than the positive maximum
+            let max_value = states.len().max(reduce_indices.len() - 1);
```
Dylan-DPC: you are using unstable library feature 'ord_max_min'

Marwes: Oops, didn't have time to run through the full tests, was sure that I had used this on stable already.

PR #310, opened 2018-02-05
move the setup instrutions out of README and into the book

nikomatsakis:
This simplifies the README overall and highlights the book more. Thoughts?

### DIFF
--- a/doc/src/advanced_setup.md
+++ b/doc/src/advanced_setup.md
@@ -1,0 +1,31 @@
+When you setup LALRPOP, you create a `build.rs` file that looks something
+like this:
+
+```rust
+extern crate lalrpop;
+
+fn main() {
+    lalrpop::process_root().unwrap();
+}
+```
+
+This `process_root()` call simply applies the default configuration:
+so it will transform `.lalrpop` files into `.rs` files *in-place* (in
+your `src` directory), and it will only do so if the `.lalrpop` file
+has actually changed. But you can also use the
+[`Configuration`][config] struct to get more detailed control.
+
+[config]: https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/api/mod.rs
+
+For example, to **force** the use of colors in the output (ignoring
+the TTY settings), you might make your `build.rs` file look like so:
+
+```rust
+extern crate lalrpop;
+
+fn main() {
+    Configuration::new()
+        .always_use_colors()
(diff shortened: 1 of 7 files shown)


doc/src/out_dir.md
```diff
@@ -0,0 +1,24 @@
...
+In addition, because the modules are generated out of the `src`
+directory, for each `foo.lalrpop` file you can't simply have `mod
+foo;` in your source.  Instead, you must put the following in the
+parent module:
+
+```rust
+include!(concat!(env!("OUT_DIR"), "/path/to/foo.rs"));
+```
+
+Here the `path/to/foo.rs` should be relative to the `src` directory.
```
nikomatsakis: cc @brendanzab -- is this detail about relative paths correct?

doc/src/out_dir.md
```diff
@@ -0,0 +1,24 @@
+One common request is to have LALRPOP generate its files into Cargo's
+**output directory**, rather than storing them 'in-place' within the
+module tree. This can be done through the following configuration:
```
nikomatsakis: I know we plan to make this the default. I figured I'd write up some instructions before hand.

doc/src/advanced_setup.md
```diff
@@ -0,0 +1,31 @@
...
+}
+```
+
+This `process_root()` call simply applies the default configuration:
+so it will transform `.lalrpop` files into `.rs` files *in-place* (in
+your `src` directory), and it will only do so if the `.lalrpop` file
+has actually changed. But you can also use the
+[`Configuration`][config] struct to get more detailed control.
+
+[config]: https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/api/mod.rs
```
nikomatsakis: Maybe we should publish the rustdocs too? Linking to docs.rs requires specifying a version, from what I could tell.

ahmedcharles: You can link to https://docs.rs/lalrpop/ but that just sends you to the crate docs. Linking below that requires a specific version.

Marwes: https://docs.rs/lalrpop/*/lalrpop/ will take you to the newest version of lalrpops docs (the "version" is actually understood as semver)

nikomatsakis: Ah, nice.

PR #311, opened 2018-02-07
Always use the 'input lifetime inside Symbol.

### LINKED ISSUE #188: Trying to parse the empty string results in code which does not compile.
`pub Empty: () = () => ();` results in:

```
error[E0277]: the trait bound `usize: std::ops::Add<()>` is not satisfied
  --> src/parser.rs:78:41
   |
78 |                 let __action = __ACTION[__state * 1 + __integer];
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::ops::Add<()>` is not implemented for `usize`
   |
   = help: the following implementations were found:
   = help:   <usize as std::ops::Add>
   = help:   <&'a usize as std::ops::Add<usize>>
   = help:   <usize as std::ops::Add<&'a usize>>
   = help:   <&'b usize as std::ops::Add<&'a usize>>

error: aborting due to previous error
```

ahmedcharles:
Fixes #188.

### DIFF
--- a/lalrpop-test/src/empty_tok.lalrpop
+++ b/lalrpop-test/src/empty_tok.lalrpop
@@ -1,0 +1,13 @@
+grammar;
+
+pub Empty: () = ();
+
+use empty_tok_lib;
+
+extern {
+    type Location = usize;
+    type Error = ();
+
+    enum empty_tok_lib::Tok {
+    }
+}
--- a/lalrpop-test/src/lib.rs
+++ b/lalrpop-test/src/lib.rs
@@ -22,6 +22,12 @@ lalrpop_mod!(sub);
 lalrpop_mod!(sub_ascent);
 lalrpop_mod!(sub_table);
 
+/// test matching empty string
+mod empty;
+
+/// test matching empty string
+mod empty_tok;
+
 /// more interesting demonstration of parsing full expressions
 lalrpop_mod!(expr);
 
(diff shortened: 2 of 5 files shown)


ahmedcharles:
This only works when using the default tokenizer.

@nikomatsakis Do you know how to determine if the default tokenizer is being used?

nikomatsakis:
@ahmedcharles 

> Do you know how to determine if the default tokenizer is being used?

`if self.grammar.intern_token.is_some() {` is the right way.

Can you add a test case of some kind though?

ahmedcharles:
I did add a test? Is something wrong with it?

nikomatsakis:
@ahmedcharles I must have missed it.

nikomatsakis:
Do you think it makes sense to have a test for the case where we are *not* using the default tokenizer, too? (Since this patch would otherwise break that case?)

ahmedcharles:
I tried to add another test case, but apparently it doesn't compile. Any ideas what I'm doing wrong?

lalrpop-test/src/empty_tok.lalrpop
```diff
@@ -0,0 +1,13 @@
+grammar;
+
+pub Empty: () = ();
+
+use empty_tok_lib;
+
+extern {
+    type Location = usize;
```
nikomatsakis: your test requires a location; so your tokenizer must supply `(usize, Tok, usize)`

lalrpop-test/src/lib.rs
```diff
@@ -723,3 +729,18 @@ fn error_issue_278() {
         }
     }
 }
+
+#[test]
+fn test_empty() {
+    empty::parse_Empty("").unwrap();
+}
+
+mod empty_tok_lib {
```
nikomatsakis: however, you could just use the `util` tokenizer that the rest of the tests use (see e.g. `sub_table` and so forth)

nikomatsakis:
I'm adding this to the 1.0 milestone -- I know it's been many years, but this still seems like a useful change. Will take a look once we get a maintenance release out the door.

Issue #220, opened 2017-04-15
Tag releases


ruuda:
There are currently two tags, but more `release-*` branches. When I wanted to build 0.13.0, a tag was the first thing I went looking for, but I could not find any. An annotated tag has a few advantages over a branch:

 * It is harder to change a tag after pushing it, than to make a branch point to a different commit.
 * Annotated tags include a release timestamp, which may be different from the commit time of the last commit.

nikomatsakis:
I sometimes tag, I sometimes forget. To be honest, I consider it an antiquated part of process, so I'm curious: Why would you want to check out the older release? In general, if you want the released version, why not just grab the version out of crates.io?

ruuda:
I wanted to make a local change on top of the released version. Doing a `git clone` was faster than figuring out the exact path where Cargo stores its crates. Doing a clone also has the benefit that you get a full repository, so `git diff` and `git grep` work immediately. In this particular case I did not have to do any code archaeology, but for other repositories I have had to do so.

Another thing that I use tags for regularly, is to check whether a particular commit (usually found via `git blame`) is included in a particular release.

brendanzab:
I use tags to provide a nice set of links in the changelog (see the links at the bottom): https://raw.githubusercontent.com/brendanzab/cgmath/master/CHANGELOG.md - this is based on the suggestion from http://keepachangelog.com/en/0.3.0/

nikomatsakis:
Seems good. I should probably add it to my publish script. To be honest, I still find making releases kind of awkward. I feel like I don't have things setup *quite* right.

brendanzab:
Completely agree. I find most of my releases awkward... something like https://github.com/semantic-release/semantic-release for the Rust ecosystem would be neat. Alas, no free time :(

Marwes:
https://github.com/sunng87/cargo-release might be good though last I checked it still had some things missing to make multi crate releases easy.

Dylan-DPC:
Can help with this. Can do it manually till we find a good way of automating it

nikomatsakis:
@Dylan-DPC that'd be great! I'd love in general to improve our release process. In any case, if you're up for going through and tagging releases, I'll add you as a maintainer in the meantime. =) I do try to make a PR or other things around the time of each release, so it *should* be pretty easy to find.

Even better, can you modify the `publish.sh` script to do it?

ahmedcharles:
[...]

PR #313, opened 2018-02-07
Add support for git tags when publishing.

ahmedcharles:
#220

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # A script to bump the version number on all Cargo.toml files etc in
 # an atomic fashion.
@@ -19,6 +19,7 @@ fi
 echo "Found consistent version $VERSION"
 
 BASE_DIR="$PWD"
+TMPDIR=${TMPDIR:-"/tmp"}
 
 function publish_fail {
     printf "ERROR\n"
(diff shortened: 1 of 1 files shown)


publish.sh
```diff
@@ -39,6 +40,9 @@ publish lalrpop-util
 publish lalrpop-snap
 publish lalrpop
 
+git tag $VERSION
```
ruuda: You should create an [annotated tag](https://git-scm.com/docs/git-tag#_description) with `-a`.

publish.sh
```diff
@@ -39,6 +40,9 @@ publish lalrpop-util
...
+git push origin $VERSION
```
ruuda: I think this should be [`git push origin tag $VERSION`](https://git-scm.com/docs/git-push#git-push-ltrefspecgt82308203). Alternatively you can `git push origin --tags`, but that will push *all* tags, which might not be desirable depending on your local repository state.

ahmedcharles: I think pushing only one is preferred. Added the `tag` option.

publish.sh
ahmedcharles: The current tags aren't annotated and they don't provide any benefit I can tell unless they are signed or given a message. The entire point of automating it is so Niko doesn't have to do manual steps and adding a default message seems pointless if it contains no more information than the tag itself.

ruuda: For some reason I was under the impression that lightweight tags cannot be pushed, but that is not true, they can. One small benefit of annotated tags is that `git describe` does not ignore them by default. `git describe` can be somewhat useful in automation, but as the version number in `Cargo.toml` needs to be updated anyway, I think you are right, there is little benefit in annotated tags for this use case.

PR #315, opened 2018-02-07
Add support for <mut x:X> allowing for mutable x in action code.

### LINKED ISSUE #136: [feature request] mark production variables as "mut"
Say I have a parser that returns a list of things, in reversed order. Currently I have to do this:

```
pub StmtList : Vec<Box<Stmt>> = {
    <stmts:RevStmtList> => {
        let mut stmts = stmts;
        stmts.reverse();
        stmts
    },
}
```

It'd be nice if I could mark `stmts` as `mut` instead. Then in the right-hand side I could skip the declaration.

```
pub StmtList : Vec<Box<Stmt>> = {
    <mut stmts:RevStmtList> => {
        stmts.reverse();
        stmts
    },
}
```

ahmedcharles:
Fixes #136.


ahmedcharles:
@nikomatsakis I think the reason this test is failing (normalize::macro_expand::test::test_comma) is because there's a sort somewhere that determines which macro expansions happen first. Is that plausible?

lalrpop/src/build/action.rs
```diff
@@ -92,7 +92,10 @@ fn emit_user_action_code<W: Write>(grammar: &r::Grammar,
                                                   .iter()
                                                   .cloned()
                                                   .map(|t| grammar.types.spanned_type(t)))
-                                         .map(|(p, t)| format!("(_, {}, _): {}", p, t))
+                                         .map(|(p, t)| format!("(_, {}{}, _): {}",
```
Marwes: Perhaps `((m, p), t)` would be a bit clearer?

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -447,8 +447,8 @@ pub enum SymbolKind {
     // <X>
     Choose(Box<Symbol>),
 
-    // x:X
-    Name(Atom, Box<Symbol>),
+    // <x:X> or <mut x:X>
+    Name(bool, Atom, Box<Symbol>),
```
Marwes: Perhaps an explicit enum would be nice here?

```
enum Mutability {
    Mutable,
    Immutable,
}
```

With a `fmt::Display` impl it could simplify some other code as well

ahmedcharles: I was thinking about that, but I wanted to get it working and get some feedback to see if I missed anything. But before I decide, I want to get the test working or at least understand why it's failing.

lalrpop/src/build/action.rs
ahmedcharles: I'm considering making a struct that contains the 'Pattern', with the name and whether it's mutable. I'd like some feedback on whether that would be useful or just confuse things more.

lalrpop/src/grammar/parse_tree.rs
nikomatsakis: 👍 to an enum

lalrpop/src/grammar/repr.rs
```diff
@@ -505,7 +505,7 @@ impl UserActionFnDefn {
                self.arg_patterns
                    .iter()
                    .zip(self.arg_types.iter())
-                   .map(|(p, t)| format!("{}: {}", p, t))
+                   .map(|(p, t)| format!("{}{}: {}", if p.0 { "mut " } else { "" }, p.1, t))
```
nikomatsakis: I like the idea of having `p` be a pattern, or something that implements `Display`, so that here we can just leave this code unchanged (also the spot above), and `p` will just emit `mut x`  or `x` as appropriate.

lalrpop/src/normalize/lower/mod.rs
```diff
@@ -327,7 +327,8 @@ impl<'s> LowerState<'s> {
             Symbols::Named(names) => {
                 // if there are named symbols, we want to give the
                 // arguments the names that the user gave them:
-                let arg_patterns = patterns(names.iter().map(|&(index, ref name, _)| (index, name.clone())),
+                let arg_patterns = patterns(names.iter()
```
nikomatsakis: Similarly, using some sort of "pattern" construct would hopefully mean that a lot of this code doesn't have to explicitly thread through "mutable" as a distinct thing.

lalrpop/src/normalize/macro_expand/test.rs
```diff
@@ -42,7 +42,7 @@ grammar;
 
     `(<"Id"> ",")+`: ::std::vec::Vec<#`(<"Id"> ",")`#> = {
         `(<"Id"> ",")` => vec![<>],
-        <v:`(<"Id"> ",")+`> <e:`(<"Id"> ",")`> => { let mut v = v; v.push(e); v },
+        <v:`(<"Id"> ",")+`> <e:`(<"Id"> ",")`> => { v.push(e); v },
```
nikomatsakis: Don't we need `mut v` here? What is going on?

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -170,8 +170,8 @@ ExprSymbol: ExprSymbol =
     Symbol* => ExprSymbol { symbols: <> };
 
 Symbol: Symbol = {
-    <lo:@L> "<" @L <l:Id> ":" <s:Symbol0> ">" <hi:@R> =>
-        Symbol::new(Span(lo, hi), SymbolKind::Name(l, Box::new(s))),
+    <lo:@L> "<" <m:"mut"?> @L <l:Id> ":" <s:Symbol0> ">" <hi:@R> =>
```
nikomatsakis: 👍

nikomatsakis: Don't we need to modify the tokenizer?

CreepySkeleton:
Hi guys, what's status of this PR? I would really like this feature available in lalrpop because `let mut v = v;` quite frankly sucks. If this is stalled just because the author stopped looking after this PR and nobody has spare time to deal with it I would gladly volunteer for the job and sync this with current code-base.

ahmedcharles:
Yeah, I just haven't had time and forgot about this. Feel free to push it over the finish line.

Marwes:
Closed by #473

Issue #269, opened 2017-10-24
Performance degredation with noddy csv parser


ehiggs:
The performance of [this csv parser code](https://bitbucket.org/ewanhiggs/csv-game/src/2cbb0d1103649cfa1240748a175c90dca72a28d0/rust/lalr-reader/?at=master) went from reading 1m lines in ~1second to taking minutes with 13.1.

The grammar is:

```
grammar;                                                                        
                                                                                
pub Record = Comma<Field>;                                                      
                                                                                
Comma<T>: Vec<T> = {                                                            
    <v:(<T> ",")*> <e:T?> => match e {                                          
        None=> v,                                                               
        Some(e) => {                                                            
            let mut v = v;                                                      
            v.push(e);                                                          
            v                                                                   
        }                                                                       
    }                                                                           
};                                                                              
                                                                                
Field: String = {                                                               
    <s:r#""[^"]+""#> => {                                                       
        s[1..s.len()-1].to_string()                                             
    },                                                                          
    <s:r#"[^",][^,]*"#> => {                                                    
        s.to_string()                                                           
    },                                                                          
}; 
```

The test program counts the number of fields encountered:

```rust
pub mod csv;

use std::io::BufReader;
use std::io::BufRead;
use std::fs::File;

fn main() {
    let fpath = ::std::env::args().nth(1).unwrap();
    let f = File::open(fpath).unwrap();
    let file = BufReader::new(&f);
    let mut sum = 0;
    for line in file.lines() {
        let l = line.unwrap();
        let rec = csv::parse_Record(&l).unwrap();
        sum += rec.len();
    }
    println!("{}", sum);
}
```

Results:

```
$ for t in 1 10 100 1000 10000 100000 ; do time ./target/release/lalr-reader <(head -${t} /tmp/hello.csv ); done
5

real    0m0.006s
user    0m0.002s
sys     0m0.009s
50

real    0m0.006s
user    0m0.004s
sys     0m0.006s
500

real    0m0.014s
user    0m0.014s
sys     0m0.001s
5000

real    0m0.085s
user    0m0.084s
sys     0m0.001s
50000

real    0m0.838s
user    0m0.837s
sys     0m0.002s
500000

real    0m8.274s
user    0m8.261s
sys     0m0.007s

$ time ./target/release/lalr-reader /tmp/hello.csv 
5000000

real    1m22.654s
user    1m22.554s
sys     0m0.007s
```

nikomatsakis:
[...]

PR #317, opened 2018-02-08
Change parse_X to XParser::parse.

ahmedcharles:
#269 Step 2.

### DIFF
--- a/doc/calculator/src/main.rs
+++ b/doc/calculator/src/main.rs
@@ -4,30 +4,30 @@ pub mod calculator1; // syntesized by LALRPOP
 
 #[test]
 fn calculator1() {
-    assert!(calculator1::parse_Term("22").is_ok());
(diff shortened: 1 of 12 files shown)


ahmedcharles:
Should wait for #316

nikomatsakis:
Can you rebase?

ahmedcharles:
I rebased. I'll update the docs in another commit, if that's ok? Though, I'm really curious why the CI is passing. Testing lalrpop is weird cause the build script doesn't detect that it needs to update the `.rs` based on the `.lalrpop`. Though, CI shouldn't have that problem.

nikomatsakis:
@ahmedcharles the CI doesn't test the text of the tutorial... not sure why it should fail?

solson:
@ahmedcharles @nikomatsakis The tutorial is still not updated and I got pretty confused trying LALRPOP for the first time!

PR #320, opened 2018-02-09
fix error on parsing a single-quote char literal

### LINKED ISSUE #307: Possible lalprop parser error
Hit an issue in which legal rust makes lalrpop choke during parsing the grammar definition;

Problematic grammar file used:
```
grammar;

pub StringArg: String = {
    <s:r#"'[^']*'"#> => s.chars().skip(1).take_while(|c| *c != '\'').collect(),
};
```

Causing this error:
```
processing file `test_quote.lalrpop`
... test_quote.lalrpop:7:2: 7:2 error: unexpected token: `;`
--- stderr
  };
   ^
```

But note that it's specifically the `'\''` causing the issue, because replacing the `'\''` with any other character (Even the effectively identical `'\u{0027}'`) works just fine.

dtkerr:
Fixes #307

When encountering the sequence `'\` only search for the closing `'` after the character following the `\`.

### DIFF
--- a/lalrpop/src/tok/mod.rs
+++ b/lalrpop/src/tok/mod.rs
@@ -548,6 +548,9 @@ impl<'input> Tokenizer<'input> {
         self.lookahead.and_then( |(_, c)| {
             if c == '\\' {
                 // escape after `'` => it had to be character literal token, consume
+                // the backslash and escaped character, then consume until `'`
+                self.bump();
+                self.bump();
                 self.take_until_and_consume_terminating_character(|c:char| { c == '\'' })
             } else {
                 // no escape, then we require to see next `'` or we assume it was lifetime
--- a/lalrpop/src/tok/test.rs
+++ b/lalrpop/src/tok/test.rs
@@ -331,6 +331,14 @@ fn equalsgreaterthancode_error_end_of_input_instead_of_closing_normal_character_
 }
 
 #[test]
+fn equalsgreaterthancode_single_quote_literal() {
+    test(r#"=> { println!('\''); },"#, vec![
+        (r#"~~~~~~~~~~~~~~~~~~~~~~ "#, EqualsGreaterThanCode(r#" { println!('\''); }"#)),
+        (r#"                      ~"#, Comma),
+    ]);
+}
+
+#[test]
 fn code_paren() { // Issue #25
     test(r#"=> a("(", c),"#, vec![
         (r#"~~~~~~~~~~~~ "#, EqualsGreaterThanCode(r#" a("(", c)"#)),


lalrpop/src/tok/mod.rs
```diff
@@ -548,7 +548,13 @@ impl<'input> Tokenizer<'input> {
         self.lookahead.and_then( |(_, c)| {
             if c == '\\' {
                 // escape after `'` => it had to be character literal token, consume
-                self.take_until_and_consume_terminating_character(|c:char| { c == '\'' })
```
ahmedcharles: Wouldn't it be better to call `bump` and then call `take_until_and_consume_terminating_character` as before?

dtkerr:
It sort of already has a test - although obviously the test is errant in some way as it was passing even before this fix

```rust
#[test]
fn char_literals() {
    test(r#"'foo' 'a 'b '!' '!!' '\'' 'c"#, vec![
        ...
        (r#"                     ~~~~   "#, CharLiteral("\\'")),
        ...
    ]);
}
```

ahmedcharles:
Can you find one that fails before the change and passes after?

lalrpop/src/tok/mod.rs
dtkerr: Ah, good point.

dtkerr:
I'll see if I can't get a better test to go with this as well.

dtkerr:
OK: got a test that fails on the old code, passes with the new fix.

I'm not sure if it's the best idea for this PR to cover actually adding that test though: the core issue that the `code` function was essentially untested prior to this, so it might be a better idea to add a new issue for a more exhaustive testset to be added to cover the `code` function.

The reason the existing test previously referenced passed both before and after the fix: the function `lifetimeish` was responsible for parsing tokens in the char_literals` test, not the `take_until_and_consume_terminating_character` - the `take_until_and_consume_terminating_character` function is only called by `code`, and as stated before: `code` is lacking any test coverage.

ahmedcharles:
I think it's fine to add a single test to a previously untested function. After all, you have to start somewhere with increasing test coverage.

nikomatsakis:
Every bugfix should have a test, that's my rule. =)

lalrpop/src/tok/test.rs
```diff
@@ -509,3 +509,10 @@ fn char_literals() {
         (r#"                          ~~"#, Lifetime("'c")),
     ]);
 }
+
+#[test]
+fn code_blocks() {
+    let input = r#" {do_something(); println!('{'); println!('\''); }, { other_block(); } "#;
+    let mut tokenizer = Tokenizer::new(&input, 0);
+    assert_eq!(tokenizer.code(0, "([{", "}])"), Ok(50));
```
nikomatsakis: 👍 to a test, but can't we use the `test!` infrastructure for this? something like

```rust
#[test]
fn code_blocks() {
    test(r#" {println!('\'');}, {other_block();} "#, vec![
        (r#" ~~~~~~~~~~~~~~~~~ ", ...)]);
}
```

That would avoid hardcoding indices and things, and (of course) make things more uniform.

dtkerr:
There's a test just to cover the specific issue fixed by the changes; should prevent regression.

lalrpop/src/tok/test.rs
dtkerr: The existing `test` just seemed very token-oriented to me, whereas code is only ever really expected to return an integer once.

That said I agree it'd be nice to do something similar (but separate) with the same tilde underlining to specify expected matches. If we just merge this fix as-is I'll open and work on an issue to expand the `code` tests to something more exhaustive and we can worry about that there.

Sound good?

nikomatsakis: Hmm, I'm not sure I understand. How is this not a token-oriented test? `code` is just an internal helper function, but it's "effects" bubble up and ought to be visible in terms of the tokenization, right? Is it that you prefer to test the `code` function individually, rather than the output of the tokenizer? Put another way, I like the way the tests are written now, and this doesn't (yet) seem like an improvement to me, so I'm wondering what I'm missing. =)

dtkerr: Ah, I think I see what you want instead - check out the test as it is now in the latest commit.

nikomatsakis:
@dtkerr Thanks!

PR #324, opened 2018-02-19
feat: De-duplicate symbol variants for parse tables

Marwes:
This attempts to merge all symbol variants which would contain the same
type into one, letting us reduce the number of `pop_` functions being
generated.

Reduces the number of lines for `lrgrammar.lalrpop` from 67k to 57k
(a reduction of about 15%)

### DIFF
--- a/lalrpop/src/collections/map.rs
+++ b/lalrpop/src/collections/map.rs
@@ -1,5 +1,7 @@
 use std::collections::BTreeMap;
 
+pub use std::collections::btree_map::Entry;
+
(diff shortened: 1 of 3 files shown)


Marwes:
With "attempts"  I mean that we can only compare types syntactically, if a type is behind an alias then it will look different and get its own variant.

udoprog:
Oooh, very nice: :+1:

nikomatsakis:
@Marwes clever

PR #327, opened 2018-03-01
Run rustfmt on entire codebase

brendanzab:
I was trying to work on LALRPOP, but then realised that rustfmt hadn't been run on it. I initially was going through file-by-file to make sure the code was formatted reasonably, but quickly realised that given LALRPOP's size this would be impractical, so I ended up using `cargo fmt` on the whole lot.


brendanzab:
I normally put `match_block_trailing_comma = true` in a `rustfmt.toml` (I think this syntax is more consistent), but you aren't following this style, so I left it out.

nikomatsakis:
> I normally put match_block_trailing_comma = true in a rustfmt.toml (I think this syntax is more consistent), but you aren't following this style, so I left it out.

I would prefer the default style, whatever it is

PR #328, opened 2018-03-05
Update rand dependency.

waywardmonkeys:


waywardmonkeys:
Fixed the commit comment to give the right dependency name. Sorry about that.

PR #338, opened 2018-03-17
fix: Allow lalrpop parsers to be used with include!

Marwes:
Due to https://github.com/rust-lang/rust/issues/18810 (https://github.com/rust-lang/rfcs/issues/752)
it is not possible to have an inner attribute in a file that gets
used with `include!`.

Specifying the attribute on each parser module is less precise and more
duplication but it will at least work for that case. If a user wants to
apply the rustfmt on the whole module they can still add an outer
attribute manually as well.


nikomatsakis:
Annoying but necessary. =) Nice.

Issue #325, opened 2018-02-20
Allow `_` in `match` to appear earlier on


brson:
I have a grammar  for which the match statement looks like

```
match {
    "access_log",
    "client_body_temp_path",
    "default_server",
    "error_log",
    "events",
    "fastcgi_temp_path",
    "http",
    "index",
    "listen",
    "pid",
    "proxy_temp_path",
    "root",
    "scgi_temp_path",
    "server",
    "server_name",
    "user",
    "uwsgi_temp_path",
    "worker_connections",
    "worker_processes",
    "worker_rlimit_nofile",
    ";",
    "{",
    "}",
} else {
    r"[[:digit:]]+" => NUMBER,
} else {
    r"[[:alnum:]_]+" => NAME,
} else {
    r"[[:alnum:]/.]+" => PATH,
}
```

I sure wish I didn't have to write all those simple literals out, but I can't figure out how to avoid them. I can't put the wildcard `_` in the first arm of `match`

```
match {
    _
} else {
    r"[[:digit:]]+" => NUMBER,
} else {
    r"[[:alnum:]_]+" => NAME,
} else {
    r"[[:alnum:]/.]+" => PATH,
}
```

That doesn't work:

```
brian@DESKTOP-UCV672I:~/engine-arr⟫ cargo run -- -c examples/blocks.conf -T
   Compiling arr v0.1.0 (file:///home/brian/engine-arr)
error: failed to run custom build command for `arr v0.1.0 (file:///home/brian/engine-arr)`
process didn't exit successfully: `/home/brian/engine-arr/target/debug/build/arr-f72285b38b3e56b8/build-script-build` (exit code: 1)
--- stdout
processing file `/home/brian/engine-arr/src/config_parser.lalrpop`
/home/brian/engine-arr/src/config_parser.lalrpop:11:5: 11:5 error: Catch all must be final item

--- stderr
      _
      ^

```

The "catch all" must be the final item. So this can be written:

```
match {
    r"[[:digit:]]+" => NUMBER,
} else {
    r"[[:alnum:]_]+" => NAME,
} else {
    r"[[:alnum:]/.]+" => PATH,
} else {
    _
}
```

It doesn't do what I want it to do though. My simple tokens in the grammar seemingly get lexed as regexes, resulting in some confusing error messages:

```
   Compiling arr v0.1.0 (file:///home/brian/engine-arr)
    Finished dev [unoptimized + debuginfo] target(s) in 8.75 secs
[...]

PR #341, opened 2018-03-19
Solution for issue #325 (Allow `_` in `match` to appear earlier on)

k0pernicus:
This is my solution for the issue #325.

I didn't know how to deal with _lalrpop_snap_ tests, as the _snap_ is corresponding to the _master_ project - so I updated the code for both.

All the code modifications are in _lalrpop/src/normalize/token_check/mod.rs_, related to @nikomatsakis recommendations in #325.

The file _lalrpop-test/src/allow_catch_all.lalrpop_ is corresponding to a new test, to know if the solution works or not.


k0pernicus:
Last change: I included `lalrpop-test/src/allow_catch_all.rs` as a file to ignore in the future.

lalrpop-snap/src/normalize/prevalidate/mod.rs
```diff
@@ -75,20 +75,6 @@ impl<'grammar> Validator<'grammar> {
                             "extern and match definitions are mutually exclusive"
                         );
                     }
-
```
Marwes: You shouldn't need to change anything in  `lalrpop-snap`, it is only a "snapshot" of `lalrpop` used to compile used to compile `lalrpop` itself and is only updated occasionally by running `.snap.sh`

k0pernicus: Thanks for the comment.
I only changed tests in `lalrpop-snap` because they were running using `cargo test`, and failed.
I could have run the tests only on lalrpop, indeed, using `cargo test lalrpop`... Sorry for that.

Marwes: Hmm, as long as nothing in snap is changed I'd expect the tests to still pass regardless :/ Are you sure you didn't accidentally do this change in snap and then ran the tests? (I know I accidentally open a file in snap sometimes at least.)

k0pernicus: Indeed, I made a change accidentally.
I just removed the changes in `snap` ;-)
Thanks!

lalrpop-test/src/allow_catch_all.lalrpop
```diff
@@ -0,0 +1,16 @@
...
+}
+
+pub Term = {
+    Num,
+    "(" <Term> ")",
+    "22" => format!("Twenty-two!"),
+    r"\w+" => format!("Id({})", <>), // <-- we added this
+};
+
+Num: String = r"[0-9]+" => <>.to_string();
```
Marwes: Nit: Missing newline at EOF.

lalrpop-test/src/allow_catch_all.lalrpop
```diff
@@ -0,0 +1,16 @@
+grammar;
```
Marwes: It might be good to test that the precedence is correct by running the parser on some different inputs in https://github.com/lalrpop/lalrpop/blob/master/lalrpop-test/src/lib.rs.

lalrpop-test/src/allow_catch_all.lalrpop
```diff
@@ -0,0 +1,16 @@
...
+
+match {
+    _
+} else {
+    r"\w+"
+}
+
```
nikomatsakis: As @Marwes noted, if we're going to add a test here, we should add a `mod allow_catch_all;` in `lib.rs` and some code that uses it. That said, I think it might be better to write one of the more targetd unit tests, like those found in [this module](https://github.com/lalrpop/lalrpop/blob/df6fd076565530a5dcce269f8a171b5d4d934d36/lalrpop/src/normalize/token_check/test.rs#L159-L164)

lalrpop-test/src/allow_catch_all.rs
```diff
@@ -0,0 +1,1213 @@
+// auto-generated: "lalrpop 0.14.0"
+#![cfg_attr(rustfmt, rustfmt_skip)]
+#[allow(unused_extern_crates)]
+extern crate lalrpop_util as __lalrpop_util;
+
+mod __parse__Term {
+    #![allow(non_snake_case, non_camel_case_types, unused_mut, unused_variables, unused_imports, unused_parens)]
```
nikomatsakis: we usually add these generated files to `.gitignore` instead

nikomatsakis:
Ping @k0pernicus -- think you'll get a chance to tweak the tests soon? =)

lalrpop-test/src/allow_catch_all.lalrpop
k0pernicus: Ok, I see your point.
I will remove the integration test and add a targeted unit test in `token_check/test.rs` ;-)

lalrpop/src/normalize/token_check/test.rs
```diff
@@ -163,6 +163,27 @@ fn match_catch_all() {
...
+fn match_catch_all_in_first_arm() {
+    let grammar = r#"
+        grammar;
+        match {
+            _
+        } else {
+            r"\w+"
+        }
+        pub Term = {
+            Num,
```
nikomatsakis: let's add a case that overlaps with the `r"\w+"` that appears in the match. Maybe `r"[a-z]"`? This should not be an error, because it is given higher precedence.

lalrpop/src/normalize/token_check/test.rs
```diff
@@ -163,6 +163,27 @@ fn match_catch_all() {
...
+            "(" <Term> ")",
+            r"\w+" => format!("Id({})", <>),
+        };
+        Num: String = r"[0-9]+" => <>.to_string();
+"#;
+    assert!(validate_grammar(&grammar).is_ok())
```
nikomatsakis: Great! 

I wonder, can we use `check_intern_token` to also test how things are classified? In particular, the precedence. If we add `r"[a-z]"` as above, then we would expect that a single letter (like `x`) would match `r"[a-z]"` (which has higher precedence), but multiple letters would match `r"\w+"`:

```rust
check_intern_token(
    grammar,
    vec![
        ("x", r##"Some(r#"[a-z]"#, "x")##),
        ("xy", r##"Some(r#"\w+"#, "xy")##),
    ],
);
```

nikomatsakis:
I'm adding this to the 1.0 milestone -- I know it's been many years, but this still seems like a useful change. Will take a look once we get a maintenance release out the door.

Pat-Lafon:
I believe this has landed in #807 so I will be closing this pr. Thanks for starting this!

PR #346, opened 2018-03-23
update tutorials

nikomatsakis:
forgot a few version numbers :)

### DIFF
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -18,12 +18,12 @@ build = "build.rs" # LALRPOP preprocessing
 # (If you write your own tokenizer, or already have the regex
 # crate, you can skip this dependency.)
 [dependencies]
-lalrpop-util = "0.14.0"
+lalrpop-util = "0.15.0"
 regex = "0.2.0"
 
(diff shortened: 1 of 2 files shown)


Marwes:
@nikomatsakis I copied lalrpop's version script in gluon and extended it also replace a few specific files which I know have version numbers in them (less risk of accidental changes).

https://github.com/gluon-lang/gluon/blob/5111379ca2ac88c526a1b5e1fa9199bdc2627c99/scripts/version.sh#L30-L31

Not exactly sure of perl's regex syntax but something like this I guess?

```
perl -p -i -e 's/^lalrpop([\-a-z]*) *= *"[0-9.]+"/lalrpop\1 = "'$1'"/' \
    doc/src/quick_start_guide.md doc/src/tutorial/001_adding_lalrpop.md
```

nikomatsakis:
@Marwes nice. I've been wanting to make a better version of that script. I was just thinking that it'd be nice to make a general `publish` utility for things like LALRPOP -- I hit the same problems in basically every crate.

Marwes:
Opened #351 to automatically update these versions as well.

PR #353, opened 2018-03-27
Generate files in OUT_DIR

### LINKED ISSUE #280: Consider generating .rs files into OUT_DIR by default

ayosec:
This patch changes the default output directory for generated files, as described in https://github.com/lalrpop/lalrpop/issues/280#issuecomment-360732898.

The `lalrpop-snap` is untouched.

The binary (`cargo install lalrpop`) keeps the previous behaviour.

The documentation in `out_dir.md` is replaced with details on how to restore the previous behaviour.

### Macro `lalrpop_mod!`

There is a new macro `lalrpop_mod!` to define module for the common usage.

The `lalrpop_util` crate has to be loaded with the `macro_use` attribute.

```rust
#[macro_use] extern crate lalrpop_util; 
```

#### Examples

To define a module for `src/parser.lalrpop`:

```rust
lalrpop_mod!(parser);

// is expanded to 

mod parser {
    include!(concat!(env!("OUT_DIR"), "/parser.rs"));
}
```

```rust
lalrpop_mod!(pub parser);

// is expanded to 

pub mod parser {
    include!(concat!(env!("OUT_DIR"), "/parser.rs"));
}
```

To define a module for `src/parser/lang.lalrpop`:

```rust
lalrpop_mod!(lang, "/parser/lang.rs");

// is expanded to

mod lang {
    include!(concat!(env!("OUT_DIR"), "/parser/lang.rs")); 
}
```

```rust
lalrpop_mod!(pub lang, "/parser/lang.rs");

// is expanded to

pub mod lang {
    include!(concat!(env!("OUT_DIR"), "/parser/lang.rs")); 
}
```

Some (rare) usages will have to be written manually.
 
---

Closes #280


lalrpop-test/src/lib.rs
```diff
@@ -11,104 +11,104 @@ use util::tok::Tok;
 
 /// Tests that actions can return the grammar's type parameters' associated
 /// types.
-mod associated_types;
+lalrpop_mod!(associated_types);
```
Marwes: Something to consider with the parser modules of `lalrpop-test` is that one may want to inspect the code of these. So keeping the genereated files in the same directory might be nice. On the other hand it is nice to remove all these entries from the .gitignore as well. Not sure which way is the best here.

lalrpop-test/src/expr_module_attributes.lalrpop
```diff
@@ -1,4 +1,4 @@
-#![cfg_attr(feature = "cargo-clippy", allow(clippy))]
+//#![cfg_attr(feature = "cargo-clippy", allow(clippy))]
```
Marwes: Remove this perhaps? Or move it to an outer attribute instead as the rustc bug/limitation may take a while to be fixed.

lalrpop/src/api/mod.rs
```diff
@@ -149,7 +149,22 @@ impl Configuration {
...
+        // If in/out dir are empty, use cargo conventions by default.
+        // See https://github.com/lalrpop/lalrpop/issues/280
+        if session.in_dir.is_none() {
+            let mut in_dir = try!(env::current_dir());
+            in_dir.push("src");
+            session.in_dir = Some(in_dir);
+        }
+
+        if session.out_dir.is_none() {
+            let out_dir = env::var_os("OUT_DIR").expect("missing OUT_DIR variable");
```
Marwes: This could return an error instead of panicking (since Box<Error> is used &'static str can be returned)

lalrpop-test/src/lib.rs
ayosec: > Something to consider with the parser modules of lalrpop-test is that one may want to inspect the code of these. So keeping the genereated files in the same directory might be nice.

I think that a dedicated repository (like https://github.com/lalrpop/samples) is better for that purpose. Thus, (a) people don't need to download and compile the repository in order to see the generated code, and (b) it can be read directly in GitHub.

I created a [lalrpop-samples](https://github.com/ayosec/lalrpop-samples) repository as an example.

> On the other hand it is nice to remove all these entries from the .gitignore as well. Not sure which way is the best here.

I didn't see that. I updated the `.gitignore` file in 2c9b14487cda6e203a02366f7020d91006e764bf.

lalrpop-test/src/expr_module_attributes.lalrpop
ayosec: Removed in 576a30515e7365aa8679161b7070609f65f9109a.

lalrpop/src/api/mod.rs
ayosec: Fixed in 12933815118e93130bed4f3dfa2ecad0d0d8f0ac.

lalrpop-test/src/lib.rs
Marwes: > I think that a dedicated repository (like https://github.com/lalrpop/samples) is better for that purpose. Thus, (a) people don't need to download and compile the repository in order to see the generated code, and (b) it can be read directly in GitHub.

I were thinking of this in terms of developing on lalrpop and changing the codegen. In that case it is easier to look at the generated code if it is directly in the repo structure instead of inside `target/debug/deps/lalrpop-712387hgsagd/...`. I will let @nikomatsakis have the final say on this though (I believe he is at mozilla all hands this week so he probably don't have time this week).

> I didn't see that. I updated the .gitignore file in 2c9b144.

:+1:

nikomatsakis: On balance, not having to add to the gitignore file seems good. I was wondering though: when you are using the OUT_DIR, what happens if you get a compilation error? (I'll have to experiment myself). How good are we at giving you source error information?

nikomatsakis:
Seems like error messages work ok.

nikomatsakis:
However, I'm getting a bunch of stuff like this:

```
   Compiling lalrpop-test v0.15.0 (file:///Users/nmatsakis/versioned/lalrpop/lalrpop-test)
warning: unused import: `Tok`
 --> /Users/nmatsakis/versioned/lalrpop/target/debug/build/lalrpop-test-e5feb3db859e63b5/out/no_clone_tok.rs:2:29
  |
2 | use util::tok::{NoCloneTok, Tok};
  |                             ^^^
  |
  = note: #[warn(unused_imports)] on by default

warning: unused import: `lalrpop_util::ParseError`
 --> /Users/nmatsakis/versioned/lalrpop/target/debug/build/lalrpop-test-e5feb3db859e63b5/out/no_clone_tok.rs:3:5
  |
3 | use lalrpop_util::ParseError;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^

warning: unreachable pattern
   --> /Users/nmatsakis/versioned/lalrpop/target/debug/build/lalrpop-test-e5feb3db859e63b5/out/no_clone_tok.rs:464:21
    |
464 |                     _ => panic!("symbol type mismatch")
    |                     ^
    |
    = note: #[warn(unreachable_patterns)] on by default

warning: unreachable pattern
   --> /Users/nmatsakis/versioned/lalrpop/target/debug/build/lalrpop-test-e5feb3db859e63b5/out/no_clone_tok.rs:464:21
    |
464 |                     _ => panic!("symbol type mismatch")
    |                     ^
    |
    = note: #[warn(unreachable_patterns)] on by default

warning: method is never used: `new`
  --> /Users/nmatsakis/versioned/lalrpop/target/debug/build/lalrpop-test-e5feb3db859e63b5/out/expr_module_attributes.rs:19:9
   |
19 |         pub fn new() -> ExprParser {
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default
```

Any thoughts on what's up with that? @Marwes -- didn't you have some change relevant to this?

Marwes:
That is the first time I have seen those warnings, though I think the merging of symbol variants may have caused the unreachable pattern warning https://github.com/lalrpop/lalrpop/blob/master/lalrpop-test/src/no_clone_tok.lalrpop.

ayosec:
The patch is updated (ad0c91d4962cdea240ef2a46cd7cea7aeb509a1b) to fix the conflicts against master.

About the warnings, should I update the PR to fix them? I guess that the changes are:

* Include a `#[allow(unreachable_patterns)]` in some generated `match` blocks.
* Remove the unused imports.

Marwes:
@ayosec 

For the unreachable patterns you can add `if true =>` at https://github.com/lalrpop/lalrpop/blob/2694da775dd8535322f5add914d42bd4f377c01e/lalrpop/src/lr1/codegen/parse_table.rs#L1184 . That way rust won't figure out that the pattern after is unreachable even if there is oly one variant.

Otherwise it sounds correct 👍 .

ayosec:
> Needs to be rebased to remove the merge commit, I think.

I'm sorry (again) for the delay.

I rebased the branch with the current master.

matklad:
FWIW, nightly Cargo will issue warnings on `cargo publish`,  if build script tries to modify `/src`: https://github.com/rust-lang/cargo/pull/5584

Marwes:
(Still) looks good to me! Thanks for the PR and sorry for the delay. I will try and get a (breaking) release out some time this week.

PR #355, opened 2018-03-30
Make semicolon after `}` in rules optional

matklad:
Hi! I think we shouldn't require `;` after `}`, because this is very different from what Rust is doing. Let's make it optional for starters. I think we can add a warning and eventually remove this later as well?


matklad:
Hm, looks like the original reasoning for semicolons was that rules actually resemble let-bindings? At least for me, they feel closer to items, and that's why I find semicolon after `}` confusing.

lalrpop-test/src/optional_semicolon.rs
```diff
@@ -0,0 +1,710 @@
+// auto-generated: "lalrpop 0.15.0"
```
Marwes: This file should be .gitignored (or with #353 it may be put in OUT_DIR instead)

matklad: I think I'll just write a unit-tests instead. I think I found some in the `grammar` directory.

matklad:
Updated with a more reasonable test! :)

nikomatsakis:
Hmm. To me, having `=` and then *not* having `;` is very alien. But it's fine, we can make it optional I guess.

PR #359, opened 2018-03-31
Correctly anchor regex at the begining

### LINKED ISSUE #358: `match` doesn't handle simple alternation in regex?
Hi! I have a following grammar:

```
grammar;

pub File = bare_key bool;

match {
    r"false|true" => bool,
} else {
    r"[a-z]+" => bare_key,
}
```

it should recognize stuff like `foo true` or `bar false`. While it correctly recognizes the `false`, it is for some reason stomped by `true`...

```Rust
#[test]
fn you_shall_not_parse() {
    let p = grammar::FileParser::new();
    let r = p.parse("foo false");
    assert!(r.is_ok()); // this passes

    let r = p.parse("foo true");
    assert!(r.is_ok(), "{:?}", r);
   // this fails with
   // (UnrecognizedToken { token: Some((0, Token(1, "foo true"), 8)), expected: ["bare_key"] })
}
```

I believe this is a bug in the tokenizer? Or am I missing something obvious?

matklad:
^foo|bar is (^foo)|bar, and not ^(foo|bar). Adding parens should
fix it.

closes #358 

Funny how this is the second time I fix this bug in a parser generator, the fist fix was for fall: https://github.com/matklad/fall/commit/02b7fe1f471aeeb836b414caa3fe8a9235dc131c =P

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ lalrpop-test/src/unit.rs
 lalrpop-test/src/use_super.rs
 lalrpop-test/src/no_clone_tok.rs
 lalrpop-test/src/partial_parse.rs
+lalrpop-test/src/match_alternatives.rs
 lalrpop-test/src/match_section.rs
 lalrpop-test/src/expr_module_attributes.rs
 lalrpop-test/src/associated_types.rs
--- a/lalrpop-test/src/lib.rs
+++ b/lalrpop-test/src/lib.rs
@@ -99,6 +99,7 @@ mod unit;
 
 /// test for match section
 mod match_section;
+mod match_alternatives;
 
 /// regression test for issue #253.
 mod partial_parse;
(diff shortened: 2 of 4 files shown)


matklad:
I'd love to add a test for this, where are the tests for generated lexer?

matklad:
> the fist fix was for fall: 

Lol, moreover it was almost *exactly* a year ago :candle: :cake:

Marwes:
> I'd love to add a test for this, where are the tests for generated lexer?

I'd guess it is either dfa/test.rs or nfa/test.rs, not sure which though.

matklad:
@Marwes does not looks so: the code for `^` does not seem to be executed for those tests at all. Looks like nfa/dfa stuff interpret something, while `^` happens at the codegen.

PR #362, opened 2018-04-09
Clarify 0.15 upgrade instructions

dtolnay:
The X in `parse_X` is not the same X as in `X::new()` -- for example in my code I needed to go from `parse_Expr` to `ExprParser::new()`.


Dylan-DPC:
Thanks

Issue #260, opened 2017-09-07
lalrpop v0.13.1 is very slow to compile


vi:
`Compiling lalrpop v0.13.1`: 6 CPU minutes and 900 megabytes of used memory is a bit too big.

Why compilation is that slow?

JustANull:
With stable Rust (`rustc 1.21.0 (3b72af97e 2017-10-09)`) I'm getting stack overflows on OSX trying to compile lalrpop as a dependency of a practically empty project.

jason-wolfe:
This is by far the slowest part of a release build in a large project I have as well.

ehiggs:
Dumping compiler timings to see where the time goes...
```
$ cargo rustc --lib -- -Ztime-passes -Zperf-stats -Ztrans-time-graph
   Compiling lalrpop v0.13.1 (file:///Users/ehiggs/src/download/lalrpop/lalrpop)
  time: 0.689	parsing
  time: 0.000	recursion limit
  time: 0.000	crate injection
  time: 0.000	plugin loading
  time: 0.000	plugin registration
  time: 0.000	background load prev dep-graph
  time: 0.509	expansion
  time: 0.000	maybe building test harness
  time: 0.012	maybe creating a macro crate
  time: 0.060	creating allocators
  time: 0.000	checking for inline asm in case the target doesn't support it
  time: 0.013	AST validation
  time: 0.286	name resolution
  time: 0.060	complete gated feature checking
  time: 0.226	lowering ast -> hir
  time: 0.109	early lint checks
  time: 0.304	indexing hir
  time: 0.012	attribute checking
  time: 0.000	load query result cache
  time: 0.000	looking for entry point
  time: 0.000	looking for plugin registrar
  time: 0.012	loop checking
  time: 0.004	static item recursion checking
  time: 0.052	stability checking
  time: 0.597	type collecting
  time: 0.001	outlives testing
  time: 0.005	impl wf inference
  time: 0.129	coherence checking
  time: 0.001	variance testing
  time: 0.518	wf checking
  time: 0.956	item-types checking
  time: 337.795	item-bodies checking
  time: 2.874	const checking
  time: 0.322	privacy checking
  time: 0.021	intrinsic checking
  time: 0.170	match checking
  time: 3.567	liveness checking
  time: 19.340	borrow checking
  time: 0.007	MIR borrow checking
  time: 0.002	MIR effect checking
  time: 0.049	death checking
  time: 0.000	unused lib feature checking
[...]

PR #366, opened 2018-04-20
fix: Improve compile times by reducing the size of the reduce actions 

Marwes:
This vastly improves the compile times of generated parse table parsers. Apparently both rustc and llvm struggle with the huge `match` expression that gets created.

Fixes(?) #260 

## time cargo build -p lalrpop

(This includes compiling all of LALRPOPs dependencies as well).

### Before

```
real    5m5.064s
user    0m0.000s
sys     0m0.015s
```

### After

```
real    2m20.169s
user    0m0.000s
sys     0m0.015s
```


Marwes:
Building on top of this it would be possible to to reuse the symbol type and these generated `reduceN` functions between all of the entrypoints which could bring down the size of the generated code quite considerably (cc #304).

Marwes:
`string_cache` broke compatibility with rust < 1.23 so I locked the version to 0.7.1 which works. Could perhaps try and get a new version of string_cache but not sure it is worth the effort?

lalrpop-snap/Cargo.toml
```diff
@@ -24,7 +24,7 @@ itertools = "0.7"
 regex = "0.2.1"
 regex-syntax = "0.4.0"
 petgraph = "0.4.4"
-string_cache = "0.7.1"
+string_cache = "0.7.1,<=0.7.1"
```
brendanzab: Couldn't this be `string_cache = "=0.7.1"`?

matklad: Locking dep to a particular version can be problematic. For example, this would make LALRPOP unusable with anything which specifies `string_cache = "0.7.2"`. I think a better solution would be to do `cargo update string_cache "0.7.1"` in CI for Rust 1.20.0, if we don't want to just bump the minimal Rust version.

Dylan-DPC: I think =0.7.1 should be fine. I agree with matklad that dependencies shouldn't be locked to a specific version, but in this case there is no direct solution unless we specify a minimum compiler version.

matklad: >Since it is an internal dependency it is ok to have it locked down

Unfortunately Cargo is not able to distinguish between internal and public dependencies yet, so it just flatly prohibits resolution graphs with two semver-compatible versions. That is, if crate `foo` has

```
[dependencies]
string_cache = "0.7.2"
lalrpop = "0.13.2" # pretend this is the new published lalrpop version with `=` or `<=` constraint
``` 

then Cargo won't be able to resolve its dependencies at all. 


In general, I would argue that semver open `0.7.1` is the correct constraint here, regardless of the above issue. 

It is the leaf crate with Cargo.lock that ultimately decides about dependencies. If the leaf crate must be built with Rust 1.20.0, it should have `string_cache = 0.7.1` in lockfile. If it is build with say, `1.25.0`, it should have `string_cache = 0.7.2` in the lock. The job of LALRPOP here is to be compatible with both, and not to impose additional versioning constraints. 

Does this make sense? :)

matklad: In general, we plan to lint against using non semver-open dependencies for published crates, because there's rarely a need for that: https://github.com/rust-lang/cargo/issues/5340. 

In particular, I believe that comparability with older `rustc` is also not a reason to constraint upper versions of dependencies in Cargo.toml. Instead the right predicate here is perhaps that *minimal* versions specified in Cargo.toml are compatible with desired `rustc` version, and picking particular versions should be left to the leaf crate with a lockfile, because only the leaf knows which version of rustc is actually used.

Marwes: So cargo can only use multiple versions of the same dependency if they differ in major version? (or minor for 0.x crates)?

> In general, I would argue that semver open 0.7.1 is the correct constraint here, regardless of the above issue.

That is the better compromise, though it can lead to confusing problems for end users as well if they use <1.23. Not as bad as cargo refusing to compile though.

matklad: >So cargo can only use multiple versions of the same dependency if they differ in major version? (or minor for 0.x crates)?

That is correct! 

>though it can lead to confusing problems for end users as well if they use <1.23.

Yeah. Ideally, Cargo should be able to take current version of Rust into account during resolve, but that is non-trivial. 

Meanwhile I think we have a nightly feature (which hopefully will be stabilized soon) which should make testing rustc compat easier: https://github.com/lalrpop/lalrpop/pull/366#discussion_r183211817.

Dylan-DPC: How about getting servo to yank 0.7.2 and issue a new 0.8.0 release that will make matters easier? as this can be considered a breaking change even though BCs are allowed in 0.x land

matklad: There’s no consensus as to whether bumping required rustc version is a breaking change or not (discussion: https://github.com/rust-lang-nursery/api-guidelines/issues/123). I personally am convinced that it is not a breaking change.

Dylan-DPC: Agree.  it is one of those "debatable" topics. but won't you consider it a breaking change considering the new release won't work for anyone running  < 1.23?

matklad: Huh, turns out I feel strongly about it, so I've dropped a comment on that thread: https://github.com/rust-lang-nursery/api-guidelines/issues/123#issuecomment-383329604 :0)

Dylan-DPC: Uhmm makes sense. So i think the best thing to do here is to bump mvsr for lalrpop?

Marwes: @Dylan-DPC For the moment I think it is enough to just include a lockfile.

Marwes:
Merging this tonight as I don't think the issue of minimal versions needs to block this.

Dylan-DPC:
Moved the `string_cache` discussion to #368

Issue #14, opened 2015-09-14
Custom priorities, comment support when using an internal tokenizer


nikomatsakis:
The current tokenizer generation always uses two fixes precedence categories, so all regular expressions have equal weight. This is useful for giving a keyword like `"class"` precedence of an identifier regex, but there are times when we would like to give some regexs higher precedence over others. For example, if parsing a case-insensitive language like Pascal, you would like to use regexs like `r"[vV][aA][rR]"` rather than having to enumerate all combinations. But this won't work because of precedence conflicts. Another problem is that the tokenizer implicitly skips whitespace and there is no way to extend this set to skip other things, like comments.

### Prioritization

I've been contemplating a `match` declaration to address prioritization, which might look like:

```rust
match {
    r"[vV][aA][rR]",
    r"[a-zA-Z][a-zA-Z0-9]+",
}
```

The idea here is that when you have tokens listed in a match declaraction, we can create custom precedence levels, so that e.g. here the "var" regex takes precedence over the identifier regex. Tokens not listed would be implicitly added to the end of the list, with literals first and regex second.

### Comments and whitespace

I'm less clear on what to do here. I contemplated adding things to the match declaration with an "empty action", which would signal "do nothing":

```rust
match {
    r"{.*}" => { } // pascal style comment
}
```

or having something like an empty if declaration:

```rust
if r"{.*}";
```

I think I prefer the first alternative, but it doesn't seem great. Another thing that is unclear is if the implicit whitespace skipping should be disabled or retained. I think I prefer to retain it unless the user *asks* for it to be disabled, because it's always a bit surprising when you add something and find it implicitly removes another default. That is, adding comments to the list of things to skip implicitly removes whitespace. But not having the implicit whitespace at all feels like clearly the wrong default.

So some uncertainties here!

nikomatsakis:
I've been thinking about this more and I am feeling pretty good about `match` declarations as the means of writing tokenizer rules. To start with I'd probably just permit:

```
match {
    <pattern> => { }
}
```

Which would indicate that matches of `<pattern>` are to be ignored completely. The priority would be given the same priorities as today (0 for regular expressions, 1 for fixed strings). 

In the future, I would expand this to permit:

1. Explicit `#[priority=N]` declarations.
2. Fallible actions (`=>?`).
3. Actions that produce a token value or multiple token values. See below.
[...]

PR #373, opened 2018-05-05
Add a custom lexer to the lalrpop-util crate

Michael-F-Bryan:
Following from my comment on #14 I'd like to add a custom table-based lexer to the `lalrpop_util` crate, hidden behind a cargo feature flag.

My use case is that 90% of the time you'll want the ability to add some form of comments to the code being parsed, but getting everyone to create their own full-blown custom lexer is massively overkill. This PR adds a custom lexer which should be flexible enough to fit most people's uses.

I've copied this lexer around to two or three projects so far, so it should be fairly well documented. I *think* the doctests are enough to cover everything but let me know if it needs more tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,6 +221,9 @@ dependencies = [
 [[package]]
 name = "lalrpop-util"
 version = "0.15.2"
+dependencies = [
+ "regex 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "lazy_static"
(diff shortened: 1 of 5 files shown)


nikomatsakis:
Hi! Sorry for being slow. I've added this to my queue of things to look at.

xasopheno:
Love lalrpop.  I'm writing a language for composing music. This would be a helpful feature for me.

adjivas:
I don't know if that can help, but it's possible to use lalrpop parser with a lexer like flex, I have this example of code [lexer](https://github.com/ulysseB/telamon/blob/master/telamon-gen/src/lexer/mod.rs) and [grammar](https://github.com/ulysseB/telamon/blob/master/telamon-gen/src/exh.l).

xasopheno:
For anyone that ended up here trying to add simple comments to their lalrpop generator, my simple answer was to use a preprocessor to replace commented lines with `\n`:
```
    let f = File::open("filename.file").unwrap();
    let file = BufReader::new(&f);
    let mut result = String::new();
    for (num, line) in file.lines().enumerate() {
        let l = line.unwrap();
        let copy_l = l.trim_left();
        if copy_l.starts_with("--") {
            result.push_str("\n".to_string);
        } else {
            result.push_str(l);
        }
    };```

gyscos:
> For anyone that ended up here trying to add simple comments to their lalrpop generator, my simple answer was to use a preprocessor to replace commented lines with `\n`:

The problem is that it doesn't work if you're given a `&'input str` and must return a `ParsedItem<'input>`, and it messes up span indices in error messages.

PR #374, opened 2018-05-07
Update atty to 0.2, bit-set to 0.5 and ena to 0.9

Eijebong:


Marwes:
Thanks 👍

Dylan-DPC:
This is a breaking change though.

Dylan-DPC:
Oops. The dependencies weren't the problem. Some types of the members of the public structs had changed but when i looked at the source of code of ena, I realised that the new [type](https://github.com/rust-lang-nursery/ena/blob/master/src/unify/mod.rs#L184) is compatible with the old. I guess both versions are compatible so it isn't a breaking change.

PR #375, opened 2018-05-08
feat: Update to regex 1.0

Marwes:
This affects the generated lexer making it a breaking change.

BREAKING CHANGE

https://github.com/rust-lang/regex/releases/tag/1.0.0


Marwes:
Since we already need a breaking release with #353 I think this is fair to merge as well.

PR #384, opened 2018-06-17
Allow #![] attributes to work in 0.16

wycats:
In 0.16, the default mode for lalrpop will import the generated module
through a macro that `include!`s the contents.

PR #338 made it possible to use `include!` with vanilla modules by
moving all of the built-in whole-module attributes to individual
modules. However, #115 places user-generated outer attributes in the
same problematic location.

This PR moves user-supplied outer attributes to the same location as
the fix in #338.

### DIFF
--- a/lalrpop-snap/src/build/mod.rs
+++ b/lalrpop-snap/src/build/mod.rs
@@ -339,11 +339,17 @@ fn report_content(content: &Content) -> term::Result<()> {
     canvas.write_to(&mut stdout)
 }
 
-fn emit_module_attributes<W: Write>(
+pub fn emit_module_attributes<W: Write>(
     grammar: &r::Grammar,
     rust: &mut RustWrite<W>,
 ) -> io::Result<()> {
-    rust.write_module_attributes(grammar)
+    rust!(rust, "#[cfg_attr(rustfmt, rustfmt_skip)]");
+
+    for attribute in grammar.module_attributes.iter() {
+        rust!(rust, "#{}", &attribute[2..]);
+    }
+
+    Ok(())
 }
 
 fn emit_uses<W: Write>(grammar: &r::Grammar, rust: &mut RustWrite<W>) -> io::Result<()> {
@@ -376,7 +382,6 @@ fn emit_recursive_ascent(
     // includes things like `super::` it will resolve in the natural
     // way.
 
-    try!(emit_module_attributes(grammar, &mut rust));
     try!(emit_uses(grammar, &mut rust));
 
     if grammar.start_nonterminals.is_empty() {
--- a/lalrpop-snap/src/rust/mod.rs
+++ b/lalrpop-snap/src/rust/mod.rs
@@ -160,13 +160,6 @@ impl<W: Write> RustWrite<W> {
         Ok(())
     }
 
-    pub fn write_module_attributes(&mut self, grammar: &Grammar) -> io::Result<()> {
-        for attribute in grammar.module_attributes.iter() {
-            rust!(self, "{}", attribute);
-        }
-        Ok(())
-    }
-
     pub fn write_uses(&mut self, super_prefix: &str, grammar: &Grammar) -> io::Result<()> {
         // things the user wrote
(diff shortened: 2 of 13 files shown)


lalrpop-snap/src/rust/mod.rs
```diff
@@ -160,13 +160,6 @@ impl<W: Write> RustWrite<W> {
         Ok(())
     }
 
-    pub fn write_module_attributes(&mut self, grammar: &Grammar) -> io::Result<()> {
```
nikomatsakis: we don't usually edit lalrpop-snap

lalrpop-snap/src/build/mod.rs
```diff
@@ -339,11 +339,17 @@ fn report_content(content: &Content) -> term::Result<()> {
     canvas.write_to(&mut stdout)
 }
 
-fn emit_module_attributes<W: Write>(
+pub fn emit_module_attributes<W: Write>(
```
nikomatsakis: we don't typically edit lalrpop-snap

lalrpop/src/build/mod.rs
```diff
@@ -322,11 +322,17 @@ fn report_content(content: &Content) -> term::Result<()> {
     canvas.write_to(&mut stdout)
 }
 
-fn emit_module_attributes<W: Write>(
+pub fn emit_module_attributes<W: Write>(
     grammar: &r::Grammar,
     rust: &mut RustWrite<W>,
 ) -> io::Result<()> {
-    rust.write_module_attributes(grammar)
+    rust!(rust, "#[cfg_attr(rustfmt, rustfmt_skip)]");
```
nikomatsakis: Something feels fishy about this. This attribute is going to apply to the *next* module thing that gets emitted ... oh, I see, and the for loop below is intentionally converting `#![foo]` into `#[foo]`?

nikomatsakis (changes requested):
I have some concerns about this change -- it feels surprising to emit the module attributes also on the internal modules. Maybe we should add some kind of explicit Edition support, if that is the main motivation?

nikomatsakis:
OK, re-reading your comment, I think I understand better. So there are two distinct things going on here:

- Converting `#![foo]` to `#[foo]` -- this may indeed be a good idea, given the problems around `#![]` attributes and `include!` (IIUC)
- Apply those attributes to all the generated submodules -- I'm less sure about this, but it *may* make sense? I'm not sure if there was a direct motivation for this.

nikomatsakis:
Ah, probably we apply the attributes to the submodules just because it's not obvious how to apply them .. otherwise?

Marwes:
> Ah, probably we apply the attributes to the submodules just because it's not obvious how to apply them .. otherwise?

Yep. https://github.com/rust-lang/rust/issues/18810

Marwes:
Thinking about it, perhaps we could wrap the generated module as

```
mod name {
    #![attr]
     // Rest of the code goes here
}
pub use name::{Parser, ..};
```

alercah:
Personally I think I'm inclined to the `mod_path!` workaround, because it allows attaching other attributes from the outside too:

```rust
#[deny(...)]
lalrpop_mod!(...);
```

franleplant:
Hi guys! I have a question, https://github.com/lalrpop/lalrpop/issues/115 allegedly allowed attributes in top of the grammar file, I tried with 0.16.x and it doesnt work, and then I see this PR. 

From what I understand, and, as a nice recap for future people, did this feature broke on 0.16.x and that is why this PR exists?

franleplant:
@nikomatsakis how about adding by default a `#[allow(clippy::all)]` to the synthesized module?

NyxCode:
For everyone who stumbled upon this issue:
the `lalrpop_mod!` macro can accept attributes like this: `lalrpop_mod!(#[allow(clippy::all)] grammar);`

wycats:
Closing this 0.16-era implementation. The codebase and maintainer direction have moved on substantially; if the need remains, it should be reconsidered against current master.

Issue #376, opened 2018-05-08
Update regex_syntax dependency


Marwes:
Currently we are using 0.4 but regex 1.0 uses regex_syntax-0.6 . Looks non-trivial to update at first look.

PR #385, opened 2018-06-20
Update regex_syntax to 0.6.0

snsmac:
Hello, i updated the regex_syntax version to 0.6.0
I'm not sure whats the connection between lalrpop and lalrpop-snap, so i only updated lalrpop. Lalrpop-snap does still use 0.4

Reference: #376


Marwes:
Thanks!

Issue #303, opened 2018-02-02
Extract state machine code into lalrpop-util.


ahmedcharles:
I'm thinking of taking the following approach:

1. Change the generated code incrementally to use a trait/impl, such that the state machine code becomes static.
2. Move the state machine code to `lalrpop-util`.

This could be done for any code that is sufficiently generic.

nikomatsakis:
Yes, I'd love to see this happen. I *think* this should be possible. I have to look over the code though to figure out where to start. It might be worth sketching out first what the trait should look like that LALRPOP generates.

At minimum, the trait will need:

- The symbol type, which combines the type of nonterminals etc
- The "reduce" function that executes action code

ahmedcharles:
Given that error recovery and custom grammar parameters are a thing, I'm not sure quite how to store the parameters in a struct in a way that lets them be used the way they currently are, but perhaps I'm just missing something. Note, I haven't tried the obvious solution of just storing them in the generated struct as is and using them that way.

@nikomatsakis Any ideas?

ahmedcharles:
Note: https://github.com/ahmedcharles/lalrpop/tree/extract has WIP changes to `lalrpop-util`.

nikomatsakis:
I've been working on this in this branch:

https://github.com/lalrpop/lalrpop/tree/ndm-extract-state-machine

The branch adds the state machine here:

https://github.com/lalrpop/lalrpop/blob/ndm-extract-state-machine/lalrpop-util/src/state_machine.rs

Next up is to finish this function:

https://github.com/lalrpop/lalrpop/blob/ae69af4717b374dc35c5bf60894c32be15336155/lalrpop/src/lr1/codegen/parse_table.rs#L158-L246

I was considering modifying the `state_machine.rs` definition to use newtypes, as well. Right now it has a lot of `i32`, it'd be nice to have newtypes like `State(i32)` or `Action(i32)` etc.

. I think i've written up the state machine, but

nikomatsakis:
I haven't had any time to hack on this lately -- if anyone is interested in getting involved, let me know! I can help guide you over [on gitter](https://gitter.im/lalrpop/Lobby).

PR #395, opened 2018-08-14
extract state machine

nikomatsakis:
This branch refactors everything so that we the actual parsing is done by a generic function in lalrpop-util, and we just generate the parsing table, action code, and a "glue impl" that binds it all together. This is much nicer to read (esp. the error handling code) and opens the door to more ambitious transformations (e.g., a "push-mode" parser that you feed tokens to, instead of one that pulls tokens from you).

This is slightly backwards incompatible, in that we now require tokens to be `T: Clone` no matter what. We used to only require that if the error recovery feature was used. (Well, I suppose we don't require that if you use the non-table-driven, recursive ascent codegen feature.)

Fixes https://github.com/lalrpop/lalrpop/issues/303

cc @Marwes @ahmedcharles


lalrpop-util/src/state_machine.rs
```diff
@@ -0,0 +1,671 @@
...
+pub type SymbolTriple<D> = (Location<D>, Symbol<D>, Location<D>);
+pub type ErrorRecovery<D> = ::ErrorRecovery<Location<D>, Token<D>, Error<D>>;
+
+pub struct Parser<D, I>
+where
+    D: ParserDefinition,
+    I: Iterator<Item = Result<TokenTriple<D>, ParseError<D>>>,
+{
+    definition: D,
+    tokens: I,
```
Marwes: If `tokens` were taken as an argument instead of being part of the parser itself we could reset and reuse the parser over multiple inputs. I might do that as a followup if it sounds reasonable.

lalrpop/src/grammar/free_variables/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::iter;
+use string_cache::DefaultAtom as Atom;
+
+mod test;
+
+/// Finds the set of "free variables" in something -- that is, the
+/// type/lifetime parameters that appear and are not bound. For
+/// example, `T: Foo<U>` would return `[T, U]`.
+pub trait FreeVariables {
+    fn free_variables(&self, type_parameters: &[TypeParameter]) -> Vec<TypeParameter>;
```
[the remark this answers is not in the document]
Marwes: Figured as much :).

lalrpop/src/lib.rs
```diff
@@ -6,6 +6,9 @@
 // too darn annoying to try and make them do so.
 #![cfg_attr(test, allow(dead_code))]
 
+#![feature(in_band_lifetimes)]
+#![feature(infer_struct_outlives)]
```
Marwes: Were you testing something here? (Hope this isn't required for anything).

lalrpop-util/src/state_machine.rs
```diff
@@ -0,0 +1,671 @@
...
+    /// default tokenizer, this will be a `usize`.
+    type Location: Clone + Debug;
+
+    /// Represents a "user error" -- this can get produced by
+    /// `reduce()` if the grammar includes `=>?` actions.
+    type Error;
+
+    /// The type emitted by the user's tokenizer (excluding the
+    /// location information).
+    type Token: Clone + Debug;
```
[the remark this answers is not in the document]
Marwes: I'd just do that as a followup if anyone actually has non-cloneable tokens (which seems rare).

lalrpop/src/lib.rs
nikomatsakis: Oh yeah, let me revert that :)

lalrpop-util/src/state_machine.rs
nikomatsakis: Sure, that'd be fine. I guess there is a little bit of state we could reuse (e.g., vectors for stacks and things)

Marwes:
@nikomatsakis Is there a lot of work remaining on this? If not I may try and rebase and finish this soon-ish.

PR #398, opened 2018-08-27
feat: Allow attributes to be specified in lalrpop_mod!

Marwes:
Seems to be intentional that attributes on the macro invocation does not
work so we instead allow attributes inside it.

https://github.com/rust-lang/rust/issues/28284

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-- 1.20.0
+- 1.26.0
 - beta
 - nightly
 cache:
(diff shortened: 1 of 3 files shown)


Marwes:
Rust 1.20 seems to not work with some newer dependencies, I think we can just update the minimum supported version since we must do a breaking release anyway.

Nightly fails due to conflicts with `Iterator::find_map` which just needs a `cargo update` to get petgraph updated (also relevant for #395 ).

Dylan-DPC:
@Marwes yeah i'm fine with updating the minimum version.

PR #406, opened 2018-09-22
Process escape sequences in string literals appearing as terminals.

### LINKED ISSUE #405: Escapes in string literals are broken
Characters escaped by backslashes are not handled properly in string literals. For example, the following grammar accepts strings of two backslashes, rather than strings of one backslash. There is no way to accept strings of one backslash using LALRPOP's built-in lexical analyzer generator.
```
grammar;

pub Expr: String = {
    <slash: "\\"> => slash.to_owned()
}
```
Internally, `Tokenizer::string_literal` parses that as `Tok::StringLiteral(r#"\\"#)` (in Rust syntax), which eventually makes its way to `lexer::re::parse_literal`, which then tries to escape it again, resulting in the regular expression `r#"\\\\"#`.

jimblandy:
The "\\" and "\"" escapes are necessary to make it possible to put arbitrary text in a string literal at all. We have to support "\t" and "\n" to compile the `whitespace` example from the book. But the other Rust escape sequences like "\u{2a}" don't really seem necessary.

Fixes #405.

### DIFF
--- a/lalrpop/src/tok/mod.rs
+++ b/lalrpop/src/tok/mod.rs
@@ -757,6 +757,7 @@ pub fn apply_string_escapes(code: &str, idx0: usize) -> Result<Cow<str>, Error> 
                 ch = match next_ch {
                     '\\' | '\"' => next_ch,
                     'n' => '\n',
+                    'r' => '\r',
                     't' => '\t',
                     _ => { return error(UnrecognizedEscape, idx0 + offset); }
                 }
--- a/lalrpop/src/tok/mod.rs
+++ b/lalrpop/src/tok/mod.rs
@@ -430,7 +430,7 @@ impl<'input> Tokenizer<'input> {
         // we have to scan ahead, matching (), [], and {}, and looking
         // for a suitable terminator: `,`, `;`, `]`, `}`, or `)`.
         // Additionaly we had to take into account that we can encounter an character literal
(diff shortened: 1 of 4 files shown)


lalrpop/src/tok/mod.rs
```diff
@@ -736,3 +738,31 @@ fn is_identifier_start(c: char) -> bool {
...
+        let mut iter = code.char_indices();
+        let mut text = String::new();
+        while let Some((_, mut ch)) = iter.next() {
+            if ch == '\\' {
+                // The parser should never have accepted an ill-formed string
+                // literal, so we know it can't end in a backslash.
+                let (offset, next_ch) = iter.next().unwrap();
+                ch = match next_ch {
+                    '\\' | '\"' => next_ch,
+                    'n' => '\n',
```
Marwes: `\r` seems fair to include as well as it shouldn't be to rare.

jimblandy: Okay, that's implemented in bfa33d2.

Marwes:
Sorry for the delay here. Thanks for the PR! I am going to try and do one hopefully small change and then release 0.16.1

jimblandy:
Excellent! Really looking forward to the release. (I want to use \ as my lambda operator, like Haskell. :) )

PR #413, opened 2018-10-20
feat: Avoid including test parsers in the snapshot

Marwes:
This adds a build.rs file which uses an earlier compiled lalrpop executable to generate the grammar file containing lalrpop's test-only parsers. Thanks to not including these 2 parsers (out of 5) we can significantly reduce the size of grammar file (which also reduces the compiletime of lalrpop for users).

The downside of this is that `cargo test` will not work out of the box as one must first build a `lalrpop` executable which can then generate the grammar used for tests. This is not to onerous however as the compile error will point directly to a comment explaining the hickup letting it easily be fixed after which the build script transparently handles the rest.

As an extra this also adds `--no-whitespace` which trims the worst of the whitespace from the generated grammar, removing an additional 0.5 Mb.

- `cargo build` always works (since it does not need test parsers)
- `cargo test` will work after the first attempt to compile (as a lalrpop executable will have been created)

### Depends on:
Closes #412 
Closes #411


Marwes:
https://github.com/lalrpop/lalrpop/pull/414 is a better solution

PR #414, opened 2018-10-21
feat: Shrink LALRPOPs own parser to 1Mb

Marwes:
This shrinks the generated code to 1Mb (from 2Mb WITH the cfg approach
in #413) while also removing the cfg'ing itself as that barely reduces the
size. (Thus this supersedes #413)

While this is a bit boilerplate-y, I'd say it is worthwhile.

We could generate this in LALRPOP itself with some work #304 (comment)


lalrpop/src/parser/mod.rs
```diff
@@ -1,54 +1,84 @@
...
+    Grammar(Grammar),
+    Pattern(Pattern<TypeRef>),
+    MatchMapping(TerminalString),
+    TypeRef(TypeRef),
+    GrammarWhereClauses(Vec<WhereClause<TypeRef>>),
+}
+
+pub type ParseError<'input> = lalrpop_util::ParseError<usize, tok::Tok<'input>, tok::Error>;
+
+macro_rules! parser {
```
windelbouwman: Cool idea!

lalrpop/src/parser/mod.rs
```diff
@@ -1,54 +1,84 @@
...
+    ($input: expr, $offset: expr, $pat: ident, $tok: ident) => {{
+        let input = $input;
+        let tokenizer =
```
windelbouwman: This was my question on how to feed the extra start token into the parser, but this actually makes sense!

Issue #104, opened 2016-03-20
Problem with generics


alex-caelus:
use lalrpop_util::ParseError;

    use generator::Generator;

    grammar<Gen> where Gen: Generator;

    extern {
        type Error = String;
    }

    pub Schema : String = "grammar" "{" <id:r"[a-zA-Z0-9]*"> "}" =>? match Generator::schema(id) {
        Ok(sch) => Ok(sch),
        Err(err) => Err(ParseError::User { error: err })
    };

gives me the following error:

    src\test.rs:29:20: 29:28 error: unable to infer enough type information about `_`; type annotations or generic parameter binding required [E0282]
    src\test.rs:29         match try!(__state0(input, &mut __tokens, __lookahead)) {
                                      ^~~~~~~~
    src\test.rs:29:15: 29:64 note: in this expansion of try! (defined in <std macros>)
    src\test.rs:29:20: 29:28 help: run `rustc --explain E0282` to see a detailed explanation

I've found this previous issue #34 where it was suggested that `__state0(...)` be called like `__state0::<Gen>(...)`. Perhaps that is the necessary fix?

nikomatsakis:
Hmm, yes, I suppose that is necessary.

nikomatsakis:
For the time being you could work around this by introducing a dummy argument of type `std::marker::PhantomData<Gen>`

nikomatsakis:
(I will try to work up a fix shortly though)

nikomatsakis:
This is actually mildly tricky to fix. However, if you update to the newest LALRPOP, the state-table codegen method (now the default) I think avoids the problem here.

The reason it is tricky is that, in recursive ascent mode, we add an extra type parameter (`__TOKENS`) to states, and it's not always clear what value to pass for that -- but if we pass any type parameters, we need to pass that one. It occurs to me that maybe we can do it by passing a `PhantomData

nikomatsakis:
(Ah, no, it won't work there either.)

PR #418, opened 2018-11-11
Fix type annotation for inline actions

### LINKED ISSUE #417: Non-parameterized nonterminals cannot be used in built-in macros
A parameterized grammar such as:

```
use field::Field;

grammar<T> where T : Field;

pub Test : bool = <p:"test"?> => p.is_some();
```
causes a type error in any instance where a non-parameterized nonterminal is passed to a macro such as `*, +, ?`

The relevant generated code is 

```
#[allow(unused_variables)]
fn __action2<
    'input,
    T,
>(
    input: &'input str,
    (_, __0, _): (usize, &'input str, usize),
) -> ::std::option::Option<&'input str> where
  T:  Field,
{
    Some(__0)
}
```

and the compiler fails with 

```
error[E0283]: type annotations required: cannot resolve `_: field::Field`
    |
478 |     let __temp0 = __action2(
    |                   ^^^^^^^^^
    |
```

This looks similar to #104

tjade273:
This fixes #417. Grammars that take a type argument were not annotating inline actions, causing inline actions that don't use the type argument to not compile. 
See also #104 for a similar issue in non-inlined actions


tjade273:
I guess this fails CI since it changes the generated code in the self test by adding empty type annotations. I could tweak this to only generate annotation for nonempty type lists, but it would be somewhat less elegant

lalrpop-test/src/generics_issue_417.lalrpop
```diff
@@ -0,0 +1,11 @@
...
+
+grammar<T> where T: Generator;
+
+extern {
+    type Error = String;
+}
+
+Id : String = <id:r"[a-zA-Z0-9]*"> => T::schema(id);
+
+pub Tuple : (String, String) = "(" <(<Id> "," <Id>)> ")";
```
Marwes: Nit: Missing newline at EOF (makes git slightly confused if text is appended in a later commit).

lalrpop/src/build/action.rs
```diff
@@ -359,12 +364,17 @@ fn emit_inline_action_code<W: Write>(
...
                     temp_counter,
                     grammar.prefix,
-                    inlined_action.index()
+                    inlined_action.index(),
+                    lparen
                 );
+                for t in grammar.non_lifetime_type_parameters() {
+                    rust!(rust, "{},", t);
+                }
+                if annotate {rust!(rust, ">(")};
```
Marwes: Might be a bit cleaner to go

```rust
if annotate {
    rust!(rust, "(");
} else {
    rust!(rust, "::<")};
    for t in grammar.non_lifetime_type_parameters() {
        rust!(rust, "{},", t);
    }
    rust!(rust, ">(");
}
```

lalrpop/src/build/action.rs
```diff
@@ -396,8 +406,11 @@ fn emit_inline_action_code<W: Write>(
             }
         }
     }
-
-    rust!(rust, "{}action{}(", grammar.prefix, data.action.index());
+    rust!(rust, "{}action{}{}", grammar.prefix, data.action.index(), lparen);
+    for t in grammar.non_lifetime_type_parameters() {
+        rust!(rust, "{},", t);
+    }
+    if annotate {rust!(rust, ">(")};
```
Marwes: Ditto

lalrpop-test/src/generics_issue_417.lalrpop
tjade273: Fixed

lalrpop/src/build/action.rs
tjade273: This also breaks the CI, since the parenthesis is added on a new line now. Which, as I think about it is exactly the reason that I went with the weird conditional formatting. I can just be OK with changing the default output code, since it is still the same semantically. Should I regenerate the snapshot?

tjade273: Of course if I'm going to regenerate the snapshot then it maybe makes sense to unconditionally add the type annotation and just have it be empty most of the time.

Marwes: Ah, didn't realize it would put it on a separate line. Just reset to the original code in that case.

Marwes:
Released as 0.16.2

PR #422, opened 2018-11-29
extract state machine

### LINKED ISSUE #303: Extract state machine code into lalrpop-util.

Marwes:
Rebase of #395 

This branch refactors everything so that we the actual parsing is done by a generic function in lalrpop-util, and we just generate the parsing table, action code, and a "glue impl" that binds it all together. This is much nicer to read (esp. the error handling code) and opens the door to more ambitious transformations (e.g., a "push-mode" parser that you feed tokens to, instead of one that pulls tokens from you).

This is slightly backwards incompatible, in that we now require tokens to be T: Clone no matter what. We used to only require that if the error recovery feature was used. (Well, I suppose we don't require that if you use the non-table-driven, recursive ascent codegen feature.)

Fixes #303 
Closes #395


Marwes:
Rebasing gives a green build at least but the current nightly seems to have a ICE introduced on it https://github.com/rust-lang/rust/issues/56350

PR #431, opened 2019-01-03
Don't make generated files read-only

### LINKED ISSUE #426: cargo fix does not work out of the box due to read only output files
Since lalrpop outputs makes the output grammar files read only cargo fix will fail to update those files, resulting in partially applied fixes. Workaround is to generate the files first, make the writeable and then run cargo fix.

dflemstr:
Since `lalrpop` now defaults to outputting files into `$OUT_DIR`, there is not much sense in marking generated files as read-only.  This behavior additionally interferes with `git` and `cargo fix` anyway.

Fixes #426.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -108,7 +108,6 @@ fn process_file_into(
         if let Some(parent) = rs_file.parent() {
             try!(fs::create_dir_all(parent));
         }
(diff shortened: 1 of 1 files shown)


Marwes:
👍 Agree. Also, it is still possible to generate the file in the directory itself but having it readonly is just a nice-to-have so when it causes actual issues we should just remove the behaviour.

PR #432, opened 2019-01-05
Allow trim_left in the generated code

### LINKED ISSUE #428: Deprecation warning for trim_left on nightly
```
warning: use of deprecated item 'core::str::<impl str>::trim_left': superseded by `trim_start`
    --> /home/louispilfold/src/gleam/gleam/target/debug/build/gleam-80b0fb72690a65bb/out/grammar.rs:8
452:36
     |
8452 |             let __text = self.text.trim_left();
     |                                    ^^^^^^^^^
     |
     = note: #[warn(deprecated)] on by default
```

Is there a way to disable deprecation warnings on the generated code?

Marwes:
Fixes #428

### DIFF
--- a/lalrpop/src/lexer/intern_token/mod.rs
+++ b/lalrpop/src/lexer/intern_token/mod.rs
@@ -175,6 +175,7 @@ pub fn compile<W: Write>(
     rust!(out, "fn next(&mut self) -> Option<Self::Item> {{");
 
     // start by trimming whitespace from left
+    rust!(out, "#[allow(deprecated)]");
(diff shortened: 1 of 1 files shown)


alercah:
The changes in this PR should be split out into multiple, so that they can be evaluated individually. As-is, this doesn't build.

alercah:
It looks like the failure is due to the upgrade on the `rand` crate. It should be upgraded to avoid pulling in the older, backwards-incompatible 0.5, but it needs code fixes too.

nikomatsakis:
Seems like a simple step =)

PR #433, opened 2019-01-10
Support conditional execution by Cargo.

alercah:
Cargo allows build scripts to emit directives to specify their
dependencies, so that modifying other files does not cause them to be
rerun. This adds a configuration option to emit those directives,
defaulting to false to avoid interfering with other parts of the build
script and leading to hard-to-debug issues.

On Windows, some files cannot be represented in the UTF-8 used by Cargo,
and a build warning is emitted instead.

I've tested this change manually, but didn't see an easy way to test
it programmatically as part of the codebase, especially because
`to_string_lossy()` can never return `None` on Unix platforms.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -98,6 +98,7 @@ fn process_file_into(
     rs_file: &Path,
     report_file: &Path,
 ) -> io::Result<()> {
+    session.emit_rerun_directive(lalrpop_file);
     if session.force_build || try!(needs_rebuild(&lalrpop_file, &rs_file)) {
         log!(
             session,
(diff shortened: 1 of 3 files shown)


Marwes:
re-running travis nightly

lalrpop/src/build/mod.rs
```diff
@@ -1,562 +1,569 @@
...
-              p = grammar.prefix);
-        rust!(rust, "}}"); // match
-        rust!(rust, "}}"); // fn
-        rust!(rust, "}}"); // impl
-    }
-
-    Ok(())
-}
+//! Utilies for running in a build script.
+
```
nikomatsakis: what...is going on in this file?

alercah: Apparently the previous version had CRLF terminators. I'll pull from this branch.

nikomatsakis:
@alercah can you rebase this now that I merged https://github.com/lalrpop/lalrpop/pull/437 ?

alercah:
Done.

PR #434, opened 2019-01-10
 perf: Avoid some memcpying in reductions

Marwes:
In gluon's parser I have a lalrpop symbol type that ends up at 400Kb
which is expensive to copy around. By moving the `push` call into each
reduction instead we make it easier for LLVM to elide memcpys and I
observed a reduction in memcpy uses from 27% to 18% of of the total
parse time.


Marwes:
First commit just gets lalrpop compiling locally for me

PR #436, opened 2019-01-10
Detect when we are building our own workspace.

alercah:
This is possibly one of the most horrifying things I have ever come up
with. By putting two versions of a crate into different branches, and
using [replace] in Cargo.toml to override the dependency onto the other
branch, we can detect whether the root build is in the lalrpop
workspace.

This ensures that lalrpop does not get built more than once when used as
a dependency. This is clearly "better" than the old system.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,7 @@ dependencies = [
  "string_cache 0.7.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "term 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "unicode-xid 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "workspace-build 0.1.0 (git+https://github.com/alercah/lalrpop.git?branch=workspace-build-no)",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Marwes:
:/ Didn't think about path includes when doing that hack so `target` detection seemed decent enough as a start.

There is always the worse-is-better approach. We could only make lalrpop rebuild the snapshot if a feature flag is passed.

alercah:
Yeah, that's probably a better choice. This one's just more hilarious.

nikomatsakis:
The goal here is to avoid the snapshot stage, right? I was planning to just manually execute the lalrpop executable on its own file, and check-in the generated code. Or maybe I'm confused. This is some deep cargo voodoo =)

alercah:
I don't know what the snapshot stage is :P

The cargo voodoo is to make the `workspace_build` package (which you can see on the branches in my repo, but really they are nothing special, just two different versions of the same crate that differ by `s/true/false`) let you figure out if the `lalrpop` workspace is the root or not. That was based on the comment `// If compiling lalrpop itself, enable test parsers`. So both that behaviour, and the behaviour of rebuilding `lalrpop` itself if it detects that its output binary changes (which I presume is to try to aim for bootstrap consistency by running it until it doesn't change?), are conditional on that constant.

But this is an awful, hacky kluge. A feature flag would make more sense, almost certainly, or perhaps two, but I didn't want to dig into how important the test parser behaviour was.

I have no attachment to this, partway through it largely became an exercise on figuring out whether it was in fact possible. So I'm more than happy if you want to close this off. That said, it should be stable as long as I don't delete the branches in my repo (and they could be replicated in this repo as well, or even moved to another repo, or whatever is best).

PR #438, opened 2019-01-11
chore: Don't try to infer whether we are building in the lalrpop work…

Marwes:
…space

Instead we use --all-features to enable a special `test` feature to run tests with a freshly generated lrgrammar.rs. Since we also have a test to check that the saved lrgrammar.rs is identical to what lalrpop generates we should be ok with this approach.

Closes #436

### DIFF
--- a/lalrpop/build.rs
+++ b/lalrpop/build.rs
@@ -13,7 +13,9 @@ fn main() {
 
 fn find_lalrpop_binary(prefix: &str) -> Option<PathBuf> {
     let lalrpop_path = Path::new(prefix)
-        .join("target/debug/lalrpop")
+        .join("target")
+        .join(env::var("PROFILE").unwrap())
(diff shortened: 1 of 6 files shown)


Marwes:
cc @alercah

alercah:
Much better than my idea XD

lalrpop/build.rs
```diff
@@ -11,6 +11,17 @@ fn main() {
     }
 }
 
+fn find_lalrpop_binary(prefix: &str) -> Option<PathBuf> {
+    let lalrpop_path = Path::new(prefix)
+        .join("target/debug/lalrpop")
```
alercah: Should this use the profile environment variable rather than hard coding debug?

Marwes: 👍  Fixed.

PR #440, opened 2019-01-19
Use str::trim_start instead of str::trim_left

dlrobertson:
The method `str::trim_left` has been deprecated in favor of
`str::trim_start` and now throws a warning. Use `str::trim_start` instead of
`str::trim_left` and update test cases.


dlrobertson:
Ah, nvm... I see we need to be compatible with 1.26.

Issue #322, opened 2018-02-16
Add location for unexpected EOF errors


brendanzab:
It is sometimes helpful to be able to have the location ofthe EOF for error reporting purposes. Currently the `UnrecognizedToken` looks like this:

```rust
#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
pub enum ParseError<L,T,E> {
    ...
    UnrecognizedToken {
        token: Option<(L, T, L)>,
        expected: Vec<String>
    },
    ...
}
```

What would you thoughts be about switching to this?

```rust
#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
pub enum ParseError<L,T,E> {
    ...
    UnrecognizedEof {
        location: L,
        expected: Vec<String>
    },
    UnrecognizedToken {
        token: (L, T, L),
        expected: Vec<String>
    },
    ...
}
```

It's a breaking change, sadly, but I always found the old usage of `Option::None` for the EOF took a bit to wrap my head around, and that's reflected in the awkwardness of the documentation. At least here the point is more obvious in the name of the variant.

nikomatsakis:
As I [noted on gitter](https://gitter.im/lalrpop/Lobby?at=5a90171b012ff2bf681ccba1), I am not happy with our error handling strategy. I would like to change to something based on the `Failure` care. Perhaps @withoutboats can weigh in on what they view as the Best Practices. I was thinking of first creating independent structs for all the possible LALRPOP error cases today. Then there are two possibilities of where to go from there:

- Create an enum `LalrpopError` that includes those cases, but is marked as `#[non_exhautive]` (that's not currently possible, so we can do the "unmentionable variant" trick). This enum can include a `User` variant that carries no data. We can use causal chaining to attach the user's error.
- Forego the enum altogether and just return `failure::Error`
    - This is probably not a good idea. I don't think it's best practices.

This would allow us to remove the `type Error` business altogether, I think.

nikomatsakis:
*That said*, I guess you were talking about something more narrow. I'd be fine with having `UnrecognizedEOF` as a variant in any case. The location I guess would be the end of the final token?

brendanzab:
Yup, that's correct!

nikomatsakis:
@brendanzab think you would be up for opening a PR?

brendanzab:
[...]

PR #446, opened 2019-02-16
Split apart UnrecognizedEOF error variant from UnrecognizedToken

nwtnni:
Targets #322.

Not sure if this is still relevant, since the linked issue is from a year ago.

I have `lalrpop-util` compiling separately with the new error variant, but I don't understand the code generation logic in `lalrpop` well enough to fix the breakage. Would appreciate any help!


lalrpop-util/src/lib.rs
```diff
@@ -94,18 +104,31 @@ where
...
+                    for (i, e) in expected.iter().enumerate() {
+                        let sep = match i {
+                            0 => "Expected one of",
+                            _ if i < expected.len() - 1 => ",",
+                            // Last expected message to be written
+                            _ => " or",
+                        };
+                        try!(write!(f, "{} {}", sep, e));
+                    }
+                }
```
Marwes: Could you extract this into a function instead of duplicating it?

nwtnni:
I'm looking at a generated state function (from `lalrpop-test/src/loc.lalrpop`), but I don't see an easy way to extract the location of the final token given an unexpected EOF:

https://gist.github.com/nwtnni/b368d2477f695d292dd1269b667c2b55

It looks like we'd need to store the latest location in addition to the current lookahead?

Marwes:
> It looks like we'd need to store the latest location in addition to the current lookahead?

Yeah, that seems to be necessary. Which is rather awkward given how the recursive ascent parser is written. I suppose the lookahead could be changed from an option to something that always contains the latest location or the lookahead, or by passing another parameter. Whatever the solution this needs to be threaded through everywhere which may be a bit of a pain to thread through.

(I missed your question here, sorry!)

nwtnni:
No problem!

Do you think it's worth refactoring? I can try and give it a shot if you've got any pointers as to where to begin.

Marwes:
Actually, it shouldn't need the refactoring I mentioned. If you look at the gist you posted, the non-error case extracts `__end` from the last production that was parsed. In the case of no lookahead we can just extract the position in the same way.

https://gist.github.com/nwtnni/b368d2477f695d292dd1269b667c2b55#file-loc-rs-L17

The codegen seems to be done here https://github.com/lalrpop/lalrpop/blob/417f199ec8ec6b2b9fef5b41f819398af9e841ed/lalrpop/src/lr1/codegen/ascent.rs#L835

nwtnni:
I've implemented an initial version, but there are still a lot of test cases that don't compile.

It looks like some of the symbols passed into the state functions can be optional. Right now I only consider the last fixed symbol for a location, or `Default::default()` if there is no such symbol, but I was wondering if/how we should handle optional symbols.

The naive approach seems to mostly work, but there are [test cases where fixed symbols are shadowed][1].

[1]: https://gist.github.com/nwtnni/ef707d584c912b6ec8719cc1d3f58791#file-issue_394-rs-L20-L23

Marwes:
I think we need to locate the last symbol that isn't `None` and take end from that :/ Hopefully there is a way to detect statically that the symbol is an option however.

```rust
sym3.as_ref().map(|t| &t.2).unwrap_or_else(|| sym2.as_ref().map(|t| &t.2).unwrap_or_else(|| &sym1.2)).clone();
```

nwtnni:
I think I've fixed all of the breakage. Should I add some more test cases for UnrecognizedEOF now? (Sorry this is a little delayed; I've been traveling a lot recently.)

nwtnni:
That's weird--the latest commit builds and runs all tests on my laptop, but it something's going wrong with Travis. Do you have any idea why the CI build is broken?

Marwes:
> I think I've fixed all of the breakage. Should I add some more test cases for UnrecognizedEOF now? (Sorry this is a little delayed; I've been traveling a lot recently.)

Seems like there are some testcases for `UnrecognizedEOF` already so I think this is fine. Tried clearing the travis cache to see if that fixes the timeout.

nwtnni:
So it looks like all dependencies are being recompiled from scratch on the CI server, which is why the build is timing out. Any suggestions?

Marwes:
You could try adding `travis_wait` to the compile command that times out, not ideal but maybe it will finish then.

nwtnni:
Should I squash these into a single commit?

Marwes:
Did a squash and merge so no need.

I should be able to make a release in a couple of days, just want to try to use https://docs.rs/copyless/0.1.2/copyless/struct.VecAllocation.html to hopefully remove some memcpys in parse tables.

nwtnni:
Great. Thanks for walking me through the PR!

PR #447, opened 2019-02-21
Order of string for error message confusing

jespersm:
Before this patch, I was told:

```
src/cypher.lalrpop:309:37: 309:54 error: `nonterminal` is a AndExpression, not a macro
```

I think it should be the other way around ;-)


Marwes:
Thanks!

PR #450, opened 2019-03-03
Remove warning about hidden lifetime parameter

mikeyhew:
On Rust 1.33, the current generated code emits a warning because hidden lifetime parameters in types are deprecated. This occurs in the `impl Display for Token`, and happens because `fmt::Formatter` has a lifetime parameter.

This PR adds an explicit lifetime parameter `'f` to the generated code, so it will work on all supported versions of Rust.


Marwes:
👍 Thanks!

PR #454, opened 2019-04-28
Replace all uses of `try!` with `?`

ra-kete:
This PR replaces all uses of the `try!` macro with the `?` operator.

The `?` operator was added to replace the `try!` macro and should be
used in modern Rust code. `try` is also a reserved keyword in Rust 2018,
so this change prepares for a possible edition switch too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-- 1.26.0
+- 1.31.0
 - beta
 - nightly
 cache:
--- a/doc/whitespace/src/ast.rs
+++ b/doc/whitespace/src/ast.rs
@@ -97,8 +97,8 @@ impl<'program> Interpreter<'program> {
                 &Stmt::Push(n) => self.stack.push(n),
 
                 &Stmt::PrintChar => {
-                    let top = try!(self.pop());
-                    let c = try!(num_to_char(top));
+                    let top = self.pop()?;
+                    let c = num_to_char(top)?;
                     print!("{}", c);
                 }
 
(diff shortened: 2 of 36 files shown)


ra-kete:
The CI checks fail for two different reasons now:

- **1.26.0**: rustc-demangle, which mdbook depends on, requires Rust 1.31.0 now: https://github.com/alexcrichton/rustc-demangle/issues/25
  - Easiest way to fix this is to bump our minimum required rustc version to at least 1.31.0 too.
- **beta, nightly**: Multiline Debug representation (`{#?}`) now adds trailing commas: https://github.com/rust-lang/rust/pull/59076. Looks like the tests use this representation to compare expected output.
  - I found https://github.com/rust-lang/gll/issues/88 that's dealing with the same problem. They fixed it by simply removing all trailing commas from the Debug output before comparing.

Any thoughts on how to proceed here?

ra-kete:
I added PR #455 that fixes the beta/nightly tests as described above.

ra-kete:
Okay, I can do that. Can I do it as an independent PR? Or does it have to be part of #455 because you cannot merge over failing CI runs on beta/nightly?

ra-kete:
Just read your comment again: Updating the Rust version won't help fixing the Debug issue, since that changed hasn't landed in stable yet.

Marwes:
Ah, I misunderstood about the debug impl started to fail the build. Could you bump the version in #455 so we get a green build and then rebase this which should give us a green build here as well?

Marwes:
Thanks!

PR #458, opened 2019-04-30
Replace deprecated `trim_left` with `trim_start`

ra-kete:
Now that our minimum supported Rust version is 1.31.0, we can make the switch from `String::trim_left` (deprecated since 1.33.0) to `String::trim_start`.

This also lets us get rid of the `allow(deprecated)` introduced as a workaround for issue #428.


Marwes:
👍 Thanks for fixing this! Was planning on doing this before the release!

PR #460, opened 2019-05-02
Apply rustfmt

ra-kete:
In my previous clippy PR (#459) I purposefully made the diffs as small as possible, to allow for easy review of the changes. Formatting, which was inconsistent to begin with, suffered somewhat from that.

This PR applies rustfmt on all .rs files to enforce a consistent formatting.


ra-kete:
I didn't find a way to also apply rustfmt to the action codes in .lalrpop files, since rustfmt isn't able to parse the LALRPOP syntax. Seems our only option there would be to manually do the formatting.

Marwes:
> I didn't find a way to also apply rustfmt to the action codes in .lalrpop files, since rustfmt isn't able to parse the LALRPOP syntax. Seems our only option there would be to manually do the formatting.

Yeah, no real way around that. This looks good though 👍

PR #469, opened 2019-06-22
fix: Accept `dyn Trait` as a type

Marwes:
Rust has started warning about trait objects without dyn on nightly.


Marwes:
Also accepts `dyn Fn(A) -> B` now. I omitted the old non `dyn` since it is deprecated.

Issue #427, opened 2018-12-10
Environment variable 'OUT_DIR' not defined


Cthutu:
I get this error while following your tutorial:

    error: environment variable `OUT_DIR` not defined
     --> src/main.rs:3:1
      |
    3 | lalrpop_mod!(pub calculator1);
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z     external-macro-backtrace for more info)
    error: couldn't read src/0/calculator1.rs: No such file or directory (os error 2)
     --> src/main.rs:3:1
      |
    3 | lalrpop_mod!(pub calculator1);
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z     external-macro-backtrace for more info)
    
    error: aborting due to 2 previous errors
    
    error: Could not compile `calculator`.
    

Any advice to fix this?

Marwes:
How are you compiling this? `OUT_DIR` should always be set by cargo.

Cthutu:
I am running `cargo test`.  I am running this on Window's Linux Subsystem.   I have the build.rs like this:

    extern crate lalrpop;
    
    fn main() {
        lalrpop::process_root().unwrap();
    }

And main.rs set up like the tutorial (with the unit tests).

Cthutu:
With verbose on I get the additional info:

    Caused by:
      process didn't exit successfully: `rustc --edition=2018 --crate-name calculator src/main.rs --color always --crate-type bin --emit=dep-info,link -C debuginfo=2 -C metadata=fdfadf2fcd2d22db -C extra-filename=-fdfadf2fcd2d22db --out-dir /home/matt/calculator/target/debug/deps -C incremental=/home/matt/calculator/target/debug/incremental -L dependency=/home/matt/calculator/target/debug/deps --extern lalrpop_util=/home/matt/calculator/target/debug/deps/liblalrpop_util-a3e1485f682c46b6.rlib --extern regex=/home/matt/calculator/target/debug/deps/libregex-5524183291022439.rlib` (exit code: 1)

Cthutu:
More info: using rustc 1.31.0 (stable)

Marwes:
[...]

PR #478, opened 2019-08-18
It's not obvious where build.rs goes

AndresParraSilva:
I ran into the same problem that https://github.com/lalrpop/lalrpop/issues/427, so I propose this little change in the quick start guide to avoid it.


quick_start_guide.html
```diff
@@ -147,7 +147,7 @@ <h1 class="menu-title"></h1>
 [build-dependencies]
 lalrpop = &quot;0.17.1&quot;
 </code></pre>
-<p>Next create a <code>build.rs</code> file that looks like:</p>
+<p>Next in the same root folder of the project create a <code>build.rs</code> file that looks like:</p>
```
Marwes:
```suggestion
<p>Next in the root folder of the crate create a <code>build.rs</code> file that looks like:</p>
```

PR #486, opened 2019-10-23
depend on a more recent version of the unicode-xid crate

dkg:
unicode-xid is at version 0.2.0 in debian unstable.  I want to make sure lalrpop can work with it!

This is an aspirational PR -- i want to see what the CI does with it, and i would love to hear feedback from the lalrpop devs about any concerns that they have about this change.

### DIFF
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -30,7 +30,7 @@ serde = "1.0"
 serde_derive = "1.0"
 string_cache = "0.7.1"
 term = "0.5"
-unicode-xid = "0.1"
+unicode-xid = "0.2"
 sha2 = "0.8.0"
 
 [dev-dependencies]


dkg:
It looks to me like the changes succeeded in the CI on everything but rust-nightly, and rust-nightly is failing for some reason other than unicode-xid.

however some of the logs (e.g. https://travis-ci.org/lalrpop/lalrpop/jobs/601573971) appear to contain references to both unicode-xid 0.1.0 and 0.2.0, which confuses me.  Do i need to update `Cargo.lock` as well?  i don't like the idea of mixing two different versions of a crate in the dependencies.

Marwes:
The nightly failure looks like it is fixed on cargo's master, just waiting on a new nightly.

unicode-xid 0.1 might be used by another dependency which would also need to be updated to 0.2 if you want to get rid of it in lalrpop. You can look through `Cargo.lock` for `unicode-xid 0.1` to see who depends on it.

Marwes:
The lock file should also be updated, we should be running CI with `--locked` to make sure that it is updated.

dkg:
@Marwes thanks for taking a look.  it sounds like the upgrade to unicode-xid 0.2.0 is not problematic for lalrpop, at least :)

Do you want me to submit new patches to this series to update `Cargo.lock`?  I certainly won't mind if you make whatever changes you think are appropriate.  then i can review them and learn from them for any future contributions along these lines.

dkg:
lalrpop 0.17.2 has been [building on debian](https://buildd.debian.org/status/logs.php?arch=&pkg=rust-lalrpop) with [unicode-xid 0.2](https://sources.debian.org/src/rust-lalrpop/0.17.2-1/debian/patches/unicode-xid-bump-0.2.patch/) for months with no problem.  Any suggestions for what specifically should be changed here to make this patch acceptable to lalrpop upstream?

Marwes:
Just needed to run the nightly build to see that it was fixed 👍  (Had forgotten about this)

PR #487, opened 2019-11-02
Update lalrpop_mod!() to match on any $vis pattern

Michael-F-Bryan:
I just tried to use `lalrpop_util::lalrpop_mod!(pub(crate) grammar)` and got a `no rules expected the token "("` error during macro expansion.

By the looks of it, as well as matching the normal `ident` and `expr` "metavariables", `macro_rules!` macros can use [a `vis` metavariable][1] to match a (possibly empty) visibility specifier like `pub` or `pub(crate)`.

This PR updates the macro definition to use `vis` instead of manually matching on the private and `pub` variants.

[1]: https://doc.rust-lang.org/reference/macros-by-example.html#metavariables

### DIFF
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -184,20 +184,12 @@ pub struct ErrorRecovery<L, T, E> {
 
 #[macro_export]
 macro_rules! lalrpop_mod {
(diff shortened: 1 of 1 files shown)


Marwes:
Thanks!

Issue #480, opened 2019-08-25
Generating bad import statements with super::


valarauca:
When a `lalrpop` file is specified in the format

```rust
use super::my_module::{MyType};
grammar;
```

It appears `lalrpop` will copy that character-for-character to use in its internal generated modules. 

---

An example of this can be found in [this commit](https://github.com/valarauca/dice_backend/tree/90bb1aec2e17c4721dbf8ea7df82f52317ffea11). 

Where the [`value.lalrpop`](https://github.com/valarauca/dice_backend/blob/90bb1aec2e17c4721dbf8ea7df82f52317ffea11/src/value.lalrpop#L1) file is generating a [`value.rs`](https://github.com/valarauca/dice_backend/blob/90bb1aec2e17c4721dbf8ea7df82f52317ffea11/src/value.rs#L3) file. But the internal generated structure [mod __internal_tokenizer](https://github.com/valarauca/dice_backend/blob/90bb1aec2e17c4721dbf8ea7df82f52317ffea11/src/value.rs#L16848-#L16850) requires modification to work with this definition as an additional `super::` is needed to correctly resolve the path. 

---

Using `lalrpop` v 0.17.2 on rustc 1.37.0 (eae3437df)

valarauca:
This only ocurrs within `mod __intern_token` it is handled correctly in all other modules.

PR #490, opened 2019-11-21
issues/480

valarauca:
the `write_uses` call as defined within [lalrpop/src/rust/mod.rs](https://github.com/lalrpop/lalrpop/blob/c531b74798a4b2640784cb4909f32e294ba652d8/lalrpop/src/rust/mod.rs#L130-#L141) requires a prefix be passed to it to populate the output. It appears based on the decision at lines [133 & 134](https://github.com/lalrpop/lalrpop/blob/c531b74798a4b2640784cb4909f32e294ba652d8/lalrpop/src/rust/mod.rs#L133-#L134) if the prefix is an empty string, it will output nothing (even if an output is required). 

This patch aims to provide a default string which can be prefixed when required.

The goal is to address: https://github.com/lalrpop/lalrpop/issues/480

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -318,7 +318,7 @@ fn emit_module_attributes<W: Write>(
 }
 
 fn emit_uses<W: Write>(grammar: &r::Grammar, rust: &mut RustWrite<W>) -> io::Result<()> {
-    rust.write_uses("", grammar)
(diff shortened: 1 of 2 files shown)


valarauca:
ping @Marwes

armoha:
Since issue #480 is closed by PR https://github.com/lalrpop/lalrpop/pull/601, I think this PR can be closed.

PR #496, opened 2019-12-27
Doc improvement

Sosthene-Guedon:
I wrote a few improvements to the LALRPOP book, mainly around passing state parameter and lifetime parameters.


Marwes:
Looks good to me! Thanks!

Issue #229, opened 2017-05-03
enable byte-based regular expressions


nikomatsakis:
LALRPOP currently only supports full unicode regular expressions. We should support the byte-based ones, which can be significantly simpler/faster. This probably means two distinct things. As a first step, it just means permitting the `(?-u)` regex to appear in LALRPOP files, while still only supporting matches on `&str`. (As a later step, and I think a separate bug, we should allow the user to then opt-in to matching on `&[u8]` inputs; I opened https://github.com/nikomatsakis/lalrpop/issues/230 for that.) 

For example, this is a version of `lalrpop-tests/src/match_section.lalrpop` but where each regex is set to ignore unicode with the `-u` flag:

```rust
grammar;

// NOTE: This grammar is ambiguous without the match section

match {
    r"(?i-u)select",
    r"(?i-u)insert" => "INSERT",
    r"(?i-u)update" => UPDATE
} else {
    _
}

Keyword: String = {
    r"(?i-u)select" => String::from("SELECT"),
    "INSERT" => String::from("INSERT"),
    UPDATE => String::from("UPDATE")
};

Table: String = {
    <r"(?i-u)[a-z]+"> => String::from(<>)
};

pub Query: String = {
    <Keyword> <Table> => format!("{} {}", <>)
};
```

If you try to build this, it will fail. This is because the NFA construction is not [doesn't support the various byte-based things that `regex_syntax` gives back](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/lexer/nfa/mod.rs#L369-L376):

```rust
            // If we ever support byte regexs, these
            // can be merged in with the cases above.
            Expr::AnyByte |
            Expr::AnyByteNoNL |
            Expr::ClassBytes(_) |
            Expr::LiteralBytes { .. } => {
                Err(NFAConstructionError::ByteRegex)
            }
```

Handling these cases ought to be fairly straightforward. For example, `Expr::AnyByte` and `Expr::AnyByteNoNL` are probably **exactly the same** as the existing `AnyChar`, I would think. Similarly, `ClassBytes` ought to work *almost* the same, except that we have to implement `From<ByteRange> for Test` to supplement the existing `From<ClassRange> for Test`:

```rust
impl From<ClassRange> for Test {
    fn from(range: ClassRange) -> Test {
        Test::inclusive_range(range.start as u32, range.end as u32)
    }
}
```

Similarly with `LiteralBytes`.

cc @wieczyk -- interested in giving this a try?
[...]

PR #498, opened 2020-01-06
enable byte-based regular expressions #229

wash2:
This patch aims to provide support for byte-based regular expressions as discussed in #229 

I used the example provided as a test, but I am open to adding more tests if need be. I would also appreciate any constructive criticism.


wash2:
It looks like CI succeeded for beta and nightly, but failed on stable (1.32.0) because of an error when installing mdbook. It's unclear to me why this would be related to the changes I've made.

Marwes:
Updating to 1.36 as minimum on CI should fix it https://doc.rust-lang.org/std/mem/union.MaybeUninit.html . Looks like `SmallVec` broke its minimum supported rust version.

lalrpop-test/src/match_section_byte.lalrpop
```diff
@@ -0,0 +1,25 @@
...
+    UPDATE => String::from("UPDATE")
+};
+
+Table: String = {
+    <r"(?i-u)[a-z]+"> => String::from(<>)
+};
+
+pub Query: String = {
+    <Keyword> <Table> => format!("{} {}", <>)
+};
```
Marwes: Nit: Missing newline

wash2:
> Updating to 1.36 as minimum on CI should fix it https://doc.rust-lang.org/std/mem/union.MaybeUninit.html . Looks like `SmallVec` broke its minimum supported rust version.

I'm kinda new to using CI. How can I set 1.36 as minimum?

lalrpop-test/src/match_section_byte.lalrpop
wash2: just added the newline.

Marwes:
Nice!

PR #502, opened 2020-01-20
Doc link fixes

TheBeggs:
Add the correct directory to the start of the links. Links should also now have the correct extension.


sirxyzzy:
This closes #534

PR #503, opened 2020-02-05
Support redefining how repeater vecs are built

withoutboats:
I'm using lalrpop in a personal project, but I want the AST parsed by LALRPOP to be allocated into a particular arena. For that reason, I need the ability to declare that the vectors constructed by the `*` and `+` repeater macros are my own vector type, rather than std's.

The current design is very hacky and enough to meet my needs, would be glad to receive feedback on how it should be changed before possibly merging it into master.

The `grammar` item can be followed by a new `vec_builder` item, which looks like this:

```rust
grammar() vec_builder { "expr" Path 'lifetime? }; // the lifetime is optional
```

This allows the user to specify the vector constructor as a string of Rust code, the path of the vector type, and - optionally - an additional lifetime parameter. The vector is then expected to take one type parameter and have a push method (this is all duck typed of course and only checked by the compiler when building the generated code).

For example, this is how I use this feature:

```rust
use bumpalo::Bump;
use bumpalo::collections::Vec as BumpVec;

grammar<'a>(arena: &'a Bump) vec_builder { "BumpVec::new_in(arena)" BumpVec 'a };
```

As I said, I don't think this is a good syntax. Would be interested in a suggestion of what could make more sense with lalrpop's overall design.


Marwes:
Restricting this to one type may be acceptable, but I think this should ideally allow for multiple different repetition types in the same grammar.

```rust
pub trait Repetition<Input> {
    type Item;
    fn new(grammar_input: Input) -> Self;
    fn push(&mut self, grammar_input: Input, item: Self::Item);
}

impl<Input, T> Repetition<Input> for Vec<T> {
    type Item  = T;
    fn new(grammar_input: Input) -> Self { Vec::new() }
    fn push(&mut self, grammar_input: Input, item: Self::Item) { }
}
```

Then you could implement this trait for any types which require it.

Internally, LALRPOP would need to generate repetition actions such that they are generic over `Repetition` instead of hard codec to `Vec` (I think repetitions in the grammar should not need to specify the type but I am not 100% on that).

Wtih this, I think there is no need for any special grammar constructs but I can't say that for certain.

Marwes:
@withoutboats Have you made any progress on this? I believe I could use to myself for allocating the AST into https://github.com/SimonSapin/rust-typed-arena (so I could help if there is need).

withoutboats:
No, I've just been working on my own project off of this fork. I haven't had time & interest to go back and do the refactor you proposed. If you're motivated to do it, please do!

Marwes:
Looked into this bit, it is possible but not as easy as I had hoped. While generating generic `action` functions would be possible, the type needs to be concrete when pushed to the symbol stack and there is no way of knowing the concrete type at that point unfortunately.

The silver lining is that `+` and `*` are just desugared as lalrpop macros so implementing custom repetitions isn't to bad. I implemented repetitions which create `&mut [T]` slices created from a `typed-arena` like so (with `temp_vecs` structure to reuse `Vec<T>` values for temporary storage).

```
Many1Inner<T>: Vec<T> = {
    T => {
        let mut v = temp_vecs.exprs.pop().unwrap_or_default();
        v.push(<>);
        v
    },
    <mut xs: Many1Inner<T>> <x: T> => {
        xs.push(x);
        xs
    }
};

#[inline]
ManyInner<T>: Vec<T> = {
    => Vec::new(),
    Many1Inner<T>,
};

#[inline]
Many<T>: &'ast mut Slice<T> = {
    => &mut [],
    Many1<T>,
};

#[inline]
Many1<T>: &'ast mut Slice<T> = {
    Many1Inner<T> => {
        let mut xs = <>;
        let slice = arena.alloc_extend(xs.drain(..));
        temp_vecs.exprs.push(xs);
        slice
    }
};
```

withoutboats:
I've modified my project to use the pattern above instead of the built in repeaters. Thanks :)

PR #504, opened 2020-02-20
Update book to regex v1.0

JustAPerson:
The v0.2 regex crate came out 3 years ago.

I was about to publish a crate of my own when I saw this pre-1.0 dependency in my `Cargo.toml` that I had copied from the book.


Marwes:
:+1:

PR #509, opened 2020-03-02
feat: Allow the tokenizer to contain custom skip regexes/literals

Marwes:
Makes it possible to implement comments in the lexer. Not the most efficient approach, but it works which is enough for me at least.

cc #14


benkay86:
This is incredibly helpful!  Thank you for taking the time to implement this!

Small issue, you left an `eprintln!` in [lalrpop-util/src/lexer.rs](https://github.com/Marwes/lalrpop/blob/lexer_comment/lalrpop-util/src/lexer.rs).

Otherwise got this working with in my `Cargo.toml` with:
```
[dependencies]
lalrpop-util = { git = "https://github.com/Marwes/lalrpop", branch = "lexer_comment", features = ["lexer"] }

[build-dependencies]
lalrpop = { git = "https://github.com/Marwes/lalrpop", branch = "lexer_comment", features = ["lexer"] }
```

evbo:
@Marwes just curious for a newcomer how is this different than implementing a custom Lexer?

dburgener:
@evbo Marwes hasn't been active lately, so you may not get a response directly from him.

To answer your question, the approach here does not require you to do the full work of implementing a custom lexer.  By default, the built-in lalrpop lexer will skip whitespace.  However, if your parser declares any ignore patterns, as described in the documentation added in this PR, then the default behavior will be turned off, and the lalrpop lexer will only skip the patterns that you explicitly ignore.

If you need more control than that, then you implement a custom lexer, as described in [the book](https://lalrpop.github.io/lalrpop/lexer_tutorial/003_writing_custom_lexer.html).  In that case, the lalrpop built-in lexer isn't used, and the approach here is irrelevant and unneeded.

PR #522, opened 2020-04-20
Tutorial: Add a section about fallible actions

Kingdread:
Hey!

I've recently used LALRPOP for a small personal project and noticed that the tutorial does not yet cover fallible actions (but it is on the todo-list in the tutorial overview). I've taken this opportunity to write up a small section about the use of `=>?` and hooking up a custom error enum to `ParseError` (which is also covered in the custom lexer tutorial, but in a different context).

I've put the section after the macros but before the error recovery, as it seemed more natural to cover custom errors before error recovery. Therefore I've also moved the numbers 6/7/8 up to 7/8/9.

Please let me know what you think!

Side note: I haven't worked with `mdbook` before, so I hope I didn't mess up any references or anything :)


Marwes:
Thanks! Looks good to me!

PR #523, opened 2020-04-29
perf(parse_table): Avoid generating unused rows in the GOTO matrix

Marwes:
Only the states in the beginning of a reduction are actually used, the
states for all other rows will never be used (really, no zero elements
will be accessed in the GOTO table so we might be able to do even better
without compromising performance).

By simply reordering the states so that states needed in the GOTO table
occupy the lower indices we can simply avoid generating the remaining
part of the GOTO table.

For the LALRPOP parser itself this is a reduction of about -40% (~100kB)
of the size of the parse tables (not counting the generated code)


brendanzab:
> For the LALRPOP parser itself this is a reduction of about -40% (~100kB)
of the size of the parse tables (not counting the generated code)

This is awesome! 👏

PR #529, opened 2020-05-14
Add From<E> for ParseError<L, T, E>

ndmitchell:
Useful especially combined with the `?` syntax, as you can do `MyLocalError(...)?` and it gets transformed into a `ParseError` without having to worry about writing out the wrapper.


Marwes:
:+1:

Issue #538, opened 2020-06-24
Tutorial Broken


HiDefender:
The current release breaks the tutorial. This seems to be a repeat of [513](https://github.com/lalrpop/lalrpop/issues/513).

```
error: failed to run custom build command for `calculator v0.1.0 (~/calculator)`

Caused by:
  process didn't exit successfully: `~/calculator/target/debug/build/calculator-8ea0fd66625d066c/build-script-build` (exit code: 1)
--- stdout
processing file `~/calculator/calculator1.lalrpop`
~/calculator/calculator1.lalrpop:1:1: 1:0 error: The `lexer` feature must be specified unless an `extern` lexer is defined

--- stderr
  use std::str::FromStr;
  ^
```

michaeljones:
I found that frustrating too. The solution seems to be to use: 
```
[build-dependencies]
lalrpop = { version = "0.19", features = ["lexer"] }
```
In your `Cargo.toml`. ie. you need the `lexer` feature enabled.

mhdirkse:
With version 0.19.8 of larlpop, the error messages are different. There are many errors like the following:

```
error[E0432]: unresolved import `self::__lalrpop_util::lexer`
  --> /home/martijn/rust/quadratic-equations-2d/target/debug/build/quadratic-equations-2d-3e9c9d94e3adc760/out/lang.rs:22:31
   |
22 |     use self::__lalrpop_util::lexer::Token;
   |                               ^^^^^ could not find `lexer` in `__lalrpop_util`
```

The proposed fix, adding feature `lexer`, still works.

Pat-Lafon:
The tutorial should be fixed now, open another issue if something comes up.

mhdirkse:
Thanks, @Pat-Lafon. I have a problem with my eyes that prevents me from doing development work. Otherwise I could have checked your work.

PR #540, opened 2020-06-29
Small documentation improvements for beginners

michaeljones:
These changes aim to help out people like me that are new to the project and new-ish to Rust. I hope they seem like reasonable changes. They should also help with #538.

### DIFF
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -24,7 +24,7 @@ regex = "1"
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
 lalrpop = "0.19.1"
-# If you are supplying your own external lexer you must disable default features so that the
+# If you are supplying your own external lexer you can disable default features so that the
 # built-in lexer feature is not included
 # lalrpop = { version = "0.19.1", default-features = false }
 ```
@@ -60,4 +60,3 @@ lalrpop file.lalrpop
 This will generate `file.rs` for you. Note that it only executes if
 `file.lalrpop` is newer than `file.rs`; if you'd prefer to execute
 unconditionally, pass `-f` (also try `--help` for other options).
-
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -23,9 +23,10 @@ regex = "1"
 
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
-lalrpop = { version = "0.19.1", features = ["lexer"] }
-# If you are supplying your own external lexer you do not need the `lexer` feature:
-# lalrpop = "0.19.1"
+lalrpop = "0.19.1"
+# If you are supplying your own external lexer you must disable default features so that the
+# built-in lexer feature is not included
+# lalrpop = { version = "0.19.1", default-features = false }
 ```
 
 Next create a `build.rs` file that looks like:
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -29,7 +29,8 @@ lalrpop = "0.19.1"
 # lalrpop = { version = "0.19.1", default-features = false }
 ```
 
-Next create a `build.rs` file that looks like:
+Next create a [`build.rs`](https://doc.rust-lang.org/cargo/reference/build-scripts.html) file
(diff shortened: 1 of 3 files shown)


doc/src/quick_start_guide.md
```diff
@@ -23,9 +23,9 @@ regex = "1"
 
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
-lalrpop = "0.19.0"
-# If you do not supply your own, external lexer you also need the `lexer` feature
-# lalrpop = { version = "0.18.1", features = ["lexer"] }
+lalrpop = { version = "0.19.0", features = ["lexer"] }
```
Marwes, line 26: TBH, we should just make "lexer" a default feature, that would fix the problems at the root.

doc/src/quick_start_guide.md
```diff
@@ -23,9 +23,9 @@ regex = "1"
...
+# If you are supplying your own external lexer you do not need the `lexer` feature:
+# lalrpop = "0.19.0"
```
Marwes, line 28:
```suggestion
# If you are supplying your own external lexer you do not need the `lexer` feature:
# lalrpop = { version = "0.19.0", default-features = false }
```

doc/src/tutorial/001_adding_lalrpop.md
```diff
@@ -56,8 +56,9 @@ hand and use it with LALRPOP, then check out the [lexer tutorial].)
 [the next section]: 002_paren_numbers.md
 [lexer tutorial]: ../lexer_tutorial/index.md
 
-Next we have to add `build.rs` itself. This should just look like the
-following:
+Next we have to add `build.rs` itself. For those unfamiliar with this feature, the `build.rs` file
+should be placed next to your `Cargo.toml` file and not inside the `src` folder with the rest of
+your Rust code. This should just look like the following:
```
brendanzab, line 61: I'm wondering if it would be possible to link to an appropriate place in the Cargo guide here 🤔

michaeljones: It turns out there is a link a few lines above but I'm not against making the first `build.rs` in this paragraph a link to the cargo page too (https://doc.rust-lang.org/cargo/reference/build-scripts.html)

doc/src/quick_start_guide.md
michaeljones: I've attempted to add this though I'm not that familiar with these parts of cargo & rust.

doc/src/quick_start_guide.md
michaeljones: I've adjusted this though with slightly different wording, if that ok.

doc/src/tutorial/001_adding_lalrpop.md
michaeljones: Thanks for the thought. I've added a commit to add more links to the cargo docs for the build.rs. I don't know when enough is enough but I am certainly happy to err on having too many than too few.

michaeljones:
Thanks for the reviews.

doc/src/quick_start_guide.md
```diff
@@ -24,11 +24,13 @@ regex = "1"
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
 lalrpop = "0.19.0"
-# If you do not supply your own, external lexer you also need the `lexer` feature
-# lalrpop = { version = "0.18.1", features = ["lexer"] }
+# If you are supplying your own external lexer you must disable default features so that the
+# built-in lexer feature is not included
```
Marwes, line 28:
```suggestion
# If you are supplying your own external lexer you can disable default features so that the
# built-in lexer feature is not included
```

michaeljones:
The CI failure doesn't seem to be related to my changes.

michaeljones:
I've rebased onto latest `master`. Though I guess now the recommendation to disable default features if you have your own lexer will result in the `pico-args` being disabled too? I don't know the best way forward here.

Marwes:
I'd expect most people to use this as a library, in which case you want `pico-args` disabled. If they do use it as a binary then they will get an error message asking for that feature so this should be fine.

michaeljones:
Thanks for merging!

PR #541, opened 2020-07-05
Add required "lexer" feature to tutorial

mcon:
This is required to get the tutorial working (without a custom lexer) - as a first time user I guess the default would be use the builtin lexer, though I appreciate there may be a reason for this not being the default.


mcon:
Just noticed https://github.com/lalrpop/lalrpop/pull/540 already proposes this change, closing in favour of that.

PR #555, opened 2020-10-31
Enable precedence annotations

yannham:
## Precedence in LALRPOP

Precedence annotations in LALRPOP have been asked for a long time. The book states
> LALRPOP doesn't have any built-in features for giving precedence to operators, mostly because I consider those to be creepy

and the author stated in some issues that these kind of annotations can be abused in bison/yacc/flex/lex, which I agree. However, I think people writing full fledged languages often found themselves needing a dozen or more precedence level for operators (random example: [the Nix language](https://gist.github.com/joepie91/c3c047f3406aea9ec65eebce2ffd449d)). While it's always possible to encode precedence by adding intermediate rules by hand, this has the following drawbacks:

- it's **error-prone**: I found myself mislabeling some level (swapping `InfixExpr6` and `InfixExpr5` for example). These bugs are not so obvious to find, because depending on what you swap, it can have a limited impact on your grammar, while still introducing bugs.
- it's **annoying to extend**: if you have to add a new operator in the middle of the precedence table, you may have to rename/renumber a lot of levels.
- it's **repetitive**: it is a very mechanical task, which immediately makes you think "this should be possible to do automatically".
- it's **verbose**: it obfuscates the actual structure of the grammar.

## The approach

I propose to follow the same approach as discussed in #67 (and the corresponding aborted PR #78): enable annotations on alternatives of a non-terminal to give precedence levels. During the normalization phase,  rules where at least one alternative has a precedence annotation are expanded in the following way:
 1. a new rule is created for each level of precedence
 2. the alternatives are partitioned by precedence level, are put in the corresponding rule, and all symbols that recursively referred to the original rule are substituted for their corresponding new level-specific rule, or for the previous level, depending on associativity annotations:
    - without annotations, all occurrences are replaced by an occurrence of the current level
    - left-associativity means that only the first occurrence is replaced by the current level, and all the following by an occurrence of the previous level
    - dually, right-associativity means that only the last occurrence is replaced by the current level
    - non-associativity means that all occurrences are replaced by the previous level: the rule cannot be iterated
 3. each level of precedence is added to the next one

This transformation is done before macro expansion: it must really be thought of as syntactic sugar.

The idea is basically to automate what people are doing manually today. One advantage is that since it amounts to just symbolic rewriting of the grammar, it doesn't depend on the underlying parser generator algorithm. The backend can be changed without breakage.

## Example

```
Expr: Expr = {
    #[precedence(level="1")]
    <Number> => ...
    #[precedence(level="1")]
    "-" <Expr> => ...

    #[precedence(level="2")]
    #[assoc(side="left")]
    <left: Expr> "*" <right: Expr> => ...,
    #[precedence(level="2")]
    #[assoc(side="left")]
    <left: Expr> "/" <right: Expr> => ..., 

    #[precedence(level="3")]
    #[assoc(side="left")]
    <left: Expr> "+" <right: Expr> => ...
    #[precedence(level="3")]
    #[assoc(side="left")]
    <left: Expr> "-" <right: Expr> => ...,
};
```
(if this feels too verbose, see the last section).

This gets expanded to:

```
Expr1: Expr = {
  <Number> => ...
  "-" <Expr1> => 
}

Expr2: Expr = {
  <left: Expr2> "*" <right: Expr1> => ...
  <left: Expr2> "/" <right: Expr1> => ...
  Expr1,
}

Expr: Expr = {
  <left: Expr> "+" <right: Expr2> => ...
  <left: Expr> "-" <right: Expr2> => ...
  Expr2,
}
```

## Prevalidation

This PR adds the following checks in the `prevalidation` module:
 - if there's at least one alternative with precedence annotations in a rule, then all the alternatives must be annotated
 - the precedence levels starts at one and are consecutive. Could be relaxed
 - syntax (precedence annotations have a `level` argument that is an integer, associativity annotations have a side argument that is etc.)
 - the first level cannot have associativity annotations. Could be relaxed if we remove the default associativity (see below).

## Limits and improvements

- I don't know if the default associativity makes sense (it's different from non assoc, it's rather "associative everywhere"). It's more consistent for unary operators and constants, for which the default, right or left associativity have the same behavior, but arbitrarily choosing left or right sounds wrong. On the other hand, it doesn't make much sense for multi-ary operators. It also has currently no proper corresponding annotation.
- this PR uses 2 different annotations for precedence and associativity. The reason is there's already an `Annotation` in LALRPOP syntax that was easy to reuse, but which only accepts one argument. Probably "#[precedence(level=1,assoc=left)]" would be a better syntax.
-We could avoid having to re-annotate each alternative, taking the convention that if an alternative is not annotated, it inherits the annotation of the last explicit annotation. Note that in that case, the default associativity is not recoverable though: once we started annotating associativity, there is no way to return to the default behavior. But we can still drop the default behavior, or give it a proper annotation.

With the two last choices, the calculator example becomes:

```
Expr: Expr = {
    #[precedence(level="1")]
    <Number> => ...
    "-" <Expr> => ...

    #[precedence(level="2", assoc="left")]
    <left: Expr> "*" <right: Expr> => ...,
    <left: Expr> "/" <right: Expr> => ..., 

    #[precedence(level="3", assoc="left")]
    <left: Expr> "+" <right: Expr> => ...
    <left: Expr> "-" <right: Expr> => ...,
};
```


lalrpop/src/normalize/precedence/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+        .partition(|input| match input {
+            GrammarItem::Nonterminal(d) => has_prec_annot(d),
+            _ => false
+        });
+
+    let nonterms_prec: NormResult<Vec<Vec<GrammarItem>>> = non_terms_prec.into_iter().map(|item| match item {
+        GrammarItem::Nonterminal(d) => expand_nonterm(d),
+        _ => panic!()
+    }).collect();
+    let mut result: Vec<GrammarItem> = nonterms_prec?.into_iter().flatten().collect();
```
Marwes, line 143: This will change the order of the items which could be confusing. Is there a reason for this or could it be preserved?

lalrpop/src/normalize/precedence/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+    )
+    .unwrap_or(false)
+}
+
+/// Expand a rule with precedence annotations. As it implies to generate new rules, return a vector
+/// of grammar items.
+fn expand_nonterm(mut nonterm: NonterminalData) -> NormResult<Vec<GrammarItem>> {
+    let alt_with_ann = Vec::with_capacity(nonterm.alternatives.len());
+
+    let (mut lvls, alts_with_ann) = mem::replace(&mut nonterm.alternatives, vec![]).into_iter().fold(
```
Marwes, line 167:
```suggestion
    let (mut lvls, alts_with_ann) = nonterm.alternatives.drain(..).fold(
```

lalrpop/src/normalize/precedence/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+                format!("{}", nonterm.name)
+            }
+            else {
+                format!("{}{}", nonterm.name, lvl)
+            }
+        ));
+
+        let nonterm_prev = if lvl > 1 {
+                Some(SymbolKind::Nonterminal(NonterminalString(Atom::from(
+                format!("{}{}", nonterm.name, lvl - 1)))))
```
Marwes, line 213: I might be wrong, but the levels does not need to be consecutive, right? So this would need to look to the previous level in the `lvls` Vec

Marwes: Looks like this is validated earlier at least, but it seems pretty arbitrary to limit the levels to be consecutive?

yannham: Yeah, they don't need to. My initial idea was that maybe if there's a hole, this can be a copy/paste error from the user. On the other hand, it's probably less ergonomic, as if one wants to delete one level, he has to renumber everything from there. I can remove this restriction if you wish.

lalrpop/src/normalize/precedence/mod.rs
yannham: You're right, I think it's just because I started from the code for macro expansion, but macro are eliminated, so it doesn't matter there, as opposed as here. I'll make the transformation order preserving.

yannham: Fixed in https://github.com/lalrpop/lalrpop/pull/555/commits/da2f17275ec96c2b229884bf767baf5eabf1f5b6

lalrpop/src/normalize/precedence/mod.rs
yannham: Removed the requirement for consecutive levels from `1` to `n` in a374f6d and 4c41182.

Marwes (approved):

Marwes:
Excellent! I will try to release this soon but I need to get CI up and working again first.

yannham:
Great, thanks for reviewing !

nixpulvis:
I'm just seeing this now and starting to open it up for review. I'm excited to see precedence and associativity being added!

The first thing I'm noticing is that I'm still getting `Ambiguous grammar detected` even for the grammar used in this PR's tests. I'm placing my grammar in a separate file and using `lalrpop::process_root()` as I normally do. Perhaps I'm doing something silly wrong.

I'll try to give a more complete review as a follow-up to this question. My initial thoughts are:

- `"#[precedence(level=1,assoc=left)]"` could be supported by extending `Annotation` to handle multiple arguments generally
- https://github.com/lalrpop/lalrpop/blame/master/doc/src/tutorial/004_full_expressions.md#L35-L71 should be updated
- I'm slightly concerned about how this interacts with macros since it expands first, but I'll have to play with that more once I get things building as expected

nixpulvis:
OK, I actually read over the example grammar and figured out my issue pretty quickly. I was missing the needed `#[assoc(side="left")]` annotations. Sorry for that noise.

PR #557, opened 2020-11-10
Allow string literals in patterns

### LINKED ISSUE #556: Matching tokens with &str field?
In my lexer I have this token type:

```
enum Token<'input> {
    Id(&'input str),
    ...
}
```

In my grammar I'm trying to do something like

```
use lexer::{Token, LexerError};

grammar<'input>(input: &'input str);

extern {
    type Location = usize;
    type Error = LexerError;

    enum Token {
        "foo" => Token::Id("foo"),
        "bar" => Token::Id("bar"),
        ...
    }
}
```

Normally I'd have different `Token` variants for these identifiers, but this lexer is used by multiple parsers, each have different set of "special" identifiers.

Is this currently possible? If not, would it be possible to extend lalrpop to do this? I'm happy to look into this is someone can advise.

osa1:
Fixes #556


lalrpop-test/src/util/tok.rs
```diff
@@ -13,10 +13,21 @@ pub enum Tok {
     Times,
     Div,
     Comma,
+    Open(Delim),
+    Close(Delim),
+    String(&'static str),
```
Marwes, line 18: This could just be a `String`, no? It may just be for testing but I'd still prefer not to leak the `String`.

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -427,6 +430,13 @@ StringLiteral: Atom =
         Ok(Atom::from(text))
     };
 
+StringConstant: String =
```
Marwes, line 433: This looks mostly identical to `StringLiteral` above, could you merge the common code into one nonterminal that both calls?

Marwes (changes requested):

lalrpop-test/src/util/tok.rs
osa1: You can't pattern match on `String`, can you? It needs to be `&str`. For example:

```rust
enum Tok {
    String(String),
    Str(&'static str),
}

fn main() {
    let t1 = Tok::String("x".to_owned());
    let t2 = Tok::Str("x");
    
    // if let Tok::String("x") = t1 {}
    if let Tok::Str("x") = t2 {}
}
```

If you enable the commented-out code this doesn't compile.

I still tried to make this a `String` and it indeed failed with errors like

```
error[E0308]: mismatched types
    --> /home/omer/rust/lalrpop/target/debug/build/lalrpop-test-f7207b3c7f924ac7/out/nested.rs:1464:33
     |
1461 |                 match *__token {
     |                       -------- this expression has type `util::tok::Tok`
...
1464 |                     Tok::String("hello") if true => Some(2),
     |                                 ^^^^^^^ expected struct `std::string::String`, found `&str`
```

osa1: We could also add a `'input` lifetime to the `Tok` type, but I thought that would be unnecessarily complex and would require refactoring a lot of existing tests.

Marwes: Right, I guess it should have the `'input` lifetime actually. May cause some compilation errors in other places using `Tok` though but should be fixable.

lalrpop/src/parser/lrgrammar.lalrpop
osa1: I tried this:

```diff
diff --git a/lalrpop/src/parser/lrgrammar.lalrpop b/lalrpop/src/parser/lrgrammar.lalrpop
index 51cb7e0..ab8979b 100644
--- a/lalrpop/src/parser/lrgrammar.lalrpop
+++ b/lalrpop/src/parser/lrgrammar.lalrpop
@@ -424,17 +424,19 @@ QuotedLiteral: TerminalLiteral = {
 };

 StringLiteral: Atom =
-    <lo:@L> <s:"StringLiteral"> =>? {
-        let text = tok::apply_string_escapes(s, lo + 1)
-            .map_err(|e| ParseError::User { error: e })?;
-        Ok(Atom::from(text))
+    <lo:@L> <s:String> =>? {
+        Ok(Atom::from(s))
     };

 StringConstant: String =
+    <lo:@L> <s:String> =>? {
+        Ok(s.into_owned())
+    };
+
+String: Result<Cow<str>, ParseError> =
     <lo:@L> <s:"StringLiteral"> =>? {
-        let text = tok::apply_string_escapes(s, lo + 1)
-            .map_err(|e| ParseError::User { error: e })?;
-        Ok(text.into_owned())
+        tok::apply_string_escapes(s, lo + 1)
+            .map_err(|e| ParseError::User { error: e })?
     };

 RegexLiteral: Atom =
```

but one of the tests started looping. Not sure if this is a lalrpop bug, or what.

osa1: Interesting.. after `git clean -xfd` it no longer loops, but the test `verify_lalrpop_generates_itself` fails. I don't know what the issue is..

osa1: Apparently the diff above is not right, the generated .rs does not compile.

In any case, the duplicated code is literally one statement (two lines). I think it's not too bad.

Marwes (approved):

Marwes:
travis didn't run on this PR, can you try making an irrelevant change (change a commit message or something) then push again to see if it wakes travis up (I tried disabling and re-enabling travis for the repo)

osa1:
I just pushed an empty commit.

osa1:
It doesn't seem to work. @Marwes can you test locally maybe? `cargo test` passes on my system.

osa1:
@Marwes can we merge this please? I've been using this since last week and it works great.

Marwes:
Sure, I'd like to have CI working again before making a release though.

PR #559, opened 2020-11-11
Shake dependency tree

Voultapher:
This PR drops the transitive dependency count from 70 down to 50 without loosing meaningful functionality.

cargo deps graph

before:
before

after:
after

As a consequence this dropped single threaded clean build down from 1:45 to 1:15, rough noisy measurement.

This started when I scratched my head as a downstream user of lalrpop, why was I building an expensive argument parser when really I wanted to build a parser?

---

Possible further improvements:

- Replacing/dropping ascii-term

This only seems to be good for the somewhat useful lines drawn when an error occurs. Which having in a styled term compatible utf8 way, is of questionable use anyway. Even with a medium < 500 lines lalrpop file, upon error I get 40mb of output. So the only viable strategy is to pipe it into a file. And I'd be happy with:
```
a some b
|      |
+------+
```
Attempted to replace it, but AsciiView is quite tightly coupled into other components, so I quickly gave up. It and its transitive dependencies contribute another ~7s to clean build time. If nothing else I'd love to have a simple output option that can do away with all those dependencies. Note, cargo deps is a bit overzealous here and shows the dependencies for all platforms, on win and linux only some of them are build, but winapi for one is quite expensive, and imo not needed for the core functionality of lalrpop.

- Improving the clean build time of lalrpop itself

it heavily out ways all other dependencies. In those 1:15, 40 seconds are lalrpop itself, yet with cargo bloat all I could figure out that it's a lot of functions 40+k, nothing big that stands out otherwise, and that it produces a ~100mb rlib. Not sure how to tackle that. Maybe someone more knowledgeable could take a closer look with `Ztime-passes`.

Also once https://github.com/servo/string-cache/pull/235 is released we should be able to drop serde from the transitive dependencies too.


Voultapher:
@Marwes would be great if you can take a look.

lalrpop/Cargo.toml
```diff
@@ -15,23 +15,64 @@ exclude = ["build.rs"]
...
-regex = "1"
-regex-syntax = "0.6"
-petgraph = "0.5"
-serde = "1.0"
-serde_derive = "1.0"
-string_cache = "0.8"
-term = "0.5"
-unicode-xid = "0.2"
-sha2 = "0.8"
+[dependencies.ascii-canvas]
```
Marwes, line 18: You can write these as `ascii-canvas = { version = "2.0", default_features = false }` etc. Personally I find that easier to read (less bloated).

lalrpop/Cargo.toml
```diff
@@ -15,23 +15,64 @@ exclude = ["build.rs"]
...
+
+[dependencies.bit-set]
+version = "0.5"
+default_features = false
+
+[dependencies.diff]
+version = "0.1.12"
+default_features = false
+
+[dependencies.pico-args]
```
Marwes, line 34: I think pico-args could be an optional dependency as it isn't needed when using lalrpop as a library.

Voultapher: Had to move the CLI binary into a separate workspace, not sure how that affects downstream users. I suggest we publish it separately for those that want to use it, which I guess is the minority as most people probably only care about the library part. Even though now with pico-args the compile time impact is minimal.

Marwes: You should not need to do that. Make the dependency optional, use `required-features` for the bin target and then add `pico-args` to the default feature (to reduce breakage).

Marwes: Why would you need to move the binary out? Why not just make the dependency optional and make it a `required-features` for that binary?

Voultapher: feature + optional, tried that, didn't work. It requires more, some way of cfg-ing the dependency out to only show up for binary builds. And I couldn't find a way to do so. Please feel free to push to this PR yourself if you have further suggestion how this specific issue should be resolved, as I'm out of my depth.

Marwes: Yes, there is unfortunately no way to force the feature enabled on binary build but by using `required-features` anyone doing `cargo install` will get a helpful error pointing out the missing feature. Alternately `pico-args` could bemade default and a `default-features = false` will disable it easily for lib builds.

I think either of those solutions is good enough, and easier than using  separate crate.

https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries

Voultapher:
Feel free to squash commits. Is there something else?

Marwes (approved):

Voultapher:
Would be great to get this released, so I can use it in my projects.

Voultapher:
@Marwes friendly ping, is there something blocking a release of this?

PR #564, opened 2020-12-08
Allow generating no_std-compatible parsers

### LINKED ISSUE #397: Feature request: optional crate feature to generate no_std parser
I'm writing a language parser and interpreter for a `no_std` environment and I have chosen to use LALRPOP as the parser generator.  While I have made sure my grammar definition and interpreter both only use types from `core` and `alloc`, the parser generated by LALRPOP includes standard library elements (`std::vec::Vec`, `std::option::Option`, etc.)

Would it be possible to add a feature to LALRPOP where the generated parser does not use types from the standard library, thus making the parser usable in `no_std` crates?

coolreader18:
Fixes #397.

I went with an option rather than only using `core`/`alloc` because a) `alloc` still requires `extern crate alloc;` in the root of the crate, and b) `core` isn't available in Rust 2015, which `calculator`, `whitespace` etc are.

Oh, also, I turned off rustfmt since it was reformatting everything; I'm not sure what the formatting guidelines are for lalrpop, or if rustfmt just hasn't been run in a while.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,12 +407,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "serde"
-version = "1.0.117"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b88fa983de7720629c9387e9f517353ed404164b1e482c970a90c1a4aaf7dc1a"
-
-[[package]]
 name = "siphasher"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


Marwes (approved):
LGTM. 

> a) alloc still requires extern crate alloc; in the root of the crate.

Is `alloc` special in that way? Otherwise it could be declared in the generated parser's module instead.

> b) I think it core isn't available in Rust 2015, which calculator, whitespace etc are.

I think it would be fair to assume that the 2018 edition is use now (if those examples are just updated). 

So I think that perhaps the config option could be skipped if those issues are fixed?

coolreader18:
Yep, you're right! I can put `extern crate {core,alloc};` at the top of the parser module and then the options aren't necessary.

mental32 (approved):
(This would be an awesome win for the embedded ecosystem)

Marwes:
Thanks! Released as 0.19.4

PR #578, opened 2021-01-05
Add support for pub(in path).

### LINKED ISSUE #567: pub(super) and pub(in path)
#293 added support for declaring grammar items with `pub(crate)`, and is written to allow paths other than `crate`, however in practice it doesn't work with anything other than `crate`.

- `pub(super)` doesn't work, because the actual parser is declared inside of an inner module, `__parse__{}`. This gives a re-export error if you use it, like this:
  
  ```
  518 | pub(super) use self::__parse__Rule::RuleParser;
      |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ re-export of private `RuleParser

  255 |     pub(super) struct RuleParser {
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ```

  In order for pub(super) to work properly, the inner module likely has to use a more relaxed visibility declaration than the outer module.

- `pub(crate::some_path)` doesn't work, because that's not actually the correct format for this kind of visibility restriction, as the error message helpfully points out:

  ```
  255 |     pub(crate::day19) struct RuleParser {
      |         ^^^^^^^^^^^^ help: make this visible only to module `crate::day19` with `in`: `in crate::day19`
      |
      = help: some possible visibility restrictions are:
              `pub(crate)`: visible only on the current crate
              `pub(super)`: visible only in the current module's parent
              `pub(in path::to::module)`: visible only on the specified path
  ```

  Fixing this latter problem seems like a pretty simple change; Update the definition of the [Visibility](https://github.com/lalrpop/lalrpop/blob/f37e588ae19b55b3a61c68d5744dcbfac8423891/lalrpop/src/parser/lrgrammar.lalrpop#L106) rule in the grammar to something like this:
  
  ```rust
  Visibility: Visibility = {
      "pub" "(" <p:Path> ")" => Visibility::PubPath(p),
      "pub" "(" "in" <p:Path> ")" => Visibility::PubInPath(p),
      "pub" => Visibility::Pub,
      () => Visibility::Priv,
  };
  ```

  And then making a simple change to the display implementation:

  ```rust
  impl Display for Visibility {
      fn fmt(&self, fmt: &mut Formatter) -> Result<(), Error> {
          match *self {
              Visibility::PubPath(ref path) => write!(fmt, "pub({}) ", path),
              Visibility::PubInPath(ref path) => write!(fmt, "pub(in {}) ", path),
              Visibility::Pub => write!(fmt, "pub "),
              Visibility::Priv => Ok(()),
          }
      }
  }
  ```
If I have a chance, I'll try to send a PR for the `pub(in path)` change in January.

I'm not sure what's the best way to handle `pub(super)` visibility, because that would require more intimate knowledge of how modules are generated than I have, and because it requires the generate to operate with more external knowledge of how visibility works in Rust. Note that since `pub(in path)` is allowed to be relative to `crate` `super` or `self`, actually-proper handling of `pub(in path)` also requires the generate to be aware of how `self` and `super` change as you go into sub-modules.

Note that if the `__parse__{}` generated module is itself private, there may be a pretty simple way to handling thisL when descending into the `__parse__{}` generated module, promote `pub(in path)` and `pub(super)` paths to `pub` or `pub(crate)`, and then just set the *correct* visibility when re-exporting from the parent.

zstewar1:
This should partially fix #567 by allowing support for `pub(in path)`. For practical purposes, supporting `pub(in path)` covers the use-case of `pub(super)` as well, since you can do the equivalent of `pub(super)` with `pub(in crate::path::to::super)`, which sidesteps the issue with `pub(super)` described in #567.

This *might* be a breaking change, because I had to add `in` as a new token to the tokenizer. I'm not sure if existing grammars using the `in` token would have worked previously, since `in` should already be a keyword on the Rust side, but its possible there were places where it could have been used in a grammar previously that now it can't. I don't know enough about LALRPOP's internals to know if this is a problem or not; e.g. IDK if LALRPOP will still accept `in` as the new token in places it would have previously.


Marwes (approved):

Marwes:
Needs a rebase, otherwise LGTM

cr1901:
@zstewar1 I ran headfirst into this issue today. Is it possible you could rebase so I (and others!) could use this feature from a git dep in Cargo.toml?

zstewar1:
Rebased to current head. Sorry for the delay, I forgot I had done this and didn't get a notification from GitHub for some reason.

PR #581, opened 2021-01-18
Require bit-set 0.5.2 or greater

brson:
0.5.1 seems to not build on Rust stable,
via its transitive bit-vec 0.5.1 dependency.

I upgraded lalrpop recently and bit-set resolved to 0.5.1 for whatever reason, and the build was broken. Updating bit-set to 0.5.2 fixed the problem in my project. This patch should prevent others from running into the same problem.


Marwes:
Thanks!

PR #591, opened 2021-02-25
Spanned precedence annotations

yannham:
#555 introduced precedence annotations. However, one practical quirk is that one needs to re-specify the precedence on each alternative. Usually, alternatives are grouped by precedence in a grammar, and we could use one annotation for a whole block.

To illustrate the point, here is the calc example of the LALRPOP book with current precedence annotations:

 ```
Expr: i32 = {
    #[precedence(lvl="0")]
    Num,
    #[precedence(lvl="0")]
    "(" <Expr> ")",

    #[precedence(lvl="1")] #[assoc(side="left")]
    <l:Expr> "*" <r:Expr> => l * r,
    #[precedence(lvl="1")] #[assoc(side="left")]
    <l:Expr> "/" <r:Expr> => l / r,

    #[precedence(lvl="2")] #[assoc(side="left")]
    <l:Expr> "+" <r:Expr> => l + r,
    #[precedence(lvl="2")] #[assoc(side="left")]
    <l:Expr> "-" <r:Expr> => l - r,
};
```

Which is a bit verbose. Here is how we can write the same example after this PR:
```
Expr: i32 = {
    #[precedence(lvl="0")]
    Num,
    "(" <Expr> ")",

    #[precedence(lvl="1")] #[assoc(side="left")]
    <l:Expr> "*" <r:Expr> => l * r,
    <l:Expr> "/" <r:Expr> => l / r,

    #[precedence(lvl="2")] #[assoc(side="left")]
    <l:Expr> "+" <r:Expr> => l + r,
    <l:Expr> "-" <r:Expr> => l - r,
};
```

## Rules
- if any alternative has a precedence annotation, the first alternative MUST have a precedence annotation (otherwise, we can't know which level it is)
- a precedence annotation affects all subsequent alternatives, until the next precedence annotation, or the end of the rule
- an associativity annotation affects all subsequent alternatives, until the next associativity annotation OR the next precedence annotation. The rationale is that somehow precedence annotations define blocks in the grammar, and that associativity should be re-specified at the end of such block (or be set to default if not precised), but this design can of course be discussed.

## Default associativity
The default associativity is full associativity. This makes sense because when one write:
```
Expr: ... {
  #[precedence(lvl="0")]
  <Expr> + <Expr> => ...
}
```
Full associativity entails that the alternative has the same semantics as if there were no associativity annotation at all: all `Expr` occurrence refers to the current level.

With this PR, associativity can now be inherited from the last annotation, so it may be useful to have an explicit way of going back to the default associativity. This PR adds a `side="all"` annotation to indicate the default, full associativity.

## Backward compatibility
This change is backward compatible. Since currently all alternatives need to be annotated, our rule on resetting associativity on new precedence annotations enforces that associativity is reset to default (unless specified explicitly) each time, having thus the same semantics as before.

### DIFF
--- a/lalrpop/src/normalize/precedence/mod.rs
+++ b/lalrpop/src/normalize/precedence/mod.rs
@@ -128,7 +128,7 @@ impl FromStr for Assoc {
     }
 }
 
-/// Perform precedence expansion. Rewrite rules where at least one alternative have a precedence
+/// Perform precedence expansion. Rewrite rules where at least one alternative has a precedence
 /// annotation, and generate derived rules for each level of precedence.
 pub fn expand_precedence(input: Grammar) -> NormResult<Grammar> {
     let input = resolve::resolve(input)?;
@@ -328,7 +328,7 @@ fn replace_symbols<'a>(
     }
 }
 
-/// Perform substitution of on an non-terminal in a symbol.
+/// Perform substitution of a non-terminal in a symbol.
 fn replace_symbol<'a>(
     symbol: &mut Symbol,
     target: &NonterminalString,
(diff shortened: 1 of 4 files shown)


Marwes (approved):

Marwes:
Looks great! Thanks!

Issue #546, opened 2020-08-19
Tutorial Broken Links


AlexDikelsky:
In this page http://lalrpop.github.io/lalrpop/tutorial/003_type_inference.html, the links to the code for calculator1 and calculator2 are broken. 

Also, in this page http://lalrpop.github.io/lalrpop/tutorial/004_full_expressions.html, the link to calculator3 is broken.

EDIT: I believe these files are located here right now https://github.com/lalrpop/lalrpop/tree/master/doc/calculator/src

Teln0:
lalrpop.github.io doesn't seem to exist anymore ...

Pluriscient:
> 
> 
> lalrpop.github.io doesn't seem to exist anymore ...

https://lalrpop.github.io/lalrpop/ does so it's more a matter of redirecting to that which is not happening

jakevossen5:
Is this an issue that can be fixed in the `mdbook`, or is there something messed up in the GitHub pages config?

maxsnew:
More broken links here: https://lalrpop.github.io/lalrpop/tutorial/index.html

The link to sub-section 001 is https://lalrpop.github.io/lalrpop/tutorial/tutorial/001_adding_lalrpop.html and should be https://lalrpop.github.io/lalrpop/tutorial/001_adding_lalrpop.html etc

bryantbiggs:
should be resolved with #724 now

nikomatsakis:
Seems to work, indeed.

PR #605, opened 2021-05-05
Fix book links (issue #546)

### LINKED ISSUE #598: Improving the book
It's really hard to understand to what's going on in some LaLRPop code, and tutorial is still incomplete and doesn't cover that. Please, finish this documentation first. Because (for example) location tracking and everything like that is important features but we must for now try to guess how to use them. Thanks.

pentla:
Fix [books](http://lalrpop.github.io/lalrpop/index.html) link.

https://github.com/lalrpop/lalrpop/issues/546
This issue caused by relative path in markdown. Then I changed relative path to absolute path.

And in [tutorial first page](http://lalrpop.github.io/lalrpop/tutorial/index.html), path is changed main.rs → lib.rs.

I would appreciate it is checked.

Fixes #598 #546


nikomatsakis:
This seems good to me.

PR #607, opened 2021-05-26
Upgrade ascii-canvas to 3.0

coolreader18:
This also upgrades term to 0.7, which switches the unmaintained `dirs` crate out for `dirs-next`.


Marwes:
Thanks!

coolreader18:
Do you think you'd be able to put out a patch release with this?

PR #627, opened 2021-09-08
fixed 404 links

Volker-Weissmann:
Consider using https://crates.io/crates/mdbook-linkcheck and maybe https://github.com/budziq/rust-skeptic


nikomatsakis:
Thanks @Volker-Weissmann -- looks like this was fixed already.

PR #644, opened 2022-01-19
Fix broken tutorial links

l4l:


Marwes:
These links have been changed back an forth due to people claiming they are broken so I rather not make any further changes until someone understands why they keep changing.

l4l:
It feels these should be markdown urls actually. Also there's a [linkcheck plugin](https://github.com/Michael-F-Bryan/mdbook-linkcheck) that may help with the issue.

l4l:
The last commit introduces [CI job](https://github.com/l4l/lalrpop/runs/4993053860?check_suite_focus=true) for link check and fixes the rest of broken ones.

l4l:
Anything else I could do towards PR being merged?

PR #654, opened 2022-03-01
Doc chapter on how to lex string literals

pnkfelix:


doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,346 @@
...
+
+Our calculator example operated on numbers and arithmetic operators.
+There is no overlap between the characters for numeric digits (`0`, `1`, ...)
+the characters representing operators (`+`, `-`, ...) and parentheses
+(`(`, `)`), so it was easy to embed those tokens directly in the grammar,
+as we saw in the earlier sections.
+
+However, some languages do not have such easy separations at the lexical level.
+
+Consider parsing a language with string literals. We will deliberate choose
```
pnkfelix, line 11:
```suggestion
Consider parsing a language with string literals. We will deliberately choose
```

doc/nobol/src/nobol7.lalrpop
```diff
@@ -0,0 +1,10 @@
+use super::{Var, Lit, Stmt};
+
+grammar;
+
+pub Var: Var = <r"[a-z]"> => <>.chars().next().unwrap().into();
+
+pub Lit: Lit = <l:r#""(\\\\|\\"|[^"])*""#> =>
```
shesek, line 7: Shouldn't this be `"(\\\\|\\"|[^"\\])*"` (exclude `\` too)?

doc/nobol/src/nobol7.lalrpop
```diff
@@ -0,0 +1,10 @@
...
+    l[1..l.len()-1].replace("\\\"", "\"").replace("\\\\", "\\").into();
```
shesek, line 8: This would consider `\\"` as an escaped `"` rather than an as escaped `\`.

`\\"` wouldn't be accepted by the regex to begin with, but the problem is more apparent if more escape sequences were added. For example if `\n` was also supported, `\\n` would be parsed as `\<LF>`.

Maybe this could instead suggest an approach similar to that implemented in [`tok::apply_string_escapes()`](https://github.com/lalrpop/lalrpop/blob/d45124db36919e8a70a04581745b156fdeafbe92/lalrpop/src/tok/mod.rs#L749)?

shesek:
Thank you for documenting this, very useful! Was pleasantly surprised to find this, written just days before my search for a solution :)

doc/nobol/src/nobol7.lalrpop
pnkfelix: Probably! Let me look into that, thanks!

doc/nobol/src/nobol7.lalrpop
pnkfelix: Excellent point! You're right, I was handling this in a quick and dirty fashion, and while it "works" for this simple case (to `\\"` being illegal, as you noted), it does not scale well.

windrunner414: I'm using this `"([^"\\\r\n]|\\[^\r\n])*"`
It accept something like `\n`, `\t`

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,346 @@
...
+If we wanted to generalize it to be able to hold arbitrary characters, we would
+need some way to denote the delimiter character `"` in ths string content.
+
+The usual way to do this is via an escape sequence: `\"`, which is understood by
+the lexical analyzer as *not* ending the string content.
+
+We can generalize the regular expression in our new `Lit` rule to handle this:
+
+```lalrpop
+pub Lit: Lit = <l:r#""(\\\\|\\"|[^"])*""#> => l[1..l.len()-1].into();
```
pnkfelix, line 316:
```suggestion
pub Lit: Lit = <l:r#""(\\\\|\\"|[^"\\])*""#> => l[1..l.len()-1].into();
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,346 @@
...
+  left: `Ok(Stmt(Var('z'), Lit("\\\"\\\\")))`,
+ right: `Ok(Stmt(Var('z'), Lit("\"\\")))`', doc/nobol/src/main.rs:91:5
+```
+
+This can be readily addressed by adding some code to post-process the token to remove the
+backslashes:
+
+```lalrpop
+pub Lit: Lit = <l:r#""(\\\\|\\"|[^"])*""#> =>
+    l[1..l.len()-1].replace("\\\"", "\"").replace("\\\\", "\\").into();
```
pnkfelix, line 345:
```suggestion
pub Lit: Lit = <l:r#""(\\\\|\\"|[^"\\])*""#> => Lit(apply_string_escapes(&l[1..l.len()-1]).into());
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,346 @@
...
+```
```
pnkfelix, line 346:
```suggestion
```

where `apply_string_escape` is a helper routine that searches for occurrences of `\\` and does the appropriate substitution within the string content.
```

Comment (pnkfelix) [inline doc/src/lexer_tutorial/002_raw_delimited_content.md:346]:
````suggestion
```

where `apply_string_escapes` is a helper routine that searches for backslashes in the content and performs the corresponding replacement with the character denoted by the escape sequence.
````

**Comment by nikomatsakis** [comment]
@pnkfelix if you rebase this, we'll merge it :)

File: doc/nobol/src/main.rs [hunk 7]
```diff
@@ -0,0 +1,148 @@
...
+lalrpop_mod!(pub nobol1); // syntesized by LALRPOP
+
+#[derive(PartialEq, Debug)]
+pub struct Var(char);
+
+#[derive(PartialEq, Debug)]
+pub struct Lit(String);
+
+#[derive(PartialEq, Debug)]
+pub struct Stmt(Var, Lit);
```
yannham, line 13: Nitpick: I think `Stmt` is quite generic. What about `Assign` ?

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
+# Lexing raw delimited content
+
+Our calculator example operated on numbers and arithmetic operators.
+There is no overlap between the characters for numeric digits (`0`, `1`, ...)
```
yannham, line 4:
```suggestion
There is no overlap between the characters for numeric digits (`0`, `1`, ...),
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+the characters representing operators (`+`, `-`, ...) and parentheses
+(`(`, `)`), so it was easy to embed those tokens directly in the grammar,
+as we saw in the earlier sections.
+
+However, some languages do not have such easy separations at the lexical level.
```
yannham, line 9:
```suggestion
However, some languages do have overlapping lexical token.
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+Consider parsing a language with string literals. We will deliberately choose
+one that is very simple: All it can do is bind variables, and the variables are
+always single characters, like this:
+
+```
+x = "a"
+y = "bc"
+```
+
+Using what we have learned so far, we might try a grammar like the below.
```
yannham, line 20:
```suggestion
Using what we have learned so far, we might try a grammar like the following one:
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+```
+
+Okay, this sounds easy to fix: the regular expression for string literal content should
+be expanded to handle characters other than `a`, `b` and `c`.
+
+So does this work?
+
+```lalrpop
+pub Lit: Lit = "\"" <r"[a-z]*"> "\"" => <>.into();
+```
```
yannham, line 100: I think this section can be cut out. The guide starts by saying that the calculator had a clear separation between lexical tokens, but this document is about grammars with some form of overlapping. However, we start the section with a grammar without overlapping, and which honestly looks really artificial. I'm not sure it adds a lot of value.

My suggestion would be to get rid of this first step and directly starts with imagining a language with strings, variables and statements, and then write the naive ambiguous version of this grammar.

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+pub Lit: Lit = {
+    "\"" <r"[x-z]"> "\"" => <>.into(),
+    "\"" <r"[a-z]*"> "\"" => <>.into(),
+};
+```
+
+But this is a fragile workaround.
+
+Specifically, this workaround is only applicable because we put artificial
+limits in this language.
```
yannham, line 198:
```suggestion
limits on this language.
```

Not sure though, I'm not a native speaker :slightly_smiling_face:

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+
+we get the following error output:
+
+```
+thread 'spaceballs' panicked at 'assertion failed: `(left == right)`
+  left: `Err(UnrecognizedToken { token: (0, Token(2, "z "), 2), expected: ["r#\"[x-z]*\"#"] })`,
+ right: `Ok(Stmt(Var('z'), Lit("x")))`', doc/nobol/src/main.rs:58:5
+```
+
+Our attempt to generalize what string contents can be has caused problems for
```
yannham, line 239:
```suggestion
Our attempt to generalize what strings can contain has caused problems for
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+
+```lalrpop
+pub Lit: Lit = "\"" <r"[a-c]*"> "\"" => <>.into();
+```
+
+The heart of our problem is that we have implicitly specified distinct tokens
+for the string delimiter (`"\""`) versus the string content (in this case,
+`r"[a-c]*"`).
+
+The problem is that our intuition is that we only want to tokenize string
```
yannham, line 256:
```suggestion
Our intuition is that we only want to tokenize string
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+content when we are in the process of reading a string. In other words, we only
+want to apply the `r"[a-c]*"` rule immediately after reading a `"\""`. But the
+generated lexer does not infer this from our rules; it just blindly pattern
+matches for something matching the string content regular expression *anywhere*
+in the input.
```
yannham, line 261:
```suggestion
generated lexer does not infer this from our rules; it just blindly looks for
something matching the string content regular expression *anywhere*
in the input.
```

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+
+You could resolve this with a custom lexer (treated in the next section).
+
+But the simplest thing to do in this scenario is to read the string delimiters and the string content as a *single token*, like so:
```
yannham, line 265:
```suggestion
But the simplest thing to do is to read the string delimiters and the string content as a *single token*, like so:
```

doc/src/lexer_tutorial/index.md
```diff
@@ -3,6 +3,7 @@
 This part is about controlling the inner workings of LALRPOP's built-in lexer generator and using your own hand written parser.
 
 - [LALRPOP's lexer generator](001_lexer_gen.md)
-- [Writing a custom lexer](002_writing_custom_lexer.md)
-- [Using tokens with references](003_token_references.md)
+- [Lexing raw delimited content](002_raw_delimited_content.md)
+- [Writing a custom lexer](003_writing_custom_lexer.md)
+- [Using tokens with references](004_token_references.md)
 - [Using an external library](004_external_lib.md)
```
yannham, line 9: Shouldn't this chapter be renumbered as well?

yannham (commented):
Looks good to me overall :+1: I just left a bunch of suggestion in the hope of making the text a bit more straight to the point. I think there is one chapter that still needs to be renumbered (linking) though.

pnkfelix:
Thanks for the review! I've been neck deep in other work for weeks, but I just got out of that muck and I think I can address your feedback now!

doc/src/lexer_tutorial/002_raw_delimited_content.md
pnkfelix: Its been over a year, but from what I remember about the way that my brain works, it was easiest for *me* to get the initial grammar working by starting from the artificially limited case, get all those pieces to work, then figure out how to generalize it.

Sure, the way I restricted it was absurdly limited. But there is plenty of precedent, especially in PL work, of people forcing programmers to follow seemingly absurd syntactic restrictions in order to satisfy the limits of the lexer/parser implementation.

Anyway, my point is more that I thought it was worthwhile to spell out "here's working code, but it doesn't readily generalize." And it *is* worthwhile having those artificial limits when it makes the discussion clear; for example, in the section "The right way to do this", the places where it refers to `r"[a-c]*"` is entirely *unambiguous* because that range is solely used in the definition for the `Lit` production.

pnkfelix: Having said all that, let me take a shot at the re-write you suggest. Maybe it will come out cleaner.

doc/src/lexer_tutorial/index.md
pnkfelix: Hunh, how did that get overlooked... will fix!

doc/nobol/src/main.rs
pnkfelix: ugh, you may be right, though I fear all the ways that name leaks into the rest of the source snippets that are in the text... I'll look into it.

pnkfelix: I'll go with `Eql`; we're not implementing an interpreter here and that way all the non-terminal names have three characters each.

yannham:
Sure! Most of them are suggestions, which can be wrong: feel free to protest when you think a suggestion betrays the original text. In general I'm trying to make the text easier to read and cut the fat (because it seems our default inclination as technical human writers is to include non-essential things or be verbose/convoluted), but sometimes it might be too much.

doc/src/lexer_tutorial/002_raw_delimited_content.md
pnkfelix: Addressed in fb568a3be286438944569aae7d25b33e6e0fa17b

doc/src/lexer_tutorial/002_raw_delimited_content.md
```diff
@@ -0,0 +1,347 @@
...
+There is no overlap between the characters for numeric digits (`0`, `1`, ...)
+the characters representing operators (`+`, `-`, ...) and parentheses
+(`(`, `)`), so it was easy to embed those tokens directly in the grammar,
+as we saw in the earlier sections.
+
+However, some languages do not have such easy separations at the lexical level.
+
+Consider parsing a language with string literals. We will deliberately choose
+one that is very simple: All it can do is bind variables, and the variables are
+always single characters, like this:
```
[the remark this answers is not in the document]
pnkfelix: Addressed in fb568a3be286438944569aae7d25b33e6e0fa17b

doc/src/lexer_tutorial/002_raw_delimited_content.md
pnkfelix: Addressed in fb568a3be286438944569aae7d25b33e6e0fa17b

doc/src/lexer_tutorial/002_raw_delimited_content.md
pnkfelix: Addressed in fb568a3be286438944569aae7d25b33e6e0fa17b

doc/src/lexer_tutorial/002_raw_delimited_content.md
pnkfelix: Addressed in fb568a3be286438944569aae7d25b33e6e0fa17b

doc/nobol/src/main.rs
pnkfelix: Addressed by b84c123085fd321d0704e0b103ae960424bf68e4

doc/src/lexer_tutorial/index.md
pnkfelix: Addressed by fe67e18b0e7ee2b685de3e67a38889ebf88fcfc9

pnkfelix:
Okay I think this is ready for re-review or outright landing now.

yannham (approved):

youknowone:
Thank you!

PR #659, opened 2022-04-14
Clarify error recovery documentation

dburgener:
* Add code example of use declaration
* Add link to calculator7 example
* Explain what ErrorRecovery struct does
* Fix typo


nikomatsakis:
Thanks!

PR #664, opened 2022-04-22
Fix leading zeros in each byte of the sha3 getting dropped

coolreader18:
Was switching from my build script building lalrpop to just verifying that the
generated file is up to date via the sha3 and noticed that the sha3 is
formatted.. very wrong.


coolreader18:
Would you be able to cut a new release with this?

PR #670, opened 2022-06-20
Fix token_to_integer to not require Copyable

Destroyerrrocket:
We can simply use a reference

### DIFF
--- a/lalrpop/src/normalize/lower/mod.rs
+++ b/lalrpop/src/normalize/lower/mod.rs
@@ -176,7 +176,7 @@ impl<'s> LowerState<'s> {
 
         read_algorithm(&grammar.annotations, &mut algorithm);
 
-        let mut all_terminals: Vec<_> = self
+        let all_terminals: Vec<_> = self
             .conversions
             .iter()
             .map(|c| c.0.clone())
@@ -186,7 +186,6 @@ impl<'s> LowerState<'s> {
                 None
             })
             .collect();
-        all_terminals.sort();
 
         let terminal_bits: Map<_, _> = all_terminals.iter().cloned().zip(0..).collect();
 
--- a/lalrpop/src/normalize/mod.rs
+++ b/lalrpop/src/normalize/mod.rs
@@ -19,7 +19,7 @@ macro_rules! return_err {
         return Err(NormError {
             message: format!($($args),+),
             span: $span
-        });
+        })
     }
 }
 
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -738,7 +738,7 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
             .emit()?;
         rust!(self.out, "{{");
 
-        rust!(self.out, "match *{p}token {{", p = self.prefix);
+        rust!(self.out, "match &*{p}token {{", p = self.prefix);
 
         for (terminal, index) in self.grammar.terminals.all.iter().zip(0..) {
             if *terminal == TerminalString::Error {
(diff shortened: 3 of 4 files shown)


lalrpop/src/normalize/lower/mod.rs
```diff
@@ -176,7 +176,7 @@ impl<'s> LowerState<'s> {
 
         read_algorithm(&grammar.annotations, &mut algorithm);
 
-        let mut all_terminals: Vec<_> = self
+        let all_terminals: Vec<_> = self
```
nikomatsakis, line 179: Why this change?

lalrpop/src/normalize/mod.rs
```diff
@@ -19,7 +19,7 @@ macro_rules! return_err {
         return Err(NormError {
             message: format!($($args),+),
             span: $span
-        });
+        })
```
nikomatsakis, line 22: Isn't there a compatibility warning about this?

nikomatsakis (commented):
This seems fine, but I'm curious why the call to `sort` was removed?

lalrpop/src/normalize/mod.rs
youknowone: The original code makes warning. The changed one is fine.

lalrpop/src/normalize/lower/mod.rs
Destroyerrrocket: mut was not necessary

Destroyerrrocket:
> This seems fine, but I'm curious why the call to sort was removed?

This way the order of the match arms is preserved, otherwise some arms that are more general would be moved over more specific ones.

lalrpop/src/normalize/lower/mod.rs
youknowone: @Destroyerrrocket As I understand, he is asking about the removed sort below.

Destroyerrrocket: Yep, that sort caused problems while trying to match str&. Tbf, it's been a while and I don't remember all the details, but it gave me problems...

Destroyerrrocket:
In this issue I explained in detail what that sort does:
#671

dburgener:
I've been poking at this a bit, and locally rebased it onto the latest code.  A few thoughts/questions:

* It looks like the majority of the code changes here were just small cleanups that have been  handled in the interim by other PRs.  What's left is the `*__token` -> `&*__token` change that's the heart of it, I think, and the removal of the `all_terminals.sort()`
* The `*__token` -> `&*__token` change throws a warnings on modern rust because we're just dereferencing something and then immediately taking a reference on it.  I think the better change would be to drop the dereference (`*__token` becomes just `__token`, since `__token` is already a reference).
* The above seems like an obviously good change to make, but I'm a little confused by the PR framing of "not require Copyable".  I assume this means the Copy trait?  If so, I'm confused because token_to_integer doesn't seem bound to the Copy trait, and the Tok struct doesn't implement Copy (either in the code today, or the version from 2022 this PR was based on).  If that's not what it means, then I'm not sure what it means.  Either way, this PR doesn't seem to remove any trait bounds.
* In terms of the `sort()`, the issue is getting confidence that there aren't any subtle breakages.  It does seem like these would be better handled as two independent issues.  I looked at the code a bit, and haven't found any reason removing the `sort()` isn't fine, but I wouldn't go as far as to say I have confidence yet.  The code that touches this array later in compilation isn't all that big, or involved though, so gaining that confidence is probably doable, I just need to spend more time on review.
* The merge conflict here is pretty trivial.

So taking all that together, I have a couple of questions:

* @Destroyerrrocket I know it's been over a year since you submitted this PR.  Is this still something you're interested in working on?  I'm happy to take it over if not.  Just let me know.  If you are, can you start by rebasing this against the latest lalrpop?
* Also, can you elaborate on the "not require Copyable" aspect of this?  The change to match a reference makes sense, but I don't see any Copy trait changes, so I'm confused as to the goal
* Thirdly for me, I need to dig in more on that `sort()`, so I'll mark that down as a TODO when I get a chance.

Pat-Lafon:
Closing as completed in #918

Issue #195, opened 2017-03-15
have parser offer lexer feedback as to which tokens are expected


nikomatsakis:
It often happens in grammars that you might want to have a distinct set of tokens at different points in the grammar; LALRPOP's intertwined syntax also leads to that expectation (see e.g. #193). This comes up a lot in LALRPOP's own grammar, and an example from Rust would be macro bodies. You can usually work around this in various ways (e.g., LALRPOP has its own tokenizer that handles it) but it'd be interesting to have the parser inform the lexer as to its current state and what set of tokens it expects at this point. We could integrate this into our overlap analysis and thus avoid rejecting grammars like the one in #193.

This would have many of the advantages that the "scanner-less" parsing mode of PEG grammars offers (#143) without the disadvantages of PEG (poorer performance, can't statically distinguish ambiguous grammars).

nikomatsakis:
Another use case for this is contextual keywords, as described in #112.

nikomatsakis:
OK, so, I gave some more thought to this. I think it'd be fairly easy to use (e.g.) a combination of LR0 states and follow-sets to partition the input tokens into "sets that may be expected at one time" (I'm aiming for something cheaply computed and "good enough"), but it occurs to me that if you just do this naively, you will wind up parsing way more than you meant to. For example, many languages have "keywords", and you actually **want** to prevent them from being used as identifiers, **even if** that would be possible without ambiguity.

On the other hand, in **some** contexts (e.g., inside a comment) you do not. That is, I had hoped to use this to replace lexer states by having some kind of special nonterminal `Drop`:

```rust
Drop = { Whitespace, Comment };
Whitespace = r"\s*";
Comment = "/*" Commented "*/*;
Commented = {
    Comment, r"."
};
```

(Ignoring the ambiguity there between `r"."` and `r"/*"` for the moment, which could be resolved in various ways.)

This would only work if entering the "comment" meant that other tokens were being ignored by lexer. 

Seems like we need to give user more control over this.

remexre:
If you haven't read https://www-users.cs.umn.edu/~evw/pubs/vanwyk07gpce/vanwyk07gpce.pdf, it has an algorithm that works quite well for this. [Copper](https://github.com/melt-umn/copper) implements this.

(Disclaimer: I'm affiliated w/ the research group.)

elkowar:
[...]

PR #673, opened 2022-07-05
[POC] feat: Allow the parser to direct the which tokens it expects from the lexer

Marwes:
AKA lexer modes.

Quick-ish experiment to investigate if it is feasible to use lalrpop for https://github.com/influxdata/flux . Seems to work for the purpose of allowing division (`/`) and regex `/a.+/` to co-exist in a grammar. The code is in a hacky state right now though due to the neecessary addition of `LexerIterator` so there would need to be more work before this could be merged. I also don't know how well this scales to other grammars. Implementation is based on a cursory understanding of the paper mentioned here https://github.com/lalrpop/lalrpop/issues/195#issuecomment-562881667 .

Closes https://github.com/lalrpop/lalrpop/issues/195

### DIFF
--- a/lalrpop-util/src/state_machine.rs
+++ b/lalrpop-util/src/state_machine.rs
@@ -163,36 +163,41 @@ pub trait ParserAction<D: ParserDefinition>: Copy + Clone + Debug {
 
 pub trait IntoLexerIterator<A> {
     type Item;
-    type IntoIter: Iterator<Item = Self::Item>;
+    type IntoIter: LexerIterator<A, Item = Self::Item>;
 
-    fn into_iter(self) -> Self::IntoIter;
+    fn into_lex_iter(self) -> Self::IntoIter;
 }
 impl<I, A> IntoLexerIterator<A> for I
 where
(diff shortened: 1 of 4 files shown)


dburgener:
For anyone who is interested, here's a branch that cherry-picks the relevant parts of this onto the latest master: https://github.com/dburgener/lalrpop/tree/lexer_modes_update

I dropped the first two commits with formatting and warnings, since presumably those were cleaned up.  The major note is that neither this PR directly, or my update actually build successfully.  I haven't looked too deeply, but it seems as though not everything has yet been sorted out in code generation between when to use Iterator vs when to use LexerIterator (I'm getting errors around `into_lex_iter()` calls on things that don't define the LexerIterator trait, as well as various calls that require Iterator on things that don't define that trait).  I haven't had time to dig into those errors, and I'm not currently ready to commit to doing so in the future.

So maybe I'll come back to this, maybe not, but either way hopefully this status of where I'm leaving it now is helpful to either future me or someone else taking this over.

PR #676, opened 2022-07-21
Fix build warnings

youknowone:


yannham (approved):
Looks good. Should fix the scary warning about LALRPOP becoming rejected by a future version of Rust. Thanks!

PR #696, opened 2022-12-28
Stream errors to the terminal as they are computed

andersk:
I’m working on a project where LALRPOP typically takes many minutes or hours to compute all the error messages. It’s much more useful to show the messages as they’re computed, than to buffer them all and display them at the end.


nikomatsakis (approved):

youknowone (approved):
Thank you for contributing!

PR #698, opened 2023-01-06
Lex raw identifiers

### LINKED ISSUE #613: Allow raw identifiers in grammars
Currently the following code:

```rust
pub Term: Term<'arena> = {
    FunTerm,
    <expr: FunTerm> ":" <r#type: Term> => {
        Term::Ann(
            arena.alloc_term(expr),
            arena.alloc_term(r#type),
        )
    },
    ...
};
```

Fails to compile with the following error:

```
.../fathom/experiments/rust-minimal/src/grammar.lalrpop:32:28: 32:28 error: expected string literal; missing `"`?

  --- stderr
        <expr: FunTerm> ":" <r#type: Term> => {
                               ^
```

This is especially troublesome if you use raw identifiers in struct fields, which means you have no way of constructing the struct within the grammar! 😅

Kmeakin:
Lex raw identifiers (identifiers prefixed with `r#`). Fixes #613

### DIFF
--- a/lalrpop/src/tok/mod.rs
+++ b/lalrpop/src/tok/mod.rs
@@ -600,6 +600,11 @@ impl<'input> Tokenizer<'input> {
                     None => error(UnterminatedStringLiteral, idx0),
                 }
             }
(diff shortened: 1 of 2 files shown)


youknowone (approved):
Thank you for contributing!

PR #700, opened 2023-01-17
Use `&'static str` in `ParseError::UnrecognizedEOF` and `ParseError::UnrecognizedToken`

Kmeakin:
Replace `Vec<String>` with `Vec<&'static str>` in `ParseError::UnrecognizedEOF` and `ParseError::UnrecognizedToken`. Since the terminal names are all known ahead of time, there is no need to store them on the heap.


nikomatsakis:
Thanks for the PR. This is a reasonable change, but I'm also not sure how much I want to *commit* to knowing things up-front and not having dynamically generated names. Is there a strong motivation for this, or is it more like "general hygiene"?

nikomatsakis:
That makes sense, but then the types are more complicated, which also doesn't seem worth it to me yet.

PR #722, opened 2023-03-25
remove gitter, promote discussions

nikomatsakis:
I'm trying to remove Gitter from my life -- too many chat apps! Also, I can't keep up. I propose we close that room and shift over to GH discussions. I'm open to pushback on this, though, if other folks are getting value out of Gitter.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # LALRPOP
 
-[![Join the chat at https://gitter.im/lalrpop/Lobby](https://badges.gitter.im/lalrpop/Lobby.svg)](https://gitter.im/lalrpop/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-
 [![Build status](https://travis-ci.org/lalrpop/lalrpop.svg?branch=master)](https://travis-ci.org/lalrpop/lalrpop)
+
+**Questions?** Ask for help on [Github Discussions](https://github.com/lalrpop/lalrpop/discussions).
 
 LALRPOP is a Rust parser generator framework with *usability* as its
 primary goal. You should be able to write compact, DRY, readable


youknowone:
gitter was a reasonable choice years ago when chat apps maturity was not that differen. but these days, while other chat apps are going to improved, gitter didn't.
even if lalrpop users want to keep a chat platform, using another one will be a reasonable choice unless there is intend to support matrix.

README.md
```diff
@@ -1,9 +1,9 @@
 # LALRPOP
 
-[![Join the chat at https://gitter.im/lalrpop/Lobby](https://badges.gitter.im/lalrpop/Lobby.svg)](https://gitter.im/lalrpop/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-
 [![Build status](https://travis-ci.org/lalrpop/lalrpop.svg?branch=master)](https://travis-ci.org/lalrpop/lalrpop)
 
+**Questions?** Ask for help on [Github Discussions](https://github.com/lalrpop/lalrpop/discussions).
```
bryantbiggs, line 5: FYI - theres a few more gitter references 
- Line 39
- Line 56

linkdd:
Github Discussions feel more like a forum than a chat room (and phpBB remains the superior forum software IMHO).

With Github Discussions, I'd get notifications via mail, and not on everything, and I'll probably remove them without even reading them.

A chatroom, I can see when there are people talking (asking questions or just chatting) quite easily, and it's also easier to backlog to see what was going on.

But I'm an IRC guy, so I'm biased 😅

Schyrsivochter:
Gitter is just another Matrix server these days.

Personally, I’ve been hanging around on Gitter those past couple years helping people with LALRPOP questions. Forums may be better for keeping track of existing answers, but a chatroom is more direct communication and IMO less scary for newbies. Both have advantages and disadvantages, I guess, and I’m not partial to either of them at this point.

nikomatsakis:
OK. I had a brief chat with @youknowone and we agreed that it's nice to have a chat room -- I know I don't personally have time to monitor it,  but I'm happy as long as people are able to find answers there. 

@youknowone said that it should be possible to connect the existing gitter room into a matrix client. I do run Element so if somebody can update the instructions appropriately to suggest using Matrix, that'd be great.

yannham:
I'm late to the fight, but FWIW I also think that the gitter provides value to newcomers (special thanks to @Schyrsivochter who has answered countless questions there!)

youknowone:
Element is looking a lot better than gitter app. But I still can say Discord client is better than Element.

In Element app, recognizing new message is harder because I only can choose to fold the whole DMs/channels/invitations rather than watching only new message sources.

PR #615, opened 2021-07-10
Fix the tutorial links

lf-:
All the page links on this page were broken: http://lalrpop.github.io/lalrpop/tutorial/index.html

It appears that this is because of a duplicate tutorial/ component, so I deleted it and it seems right now.

PR #724, opened 2023-03-25
chore: Remove TravisCI references with switch to GitHub actions; update actions versions to remove warnings

bryantbiggs:
- Remove TravisCI references with switch to GitHub actions
- Update actions versions to remove warnings
- Fix mdbook deploy by adding mdbook-linkcheck to workflow

The mdbook stuff could probably be cleaned up and simplified further, but this is a start to get the main CI back green and happy again


bryantbiggs:
Tested and validated here https://github.com/clowdhaus/lalrpop/actions

.github/workflows/test.yaml
```diff
@@ -12,13 +19,19 @@ jobs:
...
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
-          override: true
           components: rustfmt
       - name: Run tests
-        run: ./ci.sh
```
youknowone, line 23: Can we keep those script as separated files? They are helpful when locally testing without CI.

linkdd: NB: I often use https://github.com/nektos/act to test my Github Action pipelines.

bryantbiggs:
@youknowone one sec on this - checking the generated mdbook output in my repro to ensure content is generated correctly for github pages. right now it doesn't look like it https://github.com/clowdhaus/lalrpop/tree/gh-pages

youknowone:
#615 is also related to mdbook deployment

bryantbiggs:
ok that fixed it now https://clowdhaus.github.io/lalrpop/

bryantbiggs:
it looks like in #615 they are removing the linkcheck which I think it would be good to keep. Also, this change here is using the mdbook github action which just uses the [latest version to build/publish the mdbook](https://github.com/lalrpop/lalrpop/pull/724/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R20-R23) which is what is already in use in the `book.yml` workflow

youknowone (approved):
Thanks a lot!

bryantbiggs:
thank you! Looks like the pages and links are good now 🎉 https://lalrpop.github.io/lalrpop/

PR #726, opened 2023-03-25
Always use the 'input lifetime inside Symbol.

youknowone:
rebase of #311 because https://github.com/ahmedcharles/lalrpop doesn't exist anymore

### DIFF
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -456,7 +456,9 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                 .variant_names
                 .insert(Symbol::Nonterminal(nt.clone()), name.clone());
         }
-
+        if self.grammar.intern_token.is_some() {
+            rust!(self.out, "{}phantom_data(::std::marker::PhantomData<&'input ()>),", self.prefix);
+        }
         rust!(self.out, "}}");
         Ok(())
     }
--- a/lalrpop-test/src/empty.lalrpop
+++ b/lalrpop-test/src/empty.lalrpop
@@ -1,0 +1,3 @@
+grammar;
+
+pub Empty: () = ();
--- a/lalrpop-test/src/empty_tok.lalrpop
+++ b/lalrpop-test/src/empty_tok.lalrpop
@@ -1,0 +1,13 @@
+grammar;
+
+pub Empty: () = ();
+
+use empty_tok_lib;
+
+extern {
+    type Location = usize;
+    type Error = ();
+
+    enum empty_tok_lib::Tok {
+    }
+}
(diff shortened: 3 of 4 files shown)


lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -456,7 +456,9 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                 .variant_names
                 .insert(Symbol::Nonterminal(nt.clone()), name.clone());
         }
-
+        if self.grammar.intern_token.is_some() {
```
yannham, line 459: I think this new case deserves a comment describing why we need that (potentially linking to the issue as well). Other steps have a short one describing what happening, and I imagine it's not entirely clear out of the context of the original issue, why we need to do that in the first place.

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -456,7 +456,9 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
...
+            rust!(self.out, "{}phantom_data(::std::marker::PhantomData<&'input ()>),", self.prefix);
```
yannham, line 460: If I understand correctly, this is adding a variant to the corresponding enum to hold phantom data. Would this be sufficient to  add this variant only when both `self.grammar.nonterminals` and `self.grammar.terminals` are empty, instead of adding unconditionally?

yannham (commented):
I'm a bit surprised by the original pull request, as the original issue seemed to conclude that it's an edge case without actually much practical motivation, and that it would probably be fine to not do anything about empty grammars. I don't know what other maintainers think of this, but I'm personally a bit reluctant to merge stuff for the sake of merging stuff, if it doesn't have a clear use-case/motivation (that is, the empty grammar seems to be a pathological case, is it worth dealing with it?)

On the other hand, the change looks small and innocuous, so if we decide to go with it, count this review as an approve, as far as the code is concerned.

yannham (approved):
Ok, I reread the issue, and I might have misunderstood it a bit. There seemed to be an agreement about this shouldn't be a special case that doesn't work, even if it's not especially useful, so fine by me :+1: (beside the review comments)

youknowone:
I only rebased the patch to see how conflict goes but didn't edit anything including the original issues' comments.
And no plan to work on it for a while due to lack of comprehension of the issue.

@yannham please feel free to take it over if you want to

yannham:
I'll try my own suggestion when I have a bit time.

yannham:
I tried it but got a an error about int underflow, as the action code generated for the empty grammar as something like a transition to state `0 - 1`. I'll have to investigate further, but the empty token seems to mandate additional special handling. Honestly I'm not sure yet if it's a good way of spending time, given the motivation is mostly "it's not really useful but it's a special case I would expect to work". I can't guarantee I'll get this through.

PR #731, opened 2023-03-26
chore: Add Makefile to replace helper scripts used for local dev and by CI workflow

bryantbiggs:
- Helper scripts replaced by Makefile targets
- CI workflow updated to use Makefile; synchronizes the local use with the CI workflow

Tested and validated here https://github.com/clowdhaus/lalrpop/actions
mdbook still builds and deploys successfully https://clowdhaus.github.io/lalrpop/


nikomatsakis:
This does seem better, but I think better yet would be to reimplement in rust as a cargo xtask:

https://crates.io/crates/cargo-xtask

nikomatsakis:
(That said, I'm in favor of landing this PR)

bryantbiggs:
yes, I just didn't want to clutter up the space. I need a bit of time to look into xtask and then I'll open a new PR, but if anyone is more familiar, please don't let me hold you up

PR #734, opened 2023-04-03
lalrpop_mod! generates rustfmt::skip and allow(clippy::all)

youknowone:
Otherwise, creating a new macro for tests and letting users manually putting these stuff will be another way.


yannham (approved):
For what it worth, I personally don't see a situation where you would want to run clippy or rustfmt on the generated code indeed :+1:

youknowone:
Thats fair point

youknowone:
@yannham @Pat-Lafon I removed clippy par and only left fmt one

PR #735, opened 2023-04-04
Apply clippy suggestions

youknowone:


yannham (approved):
LGMT

Pat-Lafon:
Would it make sense to then add clippy to the the ci?

youknowone:
Not yet. I only applied trivial suggestions in this patch.

Issue #736, opened 2023-04-08
Expected tokens in error output sometimes include invalid entries


wabain:
With lalrpop 0.19.9, for certain grammars the expected tokens given for UnrecognizedTokens / UnrecognizedEOF errors frequently include invalid entries—tokens that could not be parsed successfully at the site of the error, regardless of what input follows. This was also the case on 0.19.8 and probably earlier releases. I found a minimal-ish grammar that reproduces the problem:

```rust
grammar;

pub Program: Vec<()> = Commands;

Commands: Vec<()> = (<Command> ";")*;

Command: () = {
    "case" <_matched:Word> "in" <_cases: (<CaseItem> ";;")*> <_last:CaseItem?> "esac" => (),
    Word => (),
};

CaseItem: () = {
    "("? <_pattern:Word> ")" <_block:Commands> => (),
};

Word: () = {
    r"[A-Z]+" => ()
};
```

This is a simplified subset of the Unix shell grammar. Legal productions are things like `X; case Q in R) A; B; ;; esac;`. Given the input `X` we get an UnrecognizedEOF error. The only token that could turn this into a valid production is `";"`, but the expected tokens given by lalrpop are `")", ";", "in"`.

This seems to be happening because the top state in the parser when the error is hit could handle any of those lookaheads; it would reduce `r"[A-Z]+" => Word` and leave the lookahead token to be shifted by another rule. In the case of the input `X` only `";"` could subsequently be accepted, but the expected token generation doesn't examine the state stack to determine this. The infrastructure to perform the full check [does exist](https://github.com/lalrpop/lalrpop/blob/ecd4dec013ac7540f6b6389d0173935f8c5e66af/lalrpop-util/src/state_machine.rs#L510-L525) (for the parse table based codegen, although not for recursive ascent).

wabain:
This is likely the same issue reported in #533, although I haven't checked.

dburgener:
This seems to have been fixed in https://github.com/lalrpop/lalrpop/pull/737, but the issue wasn't closed.  Closing now.

PR #737, opened 2023-04-08
Avoid spurious expected tokens in error output

wabain:
This pull request tries to fix the issue described in #736, where the expected tokens given in UnrecognizedToken and UnrecognizedEOF errors include values that would always be rejected if inserted at the error site. The idea is to reuse the existing logic for error recovery in order to filter the expected tokens down to ones which could actually be parsed at the point of the error. The strategy here is:

1. Cause some of the support code for error recovery (specifically `__simulate_reduce()` and `__accepts()`) to always be generated.
  Unfortunately this entails some extra code size, even for grammars that don't actually have states where invalid error output could be generated. However, it looks like `__simulate_reduce()` used to always be generated prior to #510 and it doesn't seem like there was any particular impetus to make it conditional beyond avoiding dead code. The size of the `__accepts()` function doesn't vary based on the input grammar.

2. Add a defaulted method `expected_tokens_from_states` to `lalrpop_util::state_machine::ParserDefinition` in order to provide the generated parser definition with enough context to correctly identify the legal tokens. The existing `expected_tokens` method is left alone, which makes this pull request partly backwards compatible; a lalrpop_util version with this pull request can be used with a preexisting generated grammar, although a generated grammar incorporating the changes here needs to use an updated lalrpop_util. If that makes this PR a breaking change then there's no need to keep `ParserDefinition::expected_tokens` in its current form.

This only fixes the issue for the table-based parser codegen. Fixing it for the recursive ascent codegen seems more difficult, which I imagine is why error recovery isn't implemented there either.

### DIFF
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -307,6 +307,20 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
         rust!(self.out, "}}");
 
         rust!(self.out, "");
+        rust!(
+            self.out,
+            "fn expected_tokens_from_states(&self, states: &[{state_type}]) -> alloc::vec::Vec<alloc::string::String> {{",
+            state_type = state_type,
+        );
+        rust!(
+            self.out,
+            "{p}expected_tokens_from_states(states, {pde})",
+            p = self.prefix,
+            pde = phantom_data_expr,
+        );
+        rust!(self.out, "}}");
+
+        rust!(self.out, "");
         rust!(self.out, "#[inline]");
         rust!(self.out, "fn uses_error_recovery(&self) -> bool {{");
         rust!(self.out, "{}", self.grammar.uses_error_recovery);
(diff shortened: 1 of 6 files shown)


lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -1538,8 +1540,18 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
...
         }
         rust!(self.out, "];");
+        Ok(())
+    }
 
-        // Grab any terminals in the current state which would have resulted in a successful parse
+    fn emit_expected_tokens_fn(&mut self) -> io::Result<()> {
+        rust!(
+            self.out,
+            "fn {p}expected_tokens({p}state: {}) -> alloc::vec::Vec<alloc::string::String> {{",
```
nikomatsakis, line 1549: Are there still callers for this more limited version?

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -1560,6 +1572,55 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
...
+            .emit()?;
+
+        rust!(self.out, "{{");
+
+        // Grab any terminals in the current state which would have resulted in a successful parse,
+        // as verified using accepts()
+        rust!(
+            self.out,
+            "{}TERMINAL.iter().enumerate().filter_map(|(index, terminal)| {{",
+            self.prefix,
```
nikomatsakis, line 1607:
```suggestion
            "{p}TERMINAL.iter().enumerate().filter_map(|(index, terminal)| {{",
            p = self.prefix,
```

we generally use a named variable for prefix, even if only used once, I think?

nikomatsakis (approved):
This looks great! Thanks.

lalrpop/src/lr1/codegen/parse_table.rs
nikomatsakis: meh, not worth waiting for it, we're a bit inconsistent already

PR #739, opened 2023-04-12
hide clippy warnings from generated sources in tests and docs

youknowone:
alternative #734

### DIFF
--- a/doc/calculator/src/main.rs
+++ b/doc/calculator/src/main.rs
@@ -1,7 +1,12 @@
-#[macro_use]
-extern crate lalrpop_util;
+use lalrpop_util::lalrpop_mod;
 
-lalrpop_mod!(pub calculator1); // syntesized by LALRPOP
+macro_rules! lalrpop_mod_doc {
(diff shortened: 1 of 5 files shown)


yannham:
As a preamble, what is the motivation for this change? Is clippy causing any trouble inside tests or in the CI currently?

Another question: should we try instead to fix the warnings, or is it a bad idea?

I'm trying to get a bit of context, apologies if the questions are naive.

youknowone:
To fix it correctly, appropriately using `#[automatically_derived]` will be helpful

nikomatsakis:
I know I'm late to this party, but it should be possible I think to structure the lalrpop output to automatically apply "allow clippy::all" to the generated code but not the action code. 

the idea would be to factor the generated code into a module and apply the attribute to that, but have the user code factored out, something like this:

```rust
mod lalrpop_mod {
#[allow(clippy::all)]
mod generated_code {
    user super::action_code;
}

mod action_code {
    ...
}

pub use generated_code::key_things_that_are_public;
}
```

PR #741, opened 2023-04-12
Rename acronyms to CamelCase

### LINKED ISSUE #740: Renaming acronyms names to meet Rust conventions before 1.0
- NFA -> Nfa (type)
- DFA -> Dfa (type)
- EOF -> Eof (enum)
- LR -> Lr (type)
- NT -> Nt (type)

This is breaking change because they are exposed as public types.
If we have desire to change the names fit for common conventions, making it before 1.0 will be better than later.
The previous names will be type aliases and const as `#[deprecated]` marked for backward compatibility.

# Unresolved questions
- Do we want to change the names?

# References
- https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

youknowone:
closes #740

While working on it, I found `LR*` stuffs. By watching there already was `Lr1Tls`, this change will resolve the mixing states.

Please make this ready for review once we decide to go.


nikomatsakis (approved):

nikomatsakis:
Has conflicts -- but r+ from me.

PR #743, opened 2023-04-15
Improve compilation speed by up to 2x

DaniPopes:
Running [samply](https://github.com/mstange/samply) to profile `lalrpop`, I found that most of the time spent during the `LR(1) state construction (lane)` phase is in the `continue_trace` recursive function, which calls `can_shift_nonterminal` in a loop. This last function looks like it should just perform a simple equality on two symbol references, but instead unnecessarily clones a symbol.
Other changes include reducing allocations and minor clippy fixes throughout the entire crate.

I've measured both [Solang] and [Rust Python], which yielded similar times, by using the time printed with the `--level verbose` flag:

-   debug: ~7.9s -> ~6.8s
-   release: ~1.1s -> ~0.45s

[solang]: https://github.com/hyperledger/solang/blob/a14139e916a6a89875d4a1e15c9548d0038b24cd/solang-parser/src/solidity.lalrpop
[rust python]: https://github.com/RustPython/RustPython/blob/c9546c2419f2e4db7e2987f30f0508dc29b30035/compiler/parser/python.lalrpop#L1

Another similar performance improvement can be seen by switching the `Map` and `Set` types to the standard `HashMap` with the [`Fx` hasher][fx], but I ran into a lot of test failures so I didn't include it in this PR. If this change is implemented correctly, I believe it can save another 1-2s on the debug build and ~0.1s on the release build.

[fx]: https://github.com/rust-lang/rustc-hash

### DIFF
--- a/lalrpop/src/lr1/state_graph.rs
+++ b/lalrpop/src/lr1/state_graph.rs
@@ -16,7 +16,9 @@ impl StateGraph {
     where
         L: Lookahead,
     {
-        let mut graph = Graph::new();
+        let nodes = states.len();
+        let edges = states.iter().map(|s| s.shifts.len() + s.gotos.len()).sum();
+        let mut graph = Graph::with_capacity(nodes, edges);
 
         // First, create the nodes.
         for i in 0..states.len() {
@@ -29,21 +31,18 @@ impl StateGraph {
             // Successors of a node arise from:
             // - shifts (found in the `conflicts` and `tokens` maps)
             // - gotos (found in the `gotos` map)
-            graph.extend_with_edges(
-                state
-                    .shifts
-                    .iter()
-                    .map(|(terminal, &state)| (Symbol::Terminal(terminal.clone()), state))
-                    .chain(
-                        state
-                            .gotos
-                            .iter()
-                            .map(|(nt, &state)| (Symbol::Nonterminal(nt.clone()), state)),
-                    )
-                    .map(|(symbol, successor)| {
-                        (NodeIndex::new(i), NodeIndex::new(successor.0), symbol)
-                    }),
-            );
+            let shifts = state
+                .shifts
+                .iter()
+                .map(|(terminal, &state)| (Symbol::Terminal(terminal.clone()), state));
+            let gotos = state
+                .gotos
+                .iter()
+                .map(|(nt, &state)| (Symbol::Nonterminal(nt.clone()), state));
+            let it = shifts.chain(gotos).map(|(symbol, successor)| {
+                (NodeIndex::new(i), NodeIndex::new(successor.0), symbol)
+            });
+            graph.extend_with_edges(it);
         }
 
         StateGraph { graph }
@@ -77,18 +76,20 @@ impl StateGraph {
         result
     }
 
-    pub fn successors(&self, state_index: StateIndex) -> Vec<StateIndex> {
+    pub fn successors(&self, state_index: StateIndex) -> impl Iterator<Item = StateIndex> + '_ {
         self.graph
             .edges_directed(NodeIndex::new(state_index.0), EdgeDirection::Outgoing)
             .map(|edge| StateIndex(edge.target().index()))
-            .collect()
     }
 
-    pub fn predecessors(&self, state_index: StateIndex, symbol: Symbol) -> Vec<StateIndex> {
+    pub fn predecessors<'a>(
+        &'a self,
+        state_index: StateIndex,
+        symbol: &'a Symbol,
+    ) -> impl Iterator<Item = StateIndex> + 'a {
         self.graph
             .edges_directed(NodeIndex::new(state_index.0), EdgeDirection::Incoming)
-            .filter(|edge| *edge.weight() == symbol)
+            .filter(move |edge| edge.weight() == symbol)
             .map(|edge| StateIndex(edge.source().index()))
-            .collect()
     }
 }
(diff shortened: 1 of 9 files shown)


youknowone:
Could you split the PR to trivial fix and actual speed up? The diff is large so it will be helpful to review.

DaniPopes:
Hey, thanks for the comment! The diff seems large at first glance but if you enable "Hide whitespace" on the Github settings and review commit-by-commit it will make it much easier.

The main performance change is in 187833a, Let me know if you still want me to create a new PR for the other changes.

yannham:
@DaniPopes thanks a lot for the PR, the numbers do look good.

I'll take a deeper look at it as well, but out of curiosity, what was your rationale for the `#[inline]` annotations? My understanding is that inlining is a complex matter, and most of the time `rustc` (well, actually, LLVM) probably knows better. The situation where it can be interesting to do so is:

- for libraries (as, as far as I understand, there's no cross-crate inlining happening without such an annotation)
- and, inside libraries, for functions that are performance-criticial (because there's a cost to inlining, in compilation time, but also because it can become a pessimisation)

Here I don't think we need cross-crate inlining, so I wonder if you just did a big change and measured that the result was positive, or if you measured specifically that some functions did gain something from inlining?

DaniPopes:
I saw that the `can_shift_nonterminal` small function accounted for more than half the time spent in the release build, so I preemptively added the `#[inline]` attribute, but it looks like the compiler/LLVM now can inline it by itself after removing the expensive `clone`. It's about 10% slower with a manual `#[inline(never)]`, so I think it's better to keep it.

youknowone:
@yannham unlike `#[inline(never)]` or `#[inline(always)]`, `#[inline]` is not that strong. 

By looking the function size, most of them looks reasonable. Not sure about comparably bigger ones.

yannham:
> I saw that the can_shift_nonterminal small function accounted for more than half the time spent in the release build, so I preemptively added the #[inline] attribute, but it looks like the compiler/LLVM now can inline it by itself after removing the expensive clone. It's about 10% slower with a manual #[inline(never)], so I think it's better to keep it.

I think if this doesn't make a difference with no inline annotation at all in your working example (of course `#[inline(never)]` is going to be worse, but I think that's a bit beside the point), I would personally strip off the annotations. I know `#[inline]` isn't too strong, but it's still overriding in some way the decisions of `rustc` and `LLVM`, and to the best of my knowledge, isn't a common practice when you don't care about cross-crate inlining. Also, when you start to add inline hints, you have to start thinking well why this function and not this function, and you either have to come with a rationale or make somehow arbitrary micro-decisions. If we can just forget about inlining altogether and let that work to the optimizer, I looks like a win in the end!

I general I would vouch for adding code/changing code for performance reason when it's clearly motivated, that is, if adding the `#[inline]` does improve the situation in a measurable way with no annotation at all, and not as a kind of preventive measure. This just my humble opinion, and I won't die on this hill either if you're absolutely sure that this is a better idea to keep the annotation :slightly_smiling_face: 

Maybe @nikomatsakis has something to say here as well, as he's probably more competent than me on those kind of matters.

yannham:
@DaniPopes I started to review but there are spurious diff due to a PR implementing clippy suggestions that has been merged in between. Could you try a rebase please whenever you have the time to do so, and then ping me?

youknowone:
@yannham I rebased it for review if you have time now.

@DaniPopes Please rebase it again to verify signing when you have time. And additional notes. Me and @yannham are not the original owners of this repository, so tending to being conservative on decisions when we don't know about consensus. (@yannham do you agree?)

I think this PR will be more easily merged by being split to 5 pieces 
1. clippy before 1.56 (2021 edition)
2. clippy after 1.56
3. RustWrite
4. removing allocations
5. adding `#[inline]`. 

1, 3, 4 looks to be merged quickly. 2 has dependency. no idea about 5.

DaniPopes:
@yannham @yannham I've rebased onto master and squashed the commits down to one, since they have been split off into different PRs (points 1 and 2: #745, point 3: #744). I've removed all `#[inline]` attributes (point 5), which leaves this PR addressing point number 4.

youknowone (approved):
Looks good! thank you for contributing and patient for chore jobs

lalrpop/src/lr1/state_graph.rs
```diff
@@ -29,21 +31,18 @@ impl StateGraph {
...
+                .shifts
+                .iter()
+                .map(|(terminal, &state)| (Symbol::Terminal(terminal.clone()), state));
+            let gotos = state
+                .gotos
+                .iter()
+                .map(|(nt, &state)| (Symbol::Nonterminal(nt.clone()), state));
+            let it = shifts.chain(gotos).map(|(symbol, successor)| {
+                (NodeIndex::new(i), NodeIndex::new(successor.0), symbol)
+            });
```
nikomatsakis, line 44: this is just code simplification?

nikomatsakis (approved):

nikomatsakis:
This LGTM as well, thanks.

lalrpop/src/lr1/state_graph.rs
youknowone: Yes, `shifts.chain(gotos).` part was the equivalent longer chain before.

youknowone:
@DaniPopes Thank you from RustPython by using it for benchmark 😉

Issue #742, opened 2023-04-14
Define MSRV for 1.0


youknowone:
What will be the minimum supported rust version requirement of lalrpop? By looking CI, it looks like just `stable`.
That can be fine. But documenting will be a good idea.

- Our MSRV: Current stable version of Rust https://github.com/lalrpop/lalrpop/issues/742#issuecomment-1518503250
- [ ] Update cargo.toml
- [ ] Update contributing.md
- [x] Update CI to run on matching rust version

youknowone:
@Pat-Lafon  That sounds like a good suggestion.

Pat-Lafon:
I believe the msrv can also be specified in the `cargo.toml`, maybe add that to checklist?

Pat-Lafon:
I missed the `#[default]` suggestion aswell in #744, this was introduced in 1.62... just of note.

youknowone:
I'd like to suggest to mark 1.60 as current MSRV and listen more users' feedback before going to 1.0.
There are only 6 months gap after the 2021 edition, but will prohibit lalrpop users to have fragmented copies of `regex` from their dependency tree.

We can make it back or forward later before actually mark 1.0 - though backward will be a bit painful.

cc @nikomatsakis @yannham

youknowone:
One of my wishlist is this one: https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds

nikomatsakis:
I would not want to hold ourselves to the same standard as regex, personally. I'm not a big fan of old MSRVs. We don't backport security requirements and I don't think that people should use older builds. I recognize that just declaring stable is a bit aggressive. I would be more inclined to do something like "N-2" -- i.e., we track stable, but we hold off on adopting newer features gratuitously.

youknowone:
That sounds good enough. I will remove 1.60 build from CI and add the stable policy to Readme.
N-2 looks good enough to add message to pull request template rather than readme.

yannham:
[...]

PR #727, opened 2023-03-25
fix: Update edition to 2021

bryantbiggs:
- Update edition to 2021
- Update `pico-args` to 0.5
- Enable sparse index in CI - https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html

PR #744, opened 2023-04-18
Improve RustWrite + Modern style work by @DaniPopes

youknowone:
split from #743 for review 

cc @DaniPopes

### DIFF
--- a/lalrpop/src/main.rs
+++ b/lalrpop/src/main.rs
@@ -2,7 +2,6 @@ extern crate lalrpop;
 extern crate pico_args;
 
 use std::ffi::OsString;
-use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process;
 use std::str::FromStr;
@@ -11,9 +10,9 @@ use pico_args::Arguments;
 
 use lalrpop::Configuration;
 
-static VERSION: &str = env!("CARGO_PKG_VERSION");
+const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-const USAGE: &str = "
+const USAGE: &str = "\
 Usage: lalrpop [options] <inputs>...
        lalrpop --help
        lalrpop (-V | --version)
@@ -29,7 +28,7 @@ Options:
     -c, --color          Force colorful output, even if this is not a TTY.
     --no-whitespace      Removes redundant whitespace from the generated file. (Default: false)
     --comments           Enable comments in the generated code.
-    --report             Generate report files.
+    --report             Generate report files.\
 ";
 
 #[derive(Debug)]
@@ -65,12 +64,12 @@ impl FromStr for LevelFlag {
             "info" => Ok(Info),
             "verbose" => Ok(Verbose),
             "debug" => Ok(Debug),
-            x => Err(format!("Unknown level {}", x)),
+            x => Err(format!("Unknown level: {x}")),
         }
     }
 }
 
-fn parse_args(mut args: Arguments) -> Result<Args, Box<dyn std::error::Error>> {
+fn parse_args(mut args: Arguments) -> Result<Args, pico_args::Error> {
     Ok(Args {
         flag_out_dir: args.opt_value_from_fn(["-o", "--out-dir"], PathBuf::from_str)?,
         flag_features: args.opt_value_from_str("--features")?,
@@ -86,24 +85,17 @@ fn parse_args(mut args: Arguments) -> Result<Args, Box<dyn std::error::Error>> {
     })
 }
 
-fn main() {
-    main1().unwrap();
-}
-
-fn main1() -> Result<(), Box<dyn std::error::Error>> {
-    let mut stderr = std::io::stderr();
-    let mut stdout = std::io::stdout();
-
+fn main() -> Result<(), Box<dyn std::error::Error>> {
     let args = parse_args(Arguments::from_env())?;
 
     if args.flag_help {
-        writeln!(stdout, "{}", USAGE)?;
-        process::exit(0);
+        println!("{USAGE}");
+        return Ok(());
     }
 
     if args.flag_version {
-        writeln!(stdout, "{}", VERSION)?;
-        process::exit(0);
+        println!("{VERSION}");
+        return Ok(());
     }
 
     let mut config = Configuration::new();
@@ -136,14 +128,10 @@ fn main1() -> Result<(), Box<dyn std::error::Error>> {
     }
 
     if args.arg_inputs.is_empty() {
-        writeln!(
-            stderr,
-            "Error: no input files specified! Try --help for help."
-        )?;
-        process::exit(1);
+        return Err("Error: no input files specified! Try --help for help.".into());
     }
 
-    if let Some(ref out_dir) = args.flag_out_dir {
+    if let Some(out_dir) = args.flag_out_dir {
         config.set_out_dir(out_dir);
     }
 
@@ -156,12 +144,7 @@ fn main1() -> Result<(), Box<dyn std::error::Error>> {
         match config.process_file(arg) {
             Ok(()) => {}
             Err(err) => {
-                writeln!(
-                    stderr,
-                    "Error encountered processing `{}`: {}",
-                    arg.display(),
-                    err
-                )?;
+                eprintln!("Error encountered processing `{}`: {}", arg.display(), err);
                 process::exit(1);
             }
         }
@@ -210,7 +193,7 @@ mod test {
     #[test]
     fn test_usage_out_dir() {
         let args = parse_args_slice(&["--out-dir", "abc", "file.lalrpop"]);
-        assert_eq!(args.flag_out_dir, Some(PathBuf::from_str("abc").unwrap()));
+        assert_eq!(args.flag_out_dir.as_deref(), Some(Path::new("abc")));
         assert_eq!(args.arg_inputs, ["file.lalrpop"]);
     }
 
--- a/lalrpop/src/rust/mod.rs
+++ b/lalrpop/src/rust/mod.rs
@@ -7,10 +7,25 @@ use crate::tls::Tls;
 use std::fmt::{self, Display};
 use std::io::{self, Write};
 
+// The `rust` macro should be called only on a `RustWrite` instance.
+#[doc(hidden)]
+#[inline(always)]
+#[cfg(debug_assertions)]
+pub const fn assert_rust_write<W>(_: &RustWrite<W>) {}
+
+/// Like [`std::writeln!`], but for writing Rust code to a [`RustWrite`], which handles indentation.
 macro_rules! rust {
-    ($w:expr, $($args:tt)*) => {
-        ($w).writeln(&::std::fmt::format(format_args!($($args)*)))?
-    }
+    ($w:expr) => {{
+        #[cfg(debug_assertions)]
+        let _ = $crate::rust::assert_rust_write(&$w);
+        ::std::writeln!($w)?;
+    }};
+
+    ($w:expr, $($args:tt)*) => {{
+        #[cfg(debug_assertions)]
+        let _ = $crate::rust::assert_rust_write(&$w);
+        ::std::writeln!($w, $($args)*)?;
+    }};
 }
 
 /// A wrapper around a Write instance that handles indentation for
@@ -33,14 +48,14 @@ macro_rules! rust {
 /// }
 /// }
 /// ```
-pub struct RustWrite<W: Write> {
+pub struct RustWrite<W> {
     write: W,
     indent: usize,
 }
 
 const TAB: usize = 4;
 
-impl<'me, W: Write> RustWrite<W> {
+impl<W: Write> RustWrite<W> {
     pub fn new(w: W) -> RustWrite<W> {
         RustWrite {
             write: w,
@@ -59,11 +74,6 @@ impl<'me, W: Write> RustWrite<W> {
         Ok(())
     }
 
-    fn write_indented(&mut self, out: &str) -> io::Result<()> {
-        self.write_indentation()?;
-        writeln!(self.write, "{}", out)
-    }
-
     pub fn write_table_row<I, C>(&mut self, iterable: I) -> io::Result<()>
     where
         I: IntoIterator<Item = (i32, C)>,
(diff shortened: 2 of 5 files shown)


lalrpop/src/main.rs
```diff
@@ -65,12 +64,12 @@ impl FromStr for LevelFlag {
             "info" => Ok(Info),
             "verbose" => Ok(Verbose),
             "debug" => Ok(Debug),
-            x => Err(format!("Unknown level {}", x)),
+            x => Err(format!("Unknown level: {x}")),
```
youknowone, line 67: This is pretty recent changes and depending on #742

lalrpop/src/normalize/precedence/mod.rs
```diff
@@ -70,20 +70,15 @@ pub const SIDE_ARG: &str = "side";
 /// An associative rule means that all recursive occurrences are replaced with the current level,
 /// which is different from non-associativity. This can be useful for unary operators that may be
 /// iterated, such as `-` or `!`, or non ambiguous operators. This is the default associativity.
-#[derive(Clone, Copy, Eq, PartialEq)]
+#[derive(Clone, Copy, Eq, PartialEq, Default)]
 pub enum Assoc {
     Left,
     Right,
     NonAssoc,
+    #[default]
```
youknowone, line 78: This is pretty recent changes and depending on #742

lalrpop/src/rust/mod.rs
```diff
@@ -82,32 +92,48 @@ impl<'me, W: Write> RustWrite<W> {
...
+    pub fn write_fmt(&mut self, args: fmt::Arguments<'_>) -> io::Result<()> {
+        // avoid intermediate allocation for simple format strings
+        let _storage;
+        let buf = match args.as_str() {
+            Some(s) => s.as_bytes(),
+            None => {
+                _storage = fmt::format(args);
+                _storage.as_bytes()
+            }
+        };
```
youknowone, line 116:
```suggestion
        let formatted;
        let buf = match args.as_str() {
            Some(s) => s.as_bytes(),
            None => {
                formatted = fmt::format(args);
                formatted.as_bytes()
            }
        };
```

lalrpop/src/rust/mod.rs
```diff
@@ -82,32 +92,48 @@ impl<'me, W: Write> RustWrite<W> {
...
+            [b'\n'] => return self.write.write_all(b"\n"),
+            [.., b'\n'] => {}
+            _ => unreachable!("rust! macro should always end with a newline"),
         }
 
-        let n = buf.len() - 1;
-
         // If the line begins with a `}`, `]`, or `)`, first decrement the indentation.
-        if buf[0] == b'}' || buf[0] == b']' || buf[0] == b')' {
+        if matches!(buf.first().unwrap(), b'}' | b']' | b')') {
```
youknowone, line 127:
```suggestion
        if [b'}', b']', b')'].contains(buf.first().unwrap()) {
```

youknowone (commented):
I personally support the changes, but the project was edition 2015 a few weeks ago.
I'd like to check what's the consensus of minimum rust version for this project.

lalrpop/src/rust/mod.rs
DaniPopes: It probably gets optimized to the same code, but why `array.contains` instead of a match?

DaniPopes:
Understandable, although I saw that the `lalrpop` crate itself doesn't have an edition specified in the manifest file, meaning it's still on edition 2015. I'm assuming this was missed in #727

lalrpop/src/rust/mod.rs
youknowone: I guess so about the optimized result.

When I suggested it, I thought this is more like "one of collections" test rather than "pattern matching test". So suggested more typical expression for it.
But it may be regarded different for other people.

lalrpop/src/main.rs
```diff
@@ -86,24 +85,17 @@ fn parse_args(mut args: Arguments) -> Result<Args, Box<dyn std::error::Error>> {
...
-    let mut stderr = std::io::stderr();
-    let mut stdout = std::io::stdout();
-
+fn main() -> Result<(), Box<dyn std::error::Error>> {
     let args = parse_args(Arguments::from_env())?;
 
     if args.flag_help {
-        writeln!(stdout, "{}", USAGE)?;
-        process::exit(0);
+        println!("{USAGE}");
```
yannham, line 92: Beware that this is not equivalent to previous code. `println` panic in case of error, while `writeln` will bubble up the error. See https://github.com/BurntSushi/advent-of-code/issues/17#issuecomment-475443686. I'm not sure it makes a difference here, at least for now, as we are propagating the error to `main` anyway, but I wonder if we should keep the `writeln` by principle, so that if we one day we want to catch the error and do something else, we can?

lalrpop/src/main.rs
```diff
@@ -86,24 +85,17 @@ fn parse_args(mut args: Arguments) -> Result<Args, Box<dyn std::error::Error>> {
...
+        return Ok(());
     }
 
     if args.flag_version {
-        writeln!(stdout, "{}", VERSION)?;
-        process::exit(0);
+        println!("{VERSION}");
```
yannham, line 97: Same remark.

lalrpop/src/main.rs
```diff
@@ -156,12 +144,7 @@ fn main1() -> Result<(), Box<dyn std::error::Error>> {
         match config.process_file(arg) {
             Ok(()) => {}
             Err(err) => {
-                writeln!(
-                    stderr,
-                    "Error encountered processing `{}`: {}",
-                    arg.display(),
-                    err
-                )?;
+                eprintln!("Error encountered processing `{}`: {}", arg.display(), err);
```
yannham, line 147: Same remark.

lalrpop/src/rust/mod.rs
```diff
@@ -82,32 +92,48 @@ impl<'me, W: Write> RustWrite<W> {
...
+        };
 
         // pass empty lines through with no indentation
-        if buf.is_empty() {
-            return self.write.write_all("\n".as_bytes());
+        match buf {
+            [] => return Ok(()),
+            [b'\n'] => return self.write.write_all(b"\n"),
+            [.., b'\n'] => {}
+            _ => unreachable!("rust! macro should always end with a newline"),
```
yannham, line 123: It's a nitpick, but I wonder if `unreachable!` is the right semantics. I think the common usage is that `unreachable` is used when you can prove locally that the codepath will never be taken, for example:

```
if foo.is_none() {
  do_stuff();
  match foo {
    Some(_) => unreachable!()
    None => ...,
  }
}
```

Here this is an invariant that must be ensured by the caller of `rust!`, and it can most probably be broken at some point by some careless or unknowing maintainer. Maybe just a `panic!` instead? (Of course, it's mostly about good practices , because it doesn't change much in practice: this code just panics the same).

lalrpop/src/rust/mod.rs
yannham: As a datapoint, I think I prefer the pattern matching, or the initial test, because I feel like pattern matching or equality are more fundamental and common language constructs than slices and a call to the `contains` method. But it's mostly a question of taste, at this point.

yannham: On the other hand, is `buf.first().unwrap()` much better? We unwrap anyway, and it just feels a bit more cluttering (also, we use `buf[n]` below, instead of `get(n).unwrap()`)

lalrpop/src/rust/mod.rs
```diff
@@ -82,32 +92,48 @@ impl<'me, W: Write> RustWrite<W> {
...
         }
 
-        self.write_indented(out)?;
+        self.write_indentation()?;
+        self.write.write_all(buf)?;
 
-        // Detect a line that ends in a `{` or `(` and increase indentation for future lines.
-        if buf[n] == b'{' || buf[n] == b'[' || buf[n] == b'(' {
+        // If a line ends with a `{`, `[`, or `(`, increase indentation for future lines.
+        let n = buf.len().saturating_sub(2);
```
yannham, line 135: I think the removed `n` calculation was `let n = buf.len() - 1;`, but now it seems to be `buf.len() - 2`?

tools/ci.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
yannham, line 1: As a Nix user, I prefer the more portable `env` version :stuck_out_tongue: 

```suggestion
#!/usr/bin/env bash
```

lalrpop/src/rust/mod.rs
DaniPopes: Given that `buf.len() >= 2`, both index ops can never panic, though they should be written in the same style. I agree on the equality point, but personally I'd use pattern matching because it's much more concise and a very nice Rust language feature: `x, A | B | C` vs `x == A || x == B || x == C`.

lalrpop/src/main.rs
DaniPopes: The linked comment applies here too, especially since printing isn't even the purpose of the application, and is only done for errors and the version/help strings.

lalrpop/src/rust/mod.rs
yannham: Yeah, I said equality because it's fine as well, but I also find the pattern matching to be the better as well, FWIW

lalrpop/src/rust/mod.rs
yannham: Yeah, but my point is that it _could be_ called without the trailing `\n` by a careless author in the future, couldn't it? If yes, then I think unreachable isn't the right semantics. I think unreachable indicates code which is, by all means and whatever the caller is passing to this function, can be proven (on paper) to never be triggered, but the compiler isn't smart enough to prove it. However, I don't think anything forbids at compile time to call this function without a trailing `\n`. So it doesn't look like unreachable code to me, but rather it looks like an invariant which is assumed to be enforced throughout the codebase.

tools/ci.sh
yannham: Ok, fair enough, if we change this we should change every instance, and this can be done in a separate PR. My previous comment can be ignored

lalrpop/src/main.rs
yannham: Did you mean "doesn't apply"?

DaniPopes: I'm referring to the last paragraph:

> println! is fine in short programs or where this sort of thing doesn't matter,
but I was trying to write idiomatic code, and IMO, idiomatic command line
programs don't panic.

lalrpop/src/rust/mod.rs
yannham: Ah, right, I see. Thanks.

lalrpop/src/main.rs
yannham: I still think the error message is different: one is panicking with a line number pointing to the inner code of `lalrpop`, while bubbling up an error would terminate with a `EXIT_FAILURE` and no mention of any rust source code? I haven't tested it. But I think the latter is nicer: a panic looks like something went horribly wrong and is a bug of `lalrpop`, while something like a broken pipe on printing is really not lalrpop's fault.

lalrpop/src/rust/mod.rs
DaniPopes: While it is technically possible for a careless author to call the function without the trailing '\n', it is highly unlikely given that it is an implementation detail, being only public because of the needs of the macro. Someone would have to go out of their way by calling `x.write_fmt(format_args!("..."))` instead of `rust!(x, "...")`.

But we're arguing over semantics here, in practice both just panic at runtime.

yannham: I think I wouldn't use unreachable even if the method wasn't public. I don't think it's an issue of public symbols versus non public: it's an issue of locality. Exactly like `unsafe`: you can declare your method safe if you can enforce/prove the safety invariants _locally_, notwithstanding the caller. If you depend on your caller satysifying invariants, you have to declare your method unsafe as well. It's just even worse if it's public, because now any user of LALRPOP can break the invariant and potentially see the result of an `unreachable` call.

The LALRPOP developers and maintainers are subject to mistakes as well, come and go, and don't know the all of the codebase in depth. In general the whole point of safety isn't to say oh this is unlikely to happen, but rather try to foresee carefully anything that could happen, even if not very probable. Of course the cost of safety is to be taken into account (verbosity/complexity), but here it's zero.

A good test is: would you feel confident to change your `unreachable` to `unreachable_unchecked`? I know I wouldn't. I don't really care about this specific case, it doesn't change much with respect to safety or failure mode, but it's more that I don't really understand what is the argument for keeping unreachable either. The error message and the underlying intents are still a bit different, if this ever triggers.

lalrpop/src/rust/mod.rs
youknowone: Thank you all. That sounds reasonable.

lalrpop/src/rust/mod.rs
yannham: Reading the (granted, sparse) [documentation of unreachable](https://doc.rust-lang.org/std/macro.unreachable.html), I respectfully disagree. There doesn't seem to be a "user vs library" distinction there, and the examples doesn't seem to include "breaking an arbitrary invariant required by the function" but more about what I describe above (the this code path is actually impossible because I've tested the condition before but the compiler can't prove it). That being said, a bit more research seems to show that your interpretation does exist in the wild as well.

 I won't fight this further though, as the impact is limited in practice, our personal time is valuable, and @youknowone seems to have a different interpretation than me. @DaniPopes I let you do whatever you think is best and then resolve the conversation :slightly_smiling_face:

yannham: Though, in this specific case, keep in mind that this function is still part of the public API. So even with a laxer interpretation, I still think it's not a great idea to let a possible codepath where a consumer can trigger unrechable exists.

youknowone: I agree my interpretation is definitely out of the examples in the document (and barely fit in the description, possibly just unintended ambiguity)

lalrpop/src/main.rs
youknowone: This is a part of user interface. This change looks ok to me.

lalrpop/src/rust/mod.rs
youknowone: I am going to leave it as `unreachable!` because lalrpop already use unreachable! for logical unreachable code in `fn symbol_table`.

youknowone (commented):
@DaniPopes I am going to open a new PR with less arguable parts only in #793 
I am going to close this PR. Please open a new one with remaining parts to fully merge other parts.
Thank you so much and sorry to make it delayed for many days.

PR #746, opened 2023-04-18
Clippy(mostly in generated code)

Pat-Lafon:
Inspired by recent movement/discussion. I wanted to take a crack at some of the suppressed clippy lints, largely in generated code which I think hasn't been touched on yet.

In doing so, I also suggest a series of four lints that I think always want to be suppressed or atleast can be handled/fix more specifically.

```
        #[allow(clippy::extra_unused_lifetimes)]
        #[allow(clippy::needless_lifetimes)]
        #[allow(clippy::let_unit_value)]
        #[allow(clippy::just_underscores_and_digits)]
```

I roughly tried to keep things to separate commits but am happy to clean the history up/explain things as needed.

### DIFF
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -191,6 +191,8 @@ macro_rules! lalrpop_mod {
 
     ($(#[$attr:meta])* $vis:vis $modname:ident, $source:expr) => {
         #[rustfmt::skip]
+        #[allow(clippy::extra_unused_lifetimes)]
+        #[allow(clippy::just_underscores_and_digits)]
         $(#[$attr])* $vis mod $modname { include!(concat!(env!("OUT_DIR"), $source)); }
     };
 }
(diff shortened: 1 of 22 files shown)


lalrpop-util/src/lib.rs
```diff
@@ -191,6 +191,10 @@ macro_rules! lalrpop_mod {
 
     ($(#[$attr:meta])* $vis:vis $modname:ident, $source:expr) => {
         #[rustfmt::skip]
+        #[allow(clippy::extra_unused_lifetimes)]
+        #[allow(clippy::needless_lifetimes)]
```
youknowone, line 195: I understand the below 2 allows are justified as generated code and easy to avoid for human.
But is this lifetimes too?

youknowone (approved):
looks good, much better than before.

lalrpop/src/lexer/re/mod.rs
```diff
@@ -7,7 +7,7 @@ use regex_syntax::{self, Error, Parser};
 mod test;
 
 pub type Regex = Hir;
-pub type RegexError = Error;
+pub type RegexError = Box<Error>;
```
yannham, line 10: Is it a suggestion because of the size of `Error`?

yannham (approved):
LGMT :+1: thanks for contributing!

lalrpop-util/src/lib.rs
yannham: I think it's reasonable for the scope of this PR. As you say, from a quick glance, it might be doable, but it doesn't look totally trivial at least. If you're happy with this answer @youknowone I think you can press the green button :slightly_smiling_face:

PR #747, opened 2023-04-19
update lalrpop/lalrpop-test to 2021 edition

Pat-Lafon:
As noted in https://github.com/lalrpop/lalrpop/pull/744#issuecomment-1514099999, #727 did not update the edition in all of the workspace members.

I've taken the liberty to update the two remaining `Cargo.toml` files and update the corresponding tests. I think some 2015 edition idioms were left over?

I was thinking about putting the edition in the workspace toml file using https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table, though that uses Rust 1.64 (related to #742).


youknowone (approved):
Thank you!
And yes! I agree your suggestion is the best way.

Issue #675, opened 2022-07-17
"use" declaration of type used only in "extern" triggers unused import error


astralarya:
If you only use a type in the "extern" block, the imported type is considered unused.  This is probably because the extern content only occurs within modules in the file that reimport the type, making the top level import redundant.

PR #748, opened 2023-04-20
Clippy --all-targets and RangeInclusive

Pat-Lafon:
One more pr with clippy lints from `cargo clippy --all-targets`.

With this, there are three sets of lints left.

- Acronym lints which will be fixed with #740 
- unused import in the nested test case #675
- `#[default]` lint #744

Adjusting `Test` in the lexer to use an actual `Range` was non-trivial. The ranges are documented in comments as being inclusive, though the translation to using `RangeInclusive` was a bit more than expected. I think at some times the ranges were actually treated as exclusive as well, as denoted by the tests. I believe it is now consistent, passing tests, and satisfying the clippy lint but that commit could use an extra look.

Using `RangeInclusive` does mean the struct is no longer `Copy` but I don't think that's too much of an issue.

### DIFF
--- a/lalrpop/src/lexer/dfa/overlap.rs
+++ b/lalrpop/src/lexer/dfa/overlap.rs
@@ -30,8 +30,8 @@ pub fn remove_overlap(ranges: &Set<Test>) -> Vec<Test> {
 
     let mut disjoint_ranges = vec![];
 
-    for &range in ranges {
-        add_range(range, 0, &mut disjoint_ranges);
+    for range in ranges {
+        add_range(range.to_owned(), 0, &mut disjoint_ranges);
     }
 
     // the algorithm above leaves some empty ranges in for simplicity;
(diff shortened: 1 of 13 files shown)


lalrpop/src/main.rs
```diff
@@ -179,24 +179,24 @@ mod test {
         vals.iter().map(|v| v.into()).collect()
     }
 
-    fn parse_args_vec(args: &Vec<&str>) -> Args {
+    fn parse_args_vec(args: &[&str]) -> Args {
```
youknowone, line 182:
```suggestion
    fn parse_args(args: &[&str]) -> Args {
```
It's not  a vec anymore

youknowone (changes requested):
Looks good. I left a name suggestion.

lalrpop/src/lexer/dfa/overlap.rs
```diff
@@ -30,8 +30,8 @@ pub fn remove_overlap(ranges: &Set<Test>) -> Vec<Test> {
 
     let mut disjoint_ranges = vec![];
 
-    for &range in ranges {
-        add_range(range, 0, &mut disjoint_ranges);
+    for range in ranges {
+        add_range(range.clone(), 0, &mut disjoint_ranges);
```
youknowone, line 34:
```suggestion
        add_range(range.to_owned(), 0, &mut disjoint_ranges);
```
No real difference, but to avoid to looking redundant clone. `to_owned()` will clear range is reference.

youknowone (approved):
Thank you so much!

PR #751, opened 2023-04-21
Enable 'unicode' feature on regex-syntax

### LINKED ISSUE #750: Regex version 1.8.0 breaks build
Yesterday the regex crate updated it's version to 1.8.0. As a result lalrpop is no longer able to parse correctly. 
A issue on the regex crate has already been opened here https://github.com/rust-lang/regex/issues/982 
As noted by the author this seems to to only occur when no unicode feature flag is set on the regex-syntax dependency, which seems to be the case here, default_features = false.

Output:

```
Caused by:
  process didn't exit successfully: `<...>/build-script-build` (exit status: 1)
  --- stdout
  processing file `<...>/cel.lalrpop`
  Phase `Grammar validation` begun
  Phase `Grammar validation` completed in 0.000351976 seconds
  Phase `Grammar resolution` begun
  Phase `Grammar resolution` completed in 0.000095088 seconds
  Phase `Precedence expansion` begun
  Phase `Precedence expansion` completed in 0.000054054 seconds
  Phase `Macro expansion` begun
  Phase `Macro expansion` completed in 0.000255872 seconds
  Phase `Token check` begun
  /<...>/cel.lalrpop:8:4: 8:16 error: invalid regular expression: regex parse error:
      \s*
      ^^

  --- stderr
       r"\s*" => { },
       ~~~~~~~~~~~~~
    
   
   ```

### PR DESCRIPTION
**Pull request by jannic**
This fixes a compatibility issue with regex 1.8

Closes #750

### DIFF
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -24,7 +24,7 @@ is-terminal = "0.4.2"
 itertools = { version = "0.10", default_features = false, features = ["use_std"] }
 petgraph = { version = "0.6", default_features = false }
 regex = { version = "1", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.6", default_features = false }
+regex-syntax = { version = "0.6", default_features = false, features = ["unicode"] }
 string_cache = { version = "0.8", default_features = false }
 term = { version = "0.7", default_features = false }
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }

### CODE REVIEW

**Comment by dburgener** [comment]
Are we sure the unicode feature should be enabled on the regex_syntax dependency, rather than the regex_dependency?

I haven't dug in too deep, but my read on https://github.com/rust-lang/regex/issues/982 (and particularly https://github.com/rust-lang/regex/issues/982#issuecomment-1517846726) is that the recommended solution is to enable the feature on the *regex* crate, rather than *regex_syntax*

**Comment by jannic** [comment]
> Are we sure the unicode feature should be enabled on the regex_syntax dependency, rather than the regex_dependency?

Adding the feature to the regex dependency wouldn't help. In fact, that's exactly the way this issue was triggered in the first place: With regex updated to a version >= `1.8.0`, regex internally depends on `regex_syntax` version 0.7. Enabling the `unicode` feature on `regex` would enable the same feature on `regex_syntax` 0.7, but that would not automatically enable that feature on `regex_syntax` 0.6. And that's the version `lalrpop` is using. And as feature unification doesn't work across different crate versions, the feature would still be missing there.

**Comment by jannic** [comment]
> Without looking at your specific situation, my recommendation would be to set `regex = "1.8.1"` and `regex-syntax = "0.7.1"` and enable `unicode` in both. This might require some changes to handle the new `regex-syntax 0.7` release, but I expect/hope that they will be minor.

Doesn't that impose the risk that it might be a breaking change, in case some internal of `regex_syntax` is visible in the API of `lalrpop`? Therefore I think for a pure bugfix release, `lalrpop` should stay with the current dependency on `regex_syntax` 0.6, even if that means that two versions of that crate are in the dependency tree.

Of course, for the next major release of `lalrpop`, the dependency on `regex_syntax` should be bumped as well.

**Comment by jannic** [comment]
I don't know the internals of lalrpop either, so I only suggest not updating `regex-syntax` to a new semver version out of caution.
Regarding the `unicode` feature on the `regex` dependency, I can only say that for my test cases, it was sufficient to add it to the `regex_syntax` dependency. It might of course still be "more correct" to add the feature to both dependencies.

File: lalrpop/Cargo.toml [hunk 1]
```diff
@@ -24,7 +24,7 @@ is-terminal = "0.4.2"
 itertools = { version = "0.10", default_features = false, features = ["use_std"] }
 petgraph = { version = "0.6", default_features = false }
 regex = { version = "1", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.6", default_features = false }
+regex-syntax = { version = "0.6", default_features = false, features = ["unicode"] }
```
youknowone, line 27: Rather than making this change, let's pin `regex` to `1.7` for a while.
```
regex = { version = "=1.7", default_features = false, features = ["std"] }
```

youknowone (changes requested):
@BurntSushi Thank you for the advice.

I found `regex-syntax` has breaking changes. Because regex 1.7 and regex-syntax 0.6 seem pair, I'd like to earn time to follow the changes by pinning regex to `1.7` for now.

lalrpop/Cargo.toml
BurntSushi: Why? `=` constraints are really bad for the ecosystem and can lead to unresolveable build failures. And `lalrpop` is compatible with regex `1.8`.

youknowone: Oh, yep. I got something wrong while changing regex-syntax version a few times.

youknowone (approved):

PR #752, opened 2023-04-21
Fail CI if any lines fail

dburgener:
The default bash return code is the return code of the last function called.  This means that the CI will only mark as failed if the last command in ci.sh fails.

This can be observed in recent CI runs (for example https://github.com/lalrpop/lalrpop/actions/runs/4765777324/jobs/8471946793) where tests failed, but the CI marked as passed anyways.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -1,3 +1,5 @@
+set -e -u -o pipefail
+
 export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
 export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse


dburgener:
docs on the bash set flags: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

yannham (approved):
Sounds reasonable to enable those flags :+1: thanks for catching the issue. I'm reviewing your other PR right now, once it's merge and you've rebase this one, could you ping me again please?

dburgener:
@yannham Thanks!  I've rebased and CI for this PR is passing now that #753 has merged.

youknowone:
Thank you!

@yannham I am not understanding the merge queue very well yet, but just a guess. Was merge queue button enabled for you at that time? It might be able to push the button first and rebase later to trigger to queue by becoming "ready".

yannham:
@youknowone I might have been confused by the situation, which was that the CI didn't fail correctly, so I didn't want to merge a passing PR which would actually be wrong inadvertently. But I think you're right, in this case the original PR was failing, so I could just have used the merge queue :+1:

PR #753, opened 2023-04-21
Fix test failures

dburgener:
As mentioned in https://github.com/lalrpop/lalrpop/pull/752, the CI wasn't correctly propagating up failures, so "cargo test" has been failing in CI recently, despite a green checkmark.

PR #754, opened 2023-04-21
Fix tests

jannic:
This fixes two errors with the test cases.
First a compilation error:
```
error[E0308]: mismatched types
   --> lalrpop/src/main.rs:183:20
    |
183 |         parse_args(Arguments::from_vec(os_vec(args))).unwrap()
    |         ---------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&[&str]`, found `Arguments`
    |         |
    |         arguments to this function are incorrect
    |
note: function defined here
   --> lalrpop/src/main.rs:182:8
    |
182 |     fn parse_args(args: &[&str]) -> Args {
    |        ^^^^^^^^^^ -------------
```

And after I fixed that, I got a failing test case:
```
---- verify_lalrpop_generates_itself stdout ----
... more
thread 'verify_lalrpop_generates_itself' panicked at 'The snapshot does not match what lalrpop generates now.
Use ./snap.sh to generate a new snapshot of the lrgrammar', lalrpop-test/src/util/mod.rs:108:9


failures:
    verify_lalrpop_generates_itself
```

This two errors are fixed by the two commits in this pull request.

### DIFF
--- a/lalrpop/src/main.rs
+++ b/lalrpop/src/main.rs
@@ -179,7 +179,7 @@ mod test {
         vals.iter().map(|v| v.into()).collect()
     }
 
-    fn parse_args(args: &[&str]) -> Args {
+    fn parse_args_vec(args: &[&str]) -> Args {
         parse_args(Arguments::from_vec(os_vec(args))).unwrap()
     }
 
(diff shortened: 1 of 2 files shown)


jannic:
Oops, I didn't notice that @dburgener  fixed those errors in the mean time, in https://github.com/lalrpop/lalrpop/pull/753.
Both pull requests are essentially identical, so this one is no longer useful.

PR #755, opened 2023-04-21
Run cargo update before tests in CI

jannic:
Lalrpop is usually used as a dependency of some other crate. In that situation, the Cargo.lock of lalrpop would not be used.

But tests run in CI do consider the existing Cargo.lock file and will therefore run the tests against older versions of its dependencies.

If there are new breakages causes by more recent versions of lalrpop's dependencies, the CI script would not flag them.

Therefore, I propose to add a call to `cargo update` at the beginning of tools/ci.sh.

It might be worthwhile to do both: First test against the versions in Cargo.lock, then run `cargo update`, and then re-run the tests. However, that would double the CI runtime, so I didn't do that in this PR.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -2,6 +2,7 @@ export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
 export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
+cargo update
 cargo build -p lalrpop
 cargo test --all --all-features
 # Check the documentation examples separately so that the `lexer` feature specified in tests do not


Pat-Lafon:
I think it's because lalrpop can be installed and used as a binary. https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/main.rs

youknowone (approved):
Thank you! I think this is definitely better than before.
But just a question.
Rather than running before/after for every matrix, would it make sense to run it before `cargo update` only for stable rust?

dburgener:
> Rather than running before/after for every matrix, would it make sense to run it before cargo update only for stable rust?

IMHO we should take advantage of github actions matrixing to run all eight combos of rust toolchain version/cargo update or not.

If we're going to ship a Cargo.lock file, we care that it works.  If we're providing a library, we care that users can use lalrpop with the latest and greatest dependencies.  Since we're doing both, we should test both, and I'm not sure why we'd limit our testing to just certain toolchains.  There's not really a significant cost to the project of expanding CI coverage, and if CI really become too heavyweight in the future, it's easy to scale back.

dburgener:
> However, that would double the CI runtime, so I didn't do that in this PR.

Won't github run all the tests in parallel?  It will double the total work done, yes, but PR latency should remain roughly constant.

Pat-Lafon:
> > However, that would double the CI runtime, so I didn't do that in this PR.
> 
> Won't github run all the tests in parallel? It will double the total work done, yes, but PR latency should remain roughly constant.

I think this point is because much of the ci is just calling `tools/ci.sh` which if you made the modification there would cause it to run sequentially.

I'm pretty in favor of moving more outside of this one script so that there can be greater parallelism but I also don't to heavily rely on `tools/ci.sh`. IMO, that should be a lightweight check that builds/runs the pipeline on stable, and then the ci can run all the various combinations of checks via a matrix like you've suggested.

youknowone:
Though I merged #790 for the same purpose. I believe this PR derived those similar trials.
Thank you for pointing out good point to consider about.

PR #756, opened 2023-04-21
Add rust 1.60.0 clippy to CI

youknowone:


lalrpop/src/grammar/repr.rs
```diff
@@ -60,6 +60,7 @@ pub struct Grammar {
     pub module_attributes: Vec<String>,
 }
 
+#[allow(clippy::large_enum_variant)] // TODO: verify if this is justified
```
yannham, line 63: What were the details? Is `TypeRepr` the bigger variant? We might try to put it in a box then.

yannham (commented):
Sorry, I'm late to the party, but mostly LGTM

PR #758, opened 2023-04-21
Use #[default] for enum default

youknowone:
by @DaniPopes split from #743

Need decision from #742


youknowone:
@DaniPopes Thank you!

PR #762, opened 2023-04-22
Allow unused imports in nested test

dburgener:
https://github.com/lalrpop/lalrpop/issues/675

Imports in extern blocks are only needed in modules, but lalrpop can't currently differentiate that to avoid importing them globally.  Use the workaround suggested in the above linked issue to silence the warning when running test cases


youknowone (approved):
Thank you!

PR #770, opened 2023-04-24
release notes for 0.19.10

nikomatsakis:
Due to the regex changes, I've had requests for a new release.

I did my best to summarize the changes since 0.19.9 -- there are some nice things in there!

@lalrpop/maintainers, as far as I could tell, this won't require a new major version. The only thing that might've been breaking was the change in acronyms, but I think that is backwards compatible because we introduced type aliases. Do you agree?


yannham:
It's a small one, but I fear the camel case renaming changed `lalrpop_util::ParseError::UnrecognizedEOF` to `lalrpop_util::ParseError::UnrecognizedEof`, which is a public type.

nikomatsakis:
Good catch @yannham -- I'm thinking maybe I'll revert that PR on this branch.

nikomatsakis:
Oh, hmm, I'm not sure that I meant to merge those commits into *master* -- ah well, I'll do a follow-up PR I guess!

nikomatsakis:
Oh and I forgot to update the version number. I'm getting sloppy

PR #774, opened 2023-04-25
Range fix?

### LINKED ISSUE #773: Upgrading to version 0.19.10 breaks compilation
Hello,
We are using Lalrpop with version 0.19.9 and regex version 1.7.3
When we upgraded lalrpop to version 0.19.10 we are getting:
```
error: failed to run custom build command for `cel-parser v0.1.3-alpha.0 (/Users/omer/workspace/cel-rust/parser)`

Caused by:
  process didn't exit successfully: `/Users/omer/workspace/cel-rust/target/debug/build/cel-parser-062774afdad45860/build-script-build` (exit status: 101)
  --- stdout
  processing file `/Users/omer/workspace/cel-rust/parser/src/cel.lalrpop`

  --- stderr
  thread 'main' panicked at 'attempt to subtract with overflow', /Users/omer/.cargo/registry/src/github.com-1ecc6299db9ec823/lalrpop-0.19.10/src/lexer/dfa/overlap.rs:73:49
```

Any ideas what can cause that?

Pat-Lafon:
Working to fix #773. The null case is a little hard, so I think that needs special handling?

One issue I had is with the repeat test as with the new logic, I end up creating a single character range of just 'l'. I added a sort to make things nicer but it might be unnecessary. So this fix might add additional ranges

### DIFF
--- a/lalrpop/src/lexer/dfa/overlap.rs
+++ b/lalrpop/src/lexer/dfa/overlap.rs
@@ -20,6 +20,7 @@
 use crate::collections::Set;
 use crate::lexer::nfa::Test;
 use std::cmp;
+use std::ops::RangeInclusive;
 
 pub fn remove_overlap(ranges: &Set<Test>) -> Vec<Test> {
     // We will do this in the dumbest possible way to start. :)
@@ -48,6 +49,9 @@ fn add_range(range: Test, start_index: usize, disjoint_ranges: &mut Vec<Test>) {
         return;
     }
 
+    dbg!(&range);
+    dbg!(&disjoint_ranges);
+
     // Find first overlapping range in `disjoint_ranges`, if any.
     match disjoint_ranges[start_index..]
         .iter()
@@ -72,7 +76,15 @@ fn add_range(range: Test, start_index: usize, disjoint_ranges: &mut Vec<Test>) {
             // When working with inclusive ranges, we need to be sure to not double count
             // the meeting points of low-mid_range and mid-max_range.
             // So we adjust the end of the low_range and start of max_range as these elements are already included in the start of their corresponding next ranges.
-            let low_range = Test::new(min_min..=mid_min - 1);
+
+            let low_range = if mid_min == 0 {
+                // This is an edgecase where both ranges start at the null character
+                // In this case we don't want to create a range from 0 to -1
+                // Thus we create an empty range
+                Test::new(RangeInclusive::new(1, 0))
+            } else {
+                Test::new(min_min..=mid_min - 1)
+            };
             let mid_range = Test::new(mid_min..=mid_max);
             let max_range = Test::new(mid_max + 1..=max_max);
 
@@ -157,4 +169,13 @@ fn empty_range() {
         'a' ..= 'z',
     };
     assert_eq!(result, vec!['a'..='a', 'b'..='b', 'c'..='z']);
+}
+
+#[test]
+fn null() {
+    let result = test! {
+        '\0' ..= '\0',
+        '\0' ..= 'a',
+    };
+    assert_eq!(result, vec!['\0'..='\0', 1 as char..='a']);
 }
--- a/lalrpop/src/lexer/dfa/overlap.rs
+++ b/lalrpop/src/lexer/dfa/overlap.rs
@@ -38,6 +38,8 @@ pub fn remove_overlap(ranges: &Set<Test>) -> Vec<Test> {
     // prune them out.
     disjoint_ranges.retain(|r| !r.is_empty());
 
+    disjoint_ranges.sort();
+
     disjoint_ranges
 }
 
@@ -69,10 +71,10 @@ fn add_range(range: Test, start_index: usize, disjoint_ranges: &mut Vec<Test>) {
             let max_max = cmp::max(range.end(), overlapping_range.end());
             // When working with inclusive ranges, we need to be sure to not double count
             // the meeting points of low-mid_range and mid-max_range.
-            // So we subtract 1 from the end of the low_range and mid_range as these elements are already included in the start of their corresponding next ranges.
+            // So we adjust the end of the low_range and start of max_range as these elements are already included in the start of their corresponding next ranges.
             let low_range = Test::new(min_min..=mid_min - 1);
-            let mid_range = Test::new(mid_min..=mid_max - 1);
-            let max_range = Test::new(mid_max..=max_max);
+            let mid_range = Test::new(mid_min..=mid_max);
+            let max_range = Test::new(mid_max + 1..=max_max);
 
             assert!(low_range.is_disjoint(&mid_range));
             assert!(low_range.is_disjoint(&max_range));
@@ -118,7 +120,7 @@ fn alphabet() {
         'c' ..= 'l',
         '0' ..= '9',
     };
-    assert_eq!(result, vec!['0'..='9', 'a'..='b', 'c'..='k', 'l'..='z']);
+    assert_eq!(result, vec!['0'..='9', 'a'..='b', 'c'..='l', 'm'..='z']);
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


Pat-Lafon:
@dburgener Can you check this over?

dburgener:
This seems reasonable to me.  The key observation here I think is that mid_range really needs to be exactly the overlapping range.  So putting the adjustment on low and high preserves that and makes sense.

youknowone:
A question. In conclusion is this actually better than exclusive ranges?
To decide between reverting inclusive ranges or appending this patch.

Pat-Lafon:
My understanding is that the intent of `Test` was to be an inclusive range given the comments  https://github.com/lalrpop/lalrpop/blob/3b203c676cee618d20945d66ec65069372a942dd/lalrpop/src/lexer/nfa/mod.rs#L31 and examples https://github.com/lalrpop/lalrpop/blob/3b203c676cee618d20945d66ec65069372a942dd/lalrpop/src/lexer/dfa/overlap.rs#L5-L14.

However, this part of the code was introduced before `RangeInclusive` was added to the standard library(some of the commits are 7 years old) so this functionality had to be implemented manually alongside the original `Range` which is exclusive at one end.

The original motivation for the change was in clippy pointing out that because of this, some of the test cases like `0..9` are not actually covering the range of characters that one might expect(which would be `0..=9`).

In porting over to `RangeInclusive`, there are some nice benefits of having the type more strongly enforce/describe the invariant, maybe clarifying parts of the code that broke this invariant, reusing already existing standard library code instead of maintaining lalrpop's own version, and in having test cases over the full range of characters that is expected.

However, there are obvious drawbacks in doing this as we've seen, largely along the lines of "if it's not broken don't fix it", in that introducing off-by-one errors are easy to do, and that the tests we had were not sufficient to cover against breaking changes.

lalrpop/src/lexer/dfa/overlap.rs
```diff
@@ -69,10 +75,18 @@ fn add_range(range: Test, start_index: usize, disjoint_ranges: &mut Vec<Test>) {
...
+            // So we adjust the end of the low_range and start of max_range as these elements are already included in the start of their corresponding next ranges.
+
+            let low_range = if mid_min == 0 {
+                // This is an edgecase where both ranges start at the null character
+                // In this case we don't want to create a range from 0 to -1
+                // Thus we create an empty range
+                Test::new(RangeInclusive::new(1, 0))
+            } else {
+                Test::new(min_min..=mid_min - 1)
+            };
```
youknowone, line 87:
```suggestion
            // This is an edgecase where both ranges start at the null character
            // In this case we don't want to create a range from 0 to -1
            // Thus we create an empty range
            Test::new(min_min..=mid_min.checked_sub(1))
```
Is this possible? I don't know well about this code. So please check which version express the design intent better.

youknowone (commented):
The motivation sounds reasonable. Thank you for explanation!

lalrpop/src/lexer/dfa/overlap.rs
Pat-Lafon: `checked_sub` returns an option right? So you need to handle the `None` case. I guess one could write something like 

```rust
Test::new(mid_min.checked_sub(1).map_or_else(||RangeInclusive::new(1, 0), |v| min_min..=v))
```

I'm not sure if that's cleaner?

youknowone: oh, i am sorry. I was thinking of `saturated_sub` 😂

youknowone: https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub

Pat-Lafon: Currently, if mid_min would underflow, I am returning an empty range which will get ignored later. With saturating_sub, we would end up with a valid range of `\0`..=`\0`. I think this would mean having a duplicate value, probably `\0`..=`\0` twice.

eeeebbbbrrrr:
Hi, this doesn't seem to fix the problem.  It fixes the overflow error during parser generation, but now the generated parser has a number of compilation problems:

```
error[E0432]: unresolved import `self::__lalrpop_util::lexer`
  --> src/zql/parser.rs:30:31
   |
30 |     use self::__lalrpop_util::lexer::Token;
   |                               ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0432]: unresolved import `self::__lalrpop_util::lexer`
    --> src/zql/parser.rs:7221:31
     |
7221 |     use self::__lalrpop_util::lexer::Token;
     |                               ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0432]: unresolved import `self::__lalrpop_util::lexer`
     --> src/zql/parser.rs:13565:31
      |
13565 |     use self::__lalrpop_util::lexer::Token;
      |                               ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0432]: unresolved imports `self::__lalrpop_util::lexer`, `crate::zql::parser::Token`
     --> src/zql/ast.rs:13:5
      |
13    | use crate::zql::parser::Token;
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^
      |
     ::: src/zql/parser.rs:19968:38
      |
19968 | pub(crate) use self::__lalrpop_util::lexer::Token;
      |                                      ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0433]: failed to resolve: could not find `lexer` in `__lalrpop_util`
    --> src/zql/parser.rs:2351:34
     |
2351 |         builder: __lalrpop_util::lexer::MatcherBuilder,
     |                                  ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0433]: failed to resolve: could not find `lexer` in `__lalrpop_util`
    --> src/zql/parser.rs:8695:34
     |
8695 |         builder: __lalrpop_util::lexer::MatcherBuilder,
     |                                  ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0433]: failed to resolve: could not find `lexer` in `__lalrpop_util`
     --> src/zql/parser.rs:15046:34
      |
15046 |         builder: __lalrpop_util::lexer::MatcherBuilder,
      |                                  ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0433]: failed to resolve: could not find `lexer` in `__lalrpop_util`
     --> src/zql/parser.rs:19914:45
      |
19914 |     pub fn new_builder() -> __lalrpop_util::lexer::MatcherBuilder {
      |                                             ^^^^^ could not find `lexer` in `__lalrpop_util`

error[E0433]: failed to resolve: could not find `lexer` in `__lalrpop_util`
     --> src/zql/parser.rs:19965:25
      |
19965 |         __lalrpop_util::lexer::MatcherBuilder::new(__strs.iter().copied()).unwrap()
      |                         ^^^^^ could not find `lexer` in `__lalrpop_util`
```

I'm no lalrpop expert, and I doubt our parser would be easy to pull out as a standalone test, but this is the grammar in question:  https://github.com/zombodb/zombodb/blob/master/src/zql/parser.lalrpop

I'm happy to try and help however I can.

dburgener:
I think this may be a consequence of how you're doing the git build.

I took a separate repo I maintain that depends on lalrpop (which works fine with 0.19.10), and I pinned it to the 0.19.10 tag in git, and was able to reproduce similar errors to what you posted.

I think added:

```
lalrpop-util = { git = "https://github.com/lalrpop/lalrpop.git", tag = "0.19.10"}
```

(I didn't previously have any specific lalrpop-util dependency to remove).

And that made the problem go away resulting in a clean compile.  I'm not sure what's really going on, but somehow the git build seems to be preventing lalrpop-util from getting pulled in appropriately.

Can you try adding the lalrpop-util git dependency and see if that addresses it on your end as well?

If so, I think this isn't a problem with the 0.19.10 release or the fix in this issue, but something about the dependency tracking in doing manual git tracks in Cargo.toml, and presumably a yank of 0.19.10 and publish of 0.19.11 with this fix should resolve your problem and not hit the above error.

Issue #775, opened 2023-04-25
regex fails to build when pulled in via lalrpop-util 0.19.10


jonhoo:
```
$ cargo new foo
$ cd foo
$ cargo add lalrpop-util --features regex
    Updating crates.io index
      Adding lalrpop-util v0.19.10 to dependencies.
             Features:
             + regex
             + std
             - lexer
$ cargo check
  Downloaded regex v1.8.1
  Downloaded regex-syntax v0.7.1
  Downloaded 2 crates (588.1 KB) in 0.20s
    Checking regex-syntax v0.7.1
    Checking regex v1.8.1
error: `std` feature is currently required to build this crate
   --> /home/jongje/.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-1.8.1/src/lib.rs:648:1
    |
648 | compile_error!("`std` feature is currently required to build this crate");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0432]: unresolved import `crate::Error`
  --> /home/jongje/.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-1.8.1/src/compile.rs:16:5
   |
16 | use crate::Error;
   |     ^^^^^^^^^^^^ no `Error` in the root
   |
help: a similar name exists in the module
   |
16 | use crate::error;
   |            ~~~~~
help: consider importing one of these items instead
   |
16 | use core::error::Error;
   |     ~~~~~~~~~~~~~~~~~~
16 | use core::fmt::Error;
   |     ~~~~~~~~~~~~~~~~
16 | use crate::compile::fmt::Error;
   |     ~~~~~~~~~~~~~~~~~~~~~~~~~~
16 | use crate::compile::hir::Error;
   |     ~~~~~~~~~~~~~~~~~~~~~~~~~~
     and 7 other candidates

For more information about this error, try `rustc --explain E0432`.
error: could not compile `regex` due to 2 previous errors
```

This happens because the dependency on `regex` specifies `default-features = false`:
https://github.com/lalrpop/lalrpop/blob/22aff1bfb3723eceb5632ab86d8f59fdcfd40636/Cargo.toml#L20

But that also disables the `std` feature, which is currently _required_ to build `regex`:
https://github.com/rust-lang/regex/blob/1.8.1/src/lib.rs#L647-L648

[...]

PR #776, opened 2023-04-25
Fix issue #775

Pat-Lafon:
#775 has a good example of how to replicate this issue though I wasn't sure how to turn that into a test case.

I'm also trying to keep the feature sets for each dependency consistent. At the top level, we should also explicitly set the version of Regex to 1.8.

Personally, this aggressive use of `default_features = false` doesn't seem worth it but eh.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ edition = "2021"
 
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
-regex = { version = "1", default_features = false }
-regex-syntax = { version = "0.6", default_features = false }
+regex = { version = "1.8", default_features = false, features = ["std"] }
+regex-syntax = { version = "0.6", default_features = false, features = ["unicode"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -17,5 +17,5 @@ edition = "2021"
 
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
-regex = { version = "1", default_features = false }
-regex-syntax = { version = "0.6", default_features = false }
+regex = { version = "1.8", default_features = false, features = ["std"] }
```
youknowone, line 20:
```suggestion
regex = { version = "1", default_features = false, features = ["std"] }
```
If `1` is possible instead of `1.8`, the minor version seems to go with just `1`. We will be able to mark it `1.8` for lalrpop 1.0

Pat-Lafon: I guess? I'm a bit unclear but if that's needed then I'm not opposed to it.

In some sense, isn't part of the issues lalrpop had with regex is because the version was only declared as `version = "1"`? #750 Also the edition was updated recently which seems like a similar kind of breaking change?

yannham (commented):
LGMT, beside the 1.0/1.8 discussion. As mentioned in #775, a possible future discipline is to simply refrain from specifying local feature (at the level of a crate) of workspace dependencies.

Cargo.toml
Pat-Lafon: Seems like we are airing on the side of being conservative here which is fine.

There is a pedantic point to be made here which is that lalrpop doesn't actually compile with regex =1.0 because it uses the `std` feature that landed in 1.3(replacing a now deprecated `use_std` https://github.com/rust-lang/regex/blob/master/CHANGELOG.md#130-2019-09-03). I'm not sure if there is an automatic way to check this, but its something to be careful about.

nikomatsakis:
OK, I'm going to merge this.

nikomatsakis (approved):

Issue #757, opened 2023-04-21
Upgrade `regex-syntax` to 0.7


youknowone:
The original issue is reported from #750

#751 and https://github.com/lalrpop/lalrpop/pull/751#issuecomment-1518029847

PR #777, opened 2023-04-25
Upgrade regex-syntax to 0.7 #757

Pat-Lafon:
Taking a crack at upgrading regex-syntax.

Some of the non-trivial changes that could use a second look are with the new handling of Literal and all of the Anchor/Boundary stuff being converted into this Look(with the corresponding change to the error handling).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,4 +25,4 @@ rust-version = "1.64"
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
 regex = { version = "1.8", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.6", default_features = false }
+regex-syntax = { version = "0.7", default_features = false }
--- a/lalrpop/src/lexer/nfa/mod.rs
+++ b/lalrpop/src/lexer/nfa/mod.rs
@@ -291,21 +291,12 @@ impl Nfa {
                         (0, None) => self.star_expr(sub, accept, reject),
                         (1, None) => self.plus_expr(sub, accept, reject),
                         (_, Some(max)) if min == max => {
-                            let mut s = accept;
-                            for _ in 0..*max {
-                                s = self.expr(sub, s, reject)?;
-                            }
-                            Ok(s)
+                            (0..*max).try_fold(accept, |s, _| self.expr(sub, s, reject))
                         }
                         (_, Some(max)) => {
-                            let mut s = accept;
-                            for _ in *min..*max {
-                                s = self.optional_expr(sub, s, reject)?;
-                            }
-                            for _ in 0..*min {
-                                s = self.expr(sub, s, reject)?;
-                            }
-                            Ok(s)
+                            let s = (*min..*max)
+                                .try_fold(accept, |s, _| self.optional_expr(sub, s, reject))?;
+                            (0..*min).try_fold(s, |s, _| self.expr(sub, s, reject))
                         }
                         (_, None) => {
                             // +---min times----+
@@ -315,11 +306,9 @@ impl Nfa {
                             //          |      |
                             //          |      v
                             //          +-> [reject]
-                            let mut s = self.star_expr(sub, accept, reject)?;
-                            for _ in 0..*min {
-                                s = self.expr(sub, s, reject)?;
-                            }
-                            Ok(s)
+                            self.star_expr(sub, accept, reject).and_then(|s| {
+                                (0..*min).try_fold(s, |s, _| self.expr(sub, s, reject))
+                            })
                         }
                     }
                 }
--- a/lalrpop/src/lexer/nfa/mod.rs
+++ b/lalrpop/src/lexer/nfa/mod.rs
@@ -225,28 +225,12 @@ impl Nfa {
         match expr.kind() {
             HirKind::Empty => Ok(accept),
 
-            HirKind::Literal(Literal(l)) => {
-                /* // [s0] -otherwise-> [accept]
-                Literal::Unicode(c) => {
-                    let s0 = self.new_state(StateKind::Neither);
-                    self.push_edge(s0, Test::char(c), accept);
-                    self.push_edge(s0, Other, reject);
-                    Ok(s0)
-                } */
-                //// Bytes are not supported
-
-                if l.len() == 1 {
-                    let s0 = self.new_state(StateKind::Neither);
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -18,4 +18,4 @@ edition = "2021"
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
 regex = { version = "1", default_features = false }
```
youknowone, line 20:
```suggestion
regex = { version = "1.8", default_features = false }
```
regex < 1.8 seems not compatible with regex-syntax 0.7 If i correctly understood  #751

dburgener: That was not my understanding of #751.  Regex 1.8 requires regex_syntax 0.7, but regex_syntax 0.7 should be semver compatible with 0.6, so anything that needed 0.6 (such as regex 1.7) should in theory work with 0.7 as well.

If I take this PR and do:

```
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,5 +22,5 @@ rust-version = "1.64"
 
 [workspace.dependencies]
-regex = { version = "1", default_features = false }
+regex = { version = "=1.7", default_features = false }
 regex-syntax = { version = "0.7", default_features = false }
```

It builds and tests fine for me, which suggests that at least 1.7 works.  Probably we should determine our actual minimum regex version and set Cargo.toml to that, but that could be a separate work.

youknowone: Thank you for testing!
In that case, doesn't it result Cargo.lock includes both `regex-syntax 0.7` and `regex-syntax 0.6`?

youknowone: well, this is not a focus of this PR anyway.

dburgener: Yes, Cargo.lock includes 0.6 and 0.7 in that case, you're right.  I had my semver argument totally wrong above, since 0.6 and 0.7 are of course semver incompatible.  However, having both versions in Cargo.lock is more of a mild-negative affecting compile times than an actual break.

If a user for some reason only has regex 1.7 available, but has syntax 0.7, they should still be able to build lalrpop.  They'll just do extra building to get two versions of regex-syntax, with lalrpop using 0.7 and the older regex versions using 0.6.

Pat-Lafon: Yeah, it's possible to have two versions, but the more likely thing is that people will use the latest regex version(which lalrpop itself does) in which case this pr means there is only one version of regex-syntax.

Pat-Lafon: > It builds and tests fine for me, which suggests that at least 1.7 works. Probably we should determine our actual minimum regex version and set Cargo.toml to that, but that could be a separate work.

I've previously identified that as Regex 1.3, with #782 being able to catch if that increases on accident.

dburgener: I agree.  I tested regex versions this morning before I saw #782 and came to the same 1.3 conclusion.  Not sure if I'll have time to look over #782 fully today, but I have tested the 1.3 conclusion.

lalrpop/src/lexer/nfa/mod.rs
```diff
@@ -225,25 +222,29 @@ impl Nfa {
...
+                if l.len() == 1 {
+                    let s0 = self.new_state(StateKind::Neither);
+                    self.push_edge(s0, Test::byte(l[0]), accept);
+                    self.push_edge(s0, Other, reject);
+                    Ok(s0)
+                } else {
+                    let s0 = self.new_state(StateKind::Neither);
+                    self.push_edge(s0, Test::byte(*l.last().unwrap()), accept);
+                    self.push_edge(s0, Other, reject);
+                    self.expr(&Hir::literal(&l[..l.len() - 1]), s0, reject)
```
ithinuel, line 247:
```suggestion
                Ok(l.0.iter().rev().fold(accept, |accept, &b| {
                    let s0 = self.new_state(StateKind::Neither);
                    self.push_edge(s0, Test::char(c), accept);
                    self.push_edge(s0, Other, reject);
                    s0
                }))
```

lalrpop/src/normalize/token_check/mod.rs
```diff
@@ -357,9 +357,7 @@ fn construct(grammar: &mut Grammar, match_block: MatchBlock) -> NormResult<()> {
             let feature = match error {
                 NamedCaptures => r#"named captures (`(?P<foo>...)`)"#,
                 NonGreedy => r#""non-greedy" repetitions (`*?` or `+?`)"#,
-                WordBoundary => r#"word boundaries (`\b` or `\B`)"#,
-                LineBoundary => r#"line boundaries (`^` or `$`)"#,
-                TextBoundary => r#"text boundaries (`^` or `$`)"#,
+                LookAround => r#"All boundaries like `\b` or `\B` or `^` or `$`"#,
```
ithinuel, line 360: For consistentcy, this should start with a lower case
```suggestion
                LookAround => r#"all boundaries like `\b` or `\B` or `^` or `$`"#,
```

ithinuel (approved):
Text and Line boundaries are removed

lalrpop/src/lexer/nfa/mod.rs
Pat-Lafon: Hmm, is there any way to modify this suggestion in github? I think `Test::char(c)` should be `Test::byte(b)` or `Test::byte(*b)`

ithinuel: I edited the suggestion 👍

ithinuel (commented):
Not sure we want to go this way (readability isn't always the best) but here's a few suggestions using iterators.

lalrpop/src/lexer/nfa/mod.rs
```diff
@@ -280,65 +265,61 @@ impl Nfa {
...
+                        (_, Some(max)) => {
+                            let mut s = accept;
+                            for _ in *min..*max {
+                                s = self.optional_expr(sub, s, reject)?;
+                            }
+                            for _ in 0..*min {
+                                s = self.expr(sub, s, reject)?;
+                            }
+                            Ok(s)
+                        }
```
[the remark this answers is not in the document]
youknowone:
```suggestion
                        (_, Some(max)) => {
                            let s = (*min..*max)
                                .try_fold(accept, |s, _| self.optional_expr(sub, s, reject))?;
                            (0..*min).try_fold(s, |s, _| self.expr(sub, s, reject))
                        }
```

youknowone (approved):
Thank you!

PR #779, opened 2023-04-26
v0.19.9 no longer works for us due to `regex:1.8.1` changes

eeeebbbbrrrr:
Because of the problems with v0.19.10, I tried pinning our project (https://github.com/zombodb/zombodb) to 0.19.9, but that didn't work either, apparently because of differences in the new `regex` and `regex-syntax` crates.

Here's the trial-and-error I went through to get ZomboDB's parser generating/compiling again, based on lalrpop's 0.19.9 tag.

I'm just making this PR so y'all can see if this points to anything that lalrpop needs to change to better handle dependencies.  My suggestion would be to always pin to a particular regex version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


eeeebbbbrrrr:
I think #780 is a better approach.  Closing this.  Thanks!

youknowone:
I am sorry to miss this one 😅

Issue #778, opened 2023-04-26
Release a new version and yank last release


youknowone:
Edit: 0.19.11 is released without any of this changes. Those will be included in 0.20

bugs
- [x] #773
  - suggested fix is #774
- [x] #775 
  - suggested fix is #776 

0.20
- [x] #777

Rather than revert and re-revert, I'd like to suggest to create a `0.x` branch and backport only version fixes including #775. (or the speedup one if desired)

nikomatsakis:
@youknowone agreed about creating the branch. What a pain.

youknowone:
Another idea. Yank last one and release new one as 0.20
If we care about backward compatibilty, 0.19 expects only a single patch about cargo fix about regex

youknowone:
#777 is desired to be 0.20. Then regex problem will be clearly fixed.

youknowone:
yanked 0.19.10 and released 0.19.11

PR #780, opened 2023-04-26
0.19.11 release 

youknowone:
note that this PR is targetting `0.19` branch, not the master

One of suggestion for #778


youknowone:
oops, It must be 0.19.11

youknowone:
reset 0.19 branch to tag 0.19.9 and rebased patch onto it.

lalrpop-util/Cargo.toml
```diff
@@ -3,13 +3,13 @@ name = "lalrpop-util"
...
 license = "Apache-2.0 OR MIT"
-version = "0.19.9" # LALRPOP
+version = "0.19.11" # LALRPOP
 authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 workspace = ".."
 edition = "2018"
 
 [dependencies]
-regex = { version = "1", optional = true }
+regex = { version = "1", features = ["std", "unicode"], optional = true }
```
yannham, line 12: Should we apply the discipline proposed in https://github.com/lalrpop/lalrpop/issues/775#issuecomment-1523312554? That is, enable those features at the level of the workspace, and don't enable any feature at the crate-level to be sure that they won't get messed by a specific crate of the LALRPOP workspace changing configuration?

lalrpop/Cargo.toml
```diff
@@ -23,8 +23,8 @@ ena = { version = "0.14", default_features = false }
 is-terminal = "0.4.2"
 itertools = { version = "0.10", default_features = false, features = ["use_std"] }
 petgraph = { version = "0.6", default_features = false }
-regex = { version = "1", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.6", default_features = false }
+regex = { version = "1", default_features = false, features = ["std", "unicode"] }
```
yannham, line 26: As an aside, we seem to enable exactly the same features anyway

lalrpop-util/Cargo.toml
Pat-Lafon: Did you see #783? I was realizing that I'm not sure how these different levels of feature enabling intersect with lalrpop's features. Of course I'm all for turning everything on, but I just want to make sure that people who use lalrpop without the lexer still get whatever minimal setup they are looking for.

ithinuel:
I did some exploration & testing trying to fix lalrpop tests' builds when not building with `all-features` and as far as I (now) understand it, the point is that the extra features required on the regex and regex-syntax module are dependent on the grammar the user is building.

So enabling the whole "unicode" feature may not be where lalrpop want to aim (or at least not where it used to aim at).

#783 provides a set of fixes and an update of the CI script to capure similar feature related future breakage.

lalrpop-util/Cargo.toml
youknowone: I dropped this part and applied #783

PR #783, opened 2023-04-27
Fix building issues when not all features are enabled at once.

ithinuel:
The tests are currently run with all features enables which hides breakage when some are note defined.

This should help capture future issues.

**Note 1:** this PR removes a check in `normalize::token_check` that prevented most tests from running.
I am not sure about the extent of that change but all the tests now build & pass success-fully regardless of the enabled features.

**Note 2:** The subtlety of enabling the regex's and regex-syntax unicode feature for the specific needs of the grammar used, while totally sensible, is a bit unintuitive.
That's mostly because it is a transitive dependency and one does not expect it to break in such a way.
This might be worth stressing in the documentation.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,6 +19,8 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
           components: rustfmt
+      - name: Install cargo-hack
+        run: cargo install cargo-hack --locked
       - name: Run tests
         run: ./ci.sh
 
--- a/ci.sh
+++ b/ci.sh
@@ -1,7 +1,10 @@
+# Exis on error
+set -ex
+
 export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
-cargo build -p lalrpop
-cargo test --all --all-features
+cargo build --bin lalrpop --features pico-args
+cargo hack test --workspace --each-feature
 # Check the documentation examples separately so that the `lexer` feature specified in tests do not
 # leak into them
 cargo check -p calculator
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -30,10 +30,7 @@ term = { version = "0.7", default_features = false }
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }
 unicode-xid = { version = "0.2", default_features = false }
 
-# This dependency is only needed for binary builds, if you use LALRPOP as
-# library, disable it in your project by setting default_features = false.
 pico-args = { version = "0.4", default_features = false, optional = true }
-
 
 [dev-dependencies]
 regex = { version = "1", default-features = false, features = ["std", "unicode-case", "unicode-perl"] }
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -43,7 +43,7 @@ path = "../lalrpop-util"
 version = "0.19.9" # LALRPOP
 
 [features]
-default=["lexer", "pico-args"]
+default=["lexer"]
 
 # Feature used when developing LALRPOP. Tells the build script to use an existing lalrpop binary to
 # generate LALRPOPs own parser instead of using the saved parser.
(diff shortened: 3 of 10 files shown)


Pat-Lafon:
This is a good point. I think instead of the changes made to the ci here, we should instead adopt `cargo-hack` if possible. I'm not sure if we actually need to run `cargo test` for each feature set instead of just `cargo check` via `cargo hack`.

ithinuel:
There's actually 3 kind of issues addressed here:
- Lack of testing without features enabled: This is where `cargo-hack` could help I guess
- Broken features/dependencies:
  This effectively prevents the currently published version of this crate to be used with `no-default-features`.
  - `regex/std` and `std` when `lexer` is enabled
  - `pico-args` missing requirement for the binary (and maybe excessive requirement in `default` when building the lib)
  - missing `regex-syntax/unicode-case`
- and the testing requiring a `lexer` to be enabled despite the functionality being tested not necessarily requiring a lexer.
  The changes in this PR only disable the tests when `lexer` is not enabled. So even if it seems to build fine, the related functionality may still be broken is some surprising ways.

ithinuel:
Actually, I realized there's a simpler solution, I'll update this PR shortly.

.github/workflows/test.yaml
```diff
@@ -19,6 +19,8 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
           components: rustfmt
+      - name: Install cargo-hack
+        run: cargo install cargo-hack --locked
```
Pat-Lafon, line 23: Instead of running cargo-hack in the ./ci.sh file(Which requires people to have installed this and it also is going to take a little while to run), can this be put into it's own job so it runs in parallel to everything else?

ci.sh
```diff
@@ -1,7 +1,10 @@
+# Exis on error
+set -ex
+
 export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
-cargo build -p lalrpop
-cargo test --all --all-features
+cargo build --bin lalrpop --features pico-args
+cargo hack test --workspace --each-feature
```
Pat-Lafon, line 7: Do the tests actually need to be run, or is `cargo hack check` sufficient?

.github/workflows/test.yaml
ithinuel: Some (one?) of the tests require the `lalrpop` binary to be built.
If a user is looking to run `./ci.sh`, I find is safe to assume that they're trying to reproduce what the CI is actually doing.
In this case building various combinaison of features/no-default-features.

`cargo-hack` does not seem to allow parallel builds.

ci.sh
ithinuel: Cargo check merely verifies that it could possibly build.
Running the tests captures a lot more potential breakages (see the history of this PR to see all check that would've passed but actually fail when the tests is run).

lalrpop/Cargo.toml
```diff
@@ -36,14 +36,16 @@ pico-args = { version = "0.4", default_features = false, optional = true }
...
 [dependencies.lalrpop-util]
 path = "../lalrpop-util"
 version = "0.19.9" # LALRPOP
 
 [features]
-default=["lexer", "pico-args"]
+default=["lexer"]
 
 # Feature used when developing LALRPOP. Tells the build script to use an existing lalrpop binary to
 # generate LALRPOPs own parser instead of using the saved parser.
```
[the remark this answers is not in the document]
ithinuel: Which comment do you mean?

line 43 or 48 ?

ci.sh
Pat-Lafon: Yeah, I've tried this out and your right.

lalrpop/Cargo.toml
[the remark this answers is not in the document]
Pat-Lafon: Oh, I think I accidentally got the line wrong, I meant 
```rust 
# This dependency is only needed for binary builds, if you use LALRPOP as
# library, disable it in your project by setting default_features = false.
```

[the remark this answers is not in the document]
ithinuel: Done :)

youknowone:
Thank you so much. This is what we really need.

youknowone:
@ithinuel Could you also port this to the master branch?

youknowone:
cargo-hack prebuilt image seems also available
https://github.com/taiki-e/cargo-hack#on-github-actions

Issue #769, opened 2023-04-23
properly isolate `#[allow(clippy::..)]` to generated code


youknowone:
see https://github.com/lalrpop/lalrpop/pull/739#issuecomment-1518482164

dburgener:
I attempted to address this with https://github.com/lalrpop/lalrpop/pull/784, which followed the suggestion linked above.  The issue pointed out in that PR review, is that adding the layer of indirection on the non-action portions doesn't really accomplish much, since each `pub` parser is in its own module anyways.

It may be more desirable to add any needed clippy exceptions directly on the parse mods.

PR #784, opened 2023-04-28
Refactor the generated code into a module

dburgener:
This is so that we can set #[allow()] on the generated code freely, while still running things such as clippy on user supplied action code

See issue #769

### DIFF
--- a/lalrpop/src/parser/lrgrammar.rs
+++ b/lalrpop/src/parser/lrgrammar.rs
@@ -14,6 +14,9 @@ extern crate lalrpop_util as ___lalrpop_util;
 use self::___lalrpop_util::state_machine as ___state_machine;
 extern crate core;
 extern crate alloc;
+mod generated_code {
+#![allow(non_snake_case, non_camel_case_types, unused_mut, unused_variables, unused_imports, unused_parens, dead_code, clippy::all)]
+use super::*;
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 mod ___parse___Top {
@@ -26,14 +29,14 @@ use std::iter::once;
 use crate::tok::{self, Tok};
 use crate::util::strip;
 use lalrpop_util::ParseError;
-use super::super::Top;
+use super::super::super::Top;
 #[allow(unused_extern_crates)]
 extern crate lalrpop_util as ___lalrpop_util;
 #[allow(unused_imports)]
 use self::___lalrpop_util::state_machine as ___state_machine;
 extern crate core;
 extern crate alloc;
-use super::___ToTriple;
+use super::super::___ToTriple;
 #[allow(dead_code)]
 pub(crate) enum ___Symbol<'input>
  {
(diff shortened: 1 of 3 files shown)


dburgener:
You'll note that I didn't actually remove any #[allow(clippy::all)] on action code (eg reverting or partially reverting #739).  That's because some clippy warnings on generated code are still worthless.  For example, after reverting #739 with this, I get clippy warnings like:

```
warning: writing `&mut Vec` instead of `&mut [_]` involves a new object where a slice will do
    --> /home/daniel/lalrpop/target/debug/build/lalrpop-test-7b5a786080b0055d/out/error_recovery_span.rs:1100:13
     |
1100 |     errors: &'e mut Vec<(usize, usize)>,
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `&'e mut [(usize, usize)]`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
```

Clippy is complaining because in the particular generated action function, errors could be a slice.  But in the overall context, errors does need to be a Vec.  I doubt lalrpop wants to pick and chose on each action function to separate clippy.  Maybe we want to silence certain clippy warnings on the action code, but it would require more case by case checking.

In my opinion, this change has independent value, but I'd be quite happy to hold it for a while to try to clean up some of the special casing on clippy in the tests first.

dburgener:
One difference between this and the suggestion in #739, is that I didn't put the action code in a mod, just the generated code.  The reason for that is that in order to put the action code in a mod, we'd need to rewrite all the supers, which is more difficult for action code than generated code.  See https://github.com/lalrpop/lalrpop/blob/80dd78a7b677d32ba150597964dd7aa76b3c21d6/lalrpop/src/build/mod.rs#L354-L358

lalrpop/src/parser/lrgrammar.rs
```diff
@@ -14,6 +14,9 @@ extern crate lalrpop_util as ___lalrpop_util;
 use self::___lalrpop_util::state_machine as ___state_machine;
 extern crate core;
 extern crate alloc;
+mod generated_code {
+#![allow(non_snake_case, non_camel_case_types, unused_mut, unused_variables, unused_imports, unused_parens, dead_code, clippy::all)]
+use super::*;
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 mod ___parse___Top {
```
youknowone, line 22: I am not sure about this. Almost the same thing is attached on parse module, which is practically the entire contents of the `generated_code`. It seems a depth is added but almost nothing more covered.

dburgener: Yeah, I think you're right about that.  Trying without lines 18 and 19 seems to be working fine for me.  I'll push an update to remove those.

youknowone: Now the only contents of `generated_code` is `mod ___parse___Top` 😅

dburgener: Oh, I see your point.  Yeah, that does seem like an issue.

Of course, in a parser with more than one `pub` nonterminal, there would be multiple modules inside `generated_code`.  But even then, I guess the extra layer of `generated_code` may not add much value.

This seems to somewhat invalidate https://github.com/lalrpop/lalrpop/pull/739#issuecomment-1518482164.

Maybe a better solution would be to add `#[allow(clippy::all)]` directly on each parse mod?

youknowone: Yes, that makes sense.

dburgener: Okay, I'll abandon this PR and add a note in #769.  If I have time this weekend, I'll try the `#[allow(clippy::all)]` on the parse modules approach.

PR #785, opened 2023-04-28
clean up #[allow] and generated rustfmt::skip

youknowone:


lalrpop/src/lr1/codegen/base.rs
```diff
@@ -158,16 +158,13 @@ impl<'codegen, 'grammar, W: Write, C> CodeGenerator<'codegen, 'grammar, W, C> {
...
-        rust!(self.out, "#[cfg_attr(rustfmt, rustfmt_skip)]");
-        rust!(self.out, "mod {}parse{} {{", self.prefix, self.start_symbol);
-
-        // these stylistic lints are annoying for the generated code,
-        // which doesn't follow conventions:
+        rust!(self.out, "#[rustfmt::skip]");
         rust!(
             self.out,
-            "#![allow(non_snake_case, non_camel_case_types, unused_mut, unused_variables, \
+            "#[allow(non_snake_case, non_camel_case_types, unused_mut, unused_variables, \
```
youknowone, line 164: Not much reason to move this up, but keeping `#[allow]` and `#[rustfmt::skip]` at the same level will make sense

lalrpop-test/src/lib.rs
```diff
@@ -20,7 +20,6 @@ macro_rules! lalrpop_mod_test {
     ($(#[$attr:meta])* $vis:vis $modname:ident) => {
         lalrpop_mod!(
             #[allow(clippy::ptr_arg)]
```
Pat-Lafon, line 22: In doc/parcal/lalrpop/src/main.rs, both clippy lints(vec_box and ptr_arg) were removed. Can the same be done here?

lalrpop/src/lr1/codegen/base.rs
Pat-Lafon: Somewhere else I had suggested using `#[automatically_derived]`, would that work here?

youknowone: It only works for `impl`. This is `mod`, so it doesn't work.

Pat-Lafon: Shucks!

PR #786, opened 2023-04-28
Introdude `unicode` and fix various test cases when not all features are enabled.

ithinuel:
This is a "frontport" of #783.

This sligtly deviates from the afore-mention PR to adapt to the changes on `master` and hopfully for the better.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
         "lalrpop-util",
         "lalrpop",
 ]
+resolver = "2"
 
 [workspace.package]
 repository = "https://github.com/lalrpop/lalrpop"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,4 +25,4 @@ rust-version = "1.64"
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
 regex = { version = "1.8", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.6", default_features = false, features = ["unicode"] }
+regex-syntax = { version = "0.6", default_features = false }
--- a/doc/calculator/Cargo.toml
+++ b/doc/calculator/Cargo.toml
@@ -6,14 +6,8 @@ build = "build.rs" # <-- We added this and everything after!
 workspace = "../.."
 edition = "2021"
 
-[build-dependencies.lalrpop]
-version = "0.19.10"
-path = "../../lalrpop"
-features = ["lexer"]
+[build-dependencies]
+lalrpop = { version = "0.19.10", path = "../../lalrpop" }
 
 [dependencies]
-regex = "1"
-
-[dependencies.lalrpop-util]
-version = "0.19.10"
-path = "../../lalrpop-util"
+lalrpop-util = { version = "0.19.10", path = "../../lalrpop-util", features = ["lexer", "unicode"] }
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -12,14 +12,8 @@ the following lines to your `Cargo.toml`:
 build = "build.rs" # LALRPOP preprocessing
 
 # The generated code depends on lalrpop-util.
-#
-# The generated tokenizer depends on the regex crate.
-#
-# (If you write your own tokenizer, or already have the regex
-# crate, you can skip this dependency.)
 [dependencies]
 lalrpop-util = "0.19.10"
-regex = "1"
 
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -33,26 +33,21 @@ term = { version = "0.7", default_features = false }
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }
 unicode-xid = { version = "0.2", default_features = false }
 
-# This dependency is only needed for binary builds, if you use LALRPOP as
-# library, disable it in your project by setting default_features = false.
 pico-args = { version = "0.5", default_features = false, optional = true }
 
+lalrpop-util = { path = "../lalrpop-util", version = "0.19.10" }
 
 [dev-dependencies]
 rand = "0.8"
 
-[dependencies.lalrpop-util]
-path = "../lalrpop-util"
-version = "0.19.10" # LALRPOP
-
 [features]
-default=["lexer", "pico-args"]
-
-# Feature used when developing LALRPOP. Tells the build script to use an existing lalrpop binary to
-# generate LALRPOPs own parser instead of using the saved parser.
-test = []
-
+default=["lexer", "unicode"]
+unicode = ["regex/unicode", "regex-syntax/unicode"]
 lexer = ["lalrpop-util/lexer"]
 
 [package.metadata.docs.rs]
 features = ["lexer"]
+
+[[bin]]
+name = "lalrpop"
+required-features = ["pico-args"]
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -33,6 +33,8 @@ term = { version = "0.7", default_features = false }
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }
 unicode-xid = { version = "0.2", default_features = false }
 
+# This dependency is only needed for binary builds, if you use LALRPOP as
+# library, disable it in your project by setting default_features = false.
 pico-args = { version = "0.5", default_features = false, optional = true }
 
 lalrpop-util = { path = "../lalrpop-util", version = "0.19.10" }
@@ -41,7 +43,7 @@ lalrpop-util = { path = "../lalrpop-util", version = "0.19.10" }
 rand = "0.8"
 
 [features]
-default=["lexer", "unicode"]
+default=["lexer", "unicode", "pico-args"]
 unicode = ["regex/unicode", "regex-syntax/unicode"]
 lexer = ["lalrpop-util/lexer"]
 
--- a/lalrpop/src/normalize/token_check/test.rs
+++ b/lalrpop/src/normalize/token_check/test.rs
@@ -111,6 +111,8 @@ fn regex_literals() {
 
 /// Basic test for match mappings.
 #[test]
+// This test requires regex's unicode case support
+#[cfg_attr(not(feature = "unicode"), ignore)]
 fn match_mappings() {
     check_intern_token(
         r#"grammar; match { r"(?i)begin" => "BEGIN" } else { "abc" => ALPHA } X = "BEGIN" ALPHA;"#,
@@ -125,6 +127,8 @@ fn match_mappings() {
 /// Match mappings, exercising precedence. Here the ID regex *would*
 /// be ambiguous with the begin regex.
 #[test]
+// This test requires regex's unicode case support
+#[cfg_attr(not(feature = "unicode"), ignore)]
 fn match_precedence() {
     check_intern_token(
         r#"grammar; match { r"(?i)begin" => "BEGIN" } else { r"\w+" => ID } X = ();"#,
@@ -158,12 +162,16 @@ fn invalid_match_regex_literal() {
 
 /// Test that, with a catch-all, the previous two examples work.
 #[test]
+// This test requires regex's unicode case support
+#[cfg_attr(not(feature = "unicode"), ignore)]
 fn match_catch_all() {
     let grammar = r#"grammar; match { r"(?i)begin" => "BEGIN", _ } X = { "foo", r"foo" };"#;
     assert!(validate_grammar(grammar).is_ok())
 }
 
 #[test]
+// This test requires regex's unicode case support
+#[cfg_attr(not(feature = "unicode"), ignore)]
 fn complex_match() {
     let grammar = r##"
         grammar;
@@ -182,6 +190,8 @@ fn complex_match() {
 /// Test that overlapping regular expressions are still forbidden within one level
 /// of a match declaration.
 #[test]
+// This test requires regex's unicode case support
+#[cfg_attr(not(feature = "unicode"), ignore)]
 fn ambiguity_within_match() {
     check_err(
         r##"ambiguity detected between the terminal `r#"b"#` and the terminal `r#"\(\?i\)b"#`"##,
@@ -196,8 +206,8 @@ fn ambiguity_within_match() {
 #[test]
 fn same_literal_twice() {
     check_err(
-        r##"multiple match entries for `r#"\(\?i\)b"#`"##,
-        r#"grammar; match { r"(?i)b" => "B" } else { r"(?i)b" => "b" }"#,
-        r#"                                          ~~~~~~~~~~~~~~~~ "#,
+        r##"multiple match entries for `r#"\[bB\]"#`"##,
+        r#"grammar; match { r"[bB]" => "B" } else { r"[bB]" => "b" }"#,
+        r#"                                         ~~~~~~~~~~~~~~~ "#,
     );
 }
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -4,8 +4,11 @@ export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
 export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
-cargo build -p lalrpop
-cargo test --all --all-features
+cargo build --bin lalrpop --features pico-args
+# build with minimal amount of features to make sure dependencies can build too.
+cargo hack build --workspace --feature-powerset --optional-deps
+# test with minimal amount of features plus a few extra on regex/regex-syntax
+cargo hack test --workspace --feature-powerset --optional-deps
 # Check the documentation examples separately so that the `lexer` feature specified in tests do not
 # leak into them
 cargo check -p calculator
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,6 +29,8 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           components: rustfmt
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
       - name: Run tests
         run: ./tools/ci.sh
   lint:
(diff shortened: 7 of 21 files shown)


doc/calculator/Cargo.toml
```diff
@@ -6,14 +6,17 @@ build = "build.rs" # <-- We added this and everything after!
...
+[dev-dependencies]
+regex = { workspace = true, features = ["unicode-perl"] }
+
 [build-dependencies.lalrpop]
 version = "0.19.10"
 path = "../../lalrpop"
 features = ["lexer"]
 
-[dependencies]
-regex = "1"
```
youknowone, line 15: This is part of document example code. Not using workspace will helps document readers to follow the code out of lalrpop workspace.

ithinuel: deps updated not to use the workspace deps in doc examples 👍

tools/ci.sh
```diff
@@ -4,8 +4,11 @@ export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
 export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
-cargo build -p lalrpop
-cargo test --all --all-features
+cargo build --bin lalrpop --features pico-args
+# build with minimal amount of features to make sure dependencies can build too.
+cargo hack build --workspace --each-feature --optional-deps
+# test with minimal amount of features plus a few extra on regex/regex-syntax
+cargo hack test --workspace --each-feature --optional-deps
```
Pat-Lafon, line 11: Just curious if there is a difference here with `--each-feature` versus using `--feature-powerset`?

tools/ci.sh
```diff
@@ -4,8 +4,11 @@ export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
 export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
-cargo build -p lalrpop
-cargo test --all --all-features
+cargo build --bin lalrpop --features pico-args
+# build with minimal amount of features to make sure dependencies can build too.
+cargo hack build --workspace --each-feature --optional-deps
```
Pat-Lafon, line 9: Since to run the tests you need to have compiled the corresponding project, is this redundant?(Maybe we only need `cargo hack test`)

ithinuel: As far as I am aware, there is no guaranty that the binary (`cargo build -p lalrpop --features pico-args`) will be built before the builds that require it (`cargo build --no-default-features --features test` is one of them)

Pat-Lafon: I don't think I'm confident enough on the first point to speculate beyond just saying I removed this line locally and it still worked for me. But you probably understand this point better than I do.

For the second point, at least in running the script locally and in looking at the logs, `cargo hack test` does seem to recompile on some of the sets of features. In general, I think it's a fair concern that the ci goes from 6 min https://github.com/lalrpop/lalrpop/actions/runs/4833729044/jobs/8614076121 to 12 min https://github.com/lalrpop/lalrpop/actions/runs/4834066012/jobs/8614838843?pr=786 . I'm happy to try to mitigate this in a different pr if that's easier.

doc/calculator/Cargo.toml
ithinuel: Makes me wonder why is regex required at that point in the tutorial?

IIUC, it only required in the `calculator` binary because it uses `\s` which requires the `unicode-perl` feature to be enabled.

tools/ci.sh
ithinuel: I did the change locally.
I am looking at documentation updates now.

doc/calculator/Cargo.toml
ithinuel: @youknowone I updated the doc (in the quick start guide) to raise attention on the possible need for the extra regex's unicode features.

I haven't figured yet what/when are the feature required on regex-syntax or regex.

doc/src/quick_start_guide.md
```diff
@@ -12,21 +12,18 @@ the following lines to your `Cargo.toml`:
...
-regex = "1"
 
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
 lalrpop = "0.19.10"
 # If you are supplying your own external lexer you can disable default features so that the
 # built-in lexer feature is not included
 # lalrpop = { version = "0.19.1", default-features = false }
+
+# If you are using some more advanced regural expressions in your grammar, you will most likely need
```
Pat-Lafon, line 25: "regural" -> "regular"

tools/ci.sh
ithinuel: The doubling in CI time is because instead of testing 1 set of features (all enabled) it's building a lot more combinasion (even more when powerset is enabled).
But IMHO, exhaustiveness is the point of having CI as a strong PR gate.
CI may benefit from running build & tests in parallel pipelines. 

I can make something similar to this: https://github.com/rp-rs/rp-hal/pull/588
The downside being that it's not using a shell script, so it's sell trivial to run the same tests than the pipeline locally before pushing.

doc/calculator/Cargo.toml
```diff
@@ -6,14 +6,12 @@ build = "build.rs" # <-- We added this and everything after!
...
 
-[dependencies.lalrpop-util]
-version = "0.19.10"
-path = "../../lalrpop-util"
+[build-dependencies]
+lalrpop = { version = "0.19.10", path = "../../lalrpop", features = ["lexer"] }
+regex-syntax = { version = "0.6", default-features = false, features = ["unicode-perl"] }
+
+[dev-dependencies]
+regex = { version = "1.8", default-features = false, features = ["unicode-perl"] }
```
youknowone, line 17:
```suggestion
regex = "1.8"
```
Rather than making tutorial so heavy, how about just like this? The default feature is `unicode`, so it will be ok.

ithinuel: Since not having `unicode` enabled as a blanket feature (let alone default features), wouldn't that be a good opportunity to document it?

Cargo.toml
```diff
@@ -24,4 +24,4 @@ rust-version = "1.64"
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
 regex = { version = "1.8", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.6", default_features = false, features = ["unicode"] }
```
dburgener, line 27: Until we get https://github.com/lalrpop/lalrpop/issues/789, I think we have to keep specifying unicode on regex-syntax.

In particular, enabling the feature for our local testing, but not for users is a recipe to mask issues around feature underspecification.

The problem here is that users can specify a regex with unicode characters without directly depending on regex-syntax.  Lalrpop then hands those characters to regex-syntax, resulting in errors.  Users should not need to know that lalrpop depends on regex-syntax, and add a dependency on it to enable its features.

The real answer is #789, but in the interim, I think it's broken to pass on valid user input to regex-syntax and then expect users to look into lalrpop internals and set a feature on a dependency they wouldn't otherwise include at all.

In particular this causes problems related to versioning.  Lalrpop users needing unicode support are currently working around by setting the unicode feature on regex-syntax 0.6.  When lalrpop updates to regex-syntax 0.7, those users 1. Will have an unnecessary 0.6 dependency and 2. Will be once again underspecified on regex-syntax 0.7.

ithinuel: > In particular, enabling the feature for our local testing, but not for users is a recipe to mask issues around feature underspecification.

Yes. Tests should not require optional features (unless they are testing things related to that feature).
The issue with testing is that they nearly all require such unicode feature. 

I personally don't like having to enable features for a transitive dependency.

I can introduce this feature here, that's not really complicated.
Some of the tests need to be disabled when unicode is not set because they make use of `(?i)` an this eventually gets passed down to `regex` (not all tests using `(?i)` require the regex features).

PS: As I was doing this, I identified another issue (compounding with the previous) that made features even harder to identify.
The workspace does not set `resolver = "2"` so the features used in `build-dependencies` were leaking in the `dependencies`.
Expect a new commit soon™ covering all this.

doc/calculator/Cargo.toml
nikomatsakis: I'm honestly rather confused what all this breakage is about, so it's hard for me to have an informed opinion.

dburgener: It's a confusing situation, and I'm not sure I've seen a clear, concise write-up.  Let me try to provide one and see if it can clarify things for you.

When a user writes a regex in their .lalrpop file, lalrpop uses the regex-syntax crate to generate a representation of that regex (https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/lexer/re/mod.rs).  If that representation is generatable without unicode support, this always works.  However, if the regex requires unicode support, then regex-syntax *must* have a unicode feature enabled (I'm still a little vague on unicode vs unicode-perl.  I think just the unicode-perl feature will work in most situations).

A common situation here is a regex like `r"\s*"`, which matches both ASCII and unicode whitespace.

If regex-syntax is compiled without unicode support, it reports an error at runtime.

Lalrpop does not currently (on master) specify that the regex-syntax unicode feature is required.  However, due to feature unification it can be required in a number of situations.  Specifically, the regex crate depends on regex-syntax, and with its default features it includes regex-syntax unicode support.

The problem was first noticed when regex updated to version 1.8.  Version 1.8 of regex moved its regex-syntax dependency to 0.7, while lalrpop continued to depend on 0.6.  This means that anyone getting unicode support indirectly through regex (lalrpop and its users) *did* get unicode support on regex-syntax 0.7, but did *not* on regex-syntax 0.6, since regex no longer used 0.6.  Lalrpop however still used 0.6, and so built regex-syntax without unicode support, and therefore broke lalrpop users relying on it.  This was masked in lalrpop's testing by the fact that lalrpop CI runs with a Cargo.lock in place, so was testing against an older version of regex, which had the same regex-syntax dependency as lalrpop.

Fundamentally, lalrpop users shouldn't need to know about or worry about features on lalrpop dependencies.  If lalrpop updates to a new regex-syntax version, or switches to a different regex crate, that should be transparent to lalrpop users.  However, the current state (on master) is that whether regex-syntax is built with unicode support depends on a bunch of transitive dependency resolution in the specific crate.

The right answer here is definitely what is suggested in #789.  Lalrpop should expose a configuration knob to users who can opt in or out to unicode support with lalrpop directly, rather than adopting brittle workarounds by setting a feature on regex-syntax, which they may not depend on otherwise.  Lalrpop then owns the responsibility for turning around and setting regex-syntax's feature flags around unicode independently.

Separately, lalrpop should also improve future robustness by merging some variation on this PR here and some variation on #755, so that features are accurately and fully specified and tested.

ithinuel:
I rewrote the history of that PR to remove obsolete changes and give a bit more context about each changeset.

lalrpop/Cargo.toml
```diff
@@ -33,26 +33,21 @@ term = { version = "0.7", default_features = false }
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }
 unicode-xid = { version = "0.2", default_features = false }
 
-# This dependency is only needed for binary builds, if you use LALRPOP as
-# library, disable it in your project by setting default_features = false.
```
youknowone, line 37: As I remember, this is related to `cargo install`. By removing this feature from default feature, `cargo install lalrpop` will fail and have to be `cargo install lalrpop --features pico-args`. 
https://github.com/rust-lang/cargo/issues/10409

I'd like to keep this feature as default unless cargo add a feature or we make a new crate `lalrpop-cli`

lalrpop/src/normalize/token_check/test.rs
```diff
@@ -196,8 +206,8 @@ fn ambiguity_within_match() {
 #[test]
 fn same_literal_twice() {
     check_err(
-        r##"multiple match entries for `r#"\(\?i\)b"#`"##,
-        r#"grammar; match { r"(?i)b" => "B" } else { r"(?i)b" => "b" }"#,
-        r#"                                          ~~~~~~~~~~~~~~~~ "#,
+        r##"multiple match entries for `r#"\[bB\]"#`"##,
```
youknowone, line 209: 👍

Cargo.toml
youknowone: Is this resolved by adding `unicode` feature in this PR?

youknowone (commented):
Thank you so much. I have a concern about removing `pico-args` from default, but other stuffs are looking so great and a lot better organized then before.

One more question. Do we need `unicode` as a default feature only for lalrpop but not for lalrpop-util?

youknowone:
@dburgener @Pat-Lafon @jannic would you review this PR?

lalrpop/Cargo.toml
ithinuel: It will "succeed" (exit with 0) without building/installing the lalrpop binary and display:

```
warning: none of the package's binaries are available for install using the selected features
  bin "lalrpop" requires the features: `pico-args`
Consider enabling some of the needed features by passing, e.g., `--features="pico-args"`
```

Cargo.toml
ithinuel: I think so. `unicode` on `regex-syntax` gets enabled as required by enabling `unicode` on `lalrpop`.

lalrpop/Cargo.toml
Pat-Lafon: One day maybe cargo will get https://github.com/rust-lang/rfcs/pull/3374

lalrpop/Cargo.toml
```diff
@@ -33,26 +33,21 @@ term = { version = "0.7", default_features = false }
...
 
 [dev-dependencies]
 rand = "0.8"
 
-[dependencies.lalrpop-util]
-path = "../lalrpop-util"
-version = "0.19.10" # LALRPOP
-
 [features]
-default=["lexer", "pico-args"]
```
dburgener, line 49: Pretty sure this is a non-issue, since this PR is targeting 0.20, but just to highlight anyways, this is a semver breaking change, so it should definitely only go in a breaking update, and should probably get highlighted in a changelog.

I agree with the logic on removing the default feature, for the reasons @ithinuel expressed elsewhere in this PR, I just want to make sure that we highlight and communicate it properly on the off-chance that someone is affected.  One group that this might hit is anyone doing `cargo install lalrpop` in an automated script somewhere.  Presumably when they update from 0.19 to 0.20 they'll retest their script and update the command line, but it never hurts to be cautious and be able to say "look at the clear changelog" if anyone is impacted by the change to defaults.

youknowone: Since semver regards 0.x minor version changes as breaking change, I don't worry about it.
The concern is the standard user group in rust community don't expect `cargo install` has complicated requirements with more arguments.
It might be acceptable because lalrpop is more like professional dev tool rather than end user application. I personally will avoid this kind of user experience for any end-user applications.

dburgener: Yes, I think it's fine to put this in 0.20, since it's a semver change from 0.19.  My only point is that it's something worth highlighting in the 0.20 release notes.

dburgener:
This looks great to me.  Seems like a lot of great improvements and a clean solution to the regex unicode issues.

I also tested out building a simple rust binary that includes only lalrpop as a dependency, and builds a simple parser that requires unicode support.  I verified that I get build errors when disabling the unicode feature and build successfully when enabling it, so that all works as expected.

I didn't look closely at the Span related changes in commit 3, as I don't have any experience with that part of the lalrpop code to evaluate those, but I did read through everything else, and IMHO it seems ready for merge.

dburgener (approved):

youknowone (approved):
I reverted only the pico-args part. Other parts are technical and based on confirmed decisions.
But pico-args part is breaking change with pros and cons. I will leave it to @nikomatsakis 's side.
Could you please open another PR for that?

lalrpop/Cargo.toml
```diff
@@ -37,22 +37,19 @@ unicode-xid = { version = "0.2", default_features = false }
...
-
-# Feature used when developing LALRPOP. Tells the build script to use an existing lalrpop binary to
-# generate LALRPOPs own parser instead of using the saved parser.
-test = []
-
+default=["lexer", "unicode", "pico-args"]
```
jannic, line 46: Is it intentional to enable those by default? From other comments I got the impression that they should be disabled in 0.20 even though it's a breaking change.

ithinuel: `lalrpop` is used during build so it's nor really an issue for it to have more features by default.

tools/ci.sh
```diff
@@ -4,8 +4,11 @@ export RUST_BACKTRACE=1
 export CARGO_INCREMENTAL=0
 export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
-cargo build -p lalrpop
-cargo test --all --all-features
+cargo build --bin lalrpop --features pico-args
+# build with minimal amount of features to make sure dependencies can build too.
```
jannic, line 8: This comment doesn't exactly match the command, as feature-powerset tests several combinations of features, not only the minimal amount of features.
Btw, does it even work as expected to first build multiple combinations and then test multiple combinations?

lalrpop/Cargo.toml
jannic: Ok. I'm currently reading the changes on my phone, so I might be missing the big picture, literally and figuratively.

tools/ci.sh
ithinuel: you cane see the various combinasion of features for build and tests over here: https://github.com/lalrpop/lalrpop/actions/runs/4843403960/jobs/8630997943#step:5:142

jannic: Yes, that's exactly what I expected. But it's not "the minimal amount of features" - it's basically every possible combination of features.
But of course that's only a minor nit pick.
And while I'm nit-picking: What additional value does the `cargo hack build` provide over the later `cargo hack test`? Doesn't `cargo test` implicitly build the package first?

ithinuel: Ah right, when it was introduced, there was extra `[dev-dependencies]` to enable specific regex-syntax features during tests. So build & test weren't the same.
They should be the same now indeed.
Oh and initially it was using `--each-feature` rather than `--feature-powerset` hence the "minimal amount".
But yeah, this comment is a bit off now.

PR #787, opened 2023-04-28
Enable 'unicode' feature on regex-syntax

### LINKED ISSUE #750: Regex version 1.8.0 breaks build

dburgener:
This fixes a compatibility issue with regex 1.8

Closes #750

### DIFF
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -24,7 +24,7 @@ is-terminal = "0.4.2"
 itertools = { version = "0.10", default_features = false, features = ["use_std"] }
 petgraph = { version = "0.6", default_features = false }
 regex = { version = "1", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.6", default_features = false }
+regex-syntax = { version = "0.6", default_features = false, features = ["unicode"] }
 string_cache = { version = "0.8", default_features = false }
 term = { version = "0.7", default_features = false }
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }


dburgener:
Because 0.19.10 was accidentally made on the master branch, this fix never made it to the 0.19 branch, and it appears to me pulling lalrpop into another project that 0.19.11 regressed #750.

dburgener:
Note that this PR is targeting the 0.19 branch.

ithinuel:
This has been addressed by #783 and is being addressed (and somewhat documented) in #786.

Considering that all dependencies are defined with `default-features = false`, adding the blanket `unicode` feature on regex to enabled them all is self-contradictory.

ithinuel:
Oh, It's been merged already :/

dburgener:
> This has been addressed by https://github.com/lalrpop/lalrpop/pull/783 and is being addressed (and somewhat documented) in https://github.com/lalrpop/lalrpop/pull/786.

None of those are on the 0.19 branch.  This is the same minimal fix that was in 0.19.10, but was accidentally dropped for 0.19.11

ithinuel:
#783 is on 0.19 branch.

youknowone:
@ithinuel the dependency seems came from user code, but I'd like to keep 0.19 not broken for previous versions' users. I will drop it for 0.20

ithinuel:
Sorry for breaking it again in #783 😅

PR #790, opened 2023-04-29
Add a matrix to run all tests both with and without cargo update

dburgener:
An alternative option to #755.  This pulls cargo update out of ci.sh into the test.yaml, and uses a matrix strategy to run all tests with both the existing Cargo.lock and cargo update run, for each toolchain version.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,9 @@ jobs:
           - nightly
         cargo-update:
           - true
-          - false
+        include:
+          - rust: stable
+            cargo-update: false
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@master
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,6 +23,9 @@ jobs:
           - stable
           - beta
           - nightly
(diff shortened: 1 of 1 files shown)


youknowone:
Running cargo update matrix for every rust version doesn't looking worth that much. Though we don't pay for CI resource, it is a common pool for open source projects.
How about only for stable? That will be enough.

youknowone:
I prefer this approach to #755

dburgener:
> How about only for stable?

How about the other way around?  Still run only 5 workflows instead of the full 8, but run "cargo update" in front of 4 of them and only test the locked version on stable.  I would prefer to give extra test coverage to the updated version, because that most mimics what library users are likely to see.  The only users who are affected by the Cargo.lock are binary users.  We care about them and should definitely test the lock, but if we're picking and choosing, I'd prefer to give priority to the updated version to prioritize catching issues for the library users as early as possible.

youknowone:
That's making more sense! Thank you

dburgener:
I updated the PR as discussed above.  CI run for this PR looks as expected.  5 full build/test runs, four with `cargo update` and just stable on the locked version.

youknowone (approved):
Thank you!

PR #791, opened 2023-04-29
Add a default feature 'unicode'

### LINKED ISSUE #789: Adding a feature `unicode`
The problem happens because it cames from user side and lalrpop doesn't know users use what kind of features of regex.

- The feature will turn on `regex/unicode` and `regex-syntax/unicode`.
- The feature will be a default feature.

This will let lalrpop users can use any unicode regex when they want but can reduce dependency down they don't need one.

Does this make sense?

youknowone:
Fix #789


youknowone:
ah, already done in #786

PR #794, opened 2023-04-30
Move cargo hack to it's own workflow and cleanup

Pat-Lafon:
With the addition of `cargo hack` into the `ci.sh` script, the github workflows run has significantly slowed down. This also makes it more inconvenient to add further things to the workflow like #790. I think the workflow should prioritize failing fast and speed.

To this end, I've moved the `cargo hack test` call out of the `ci.sh` script into it's own workflow. This invocation is a bit heavy weight to be the first thing a contributor runs and can be better parallelized on it's own. Maybe something like https://github.com/taiki-e/cargo-hack/pull/180.

Note I've ignore two features from the `cargo hack` invocation: `pico-args` and `default`. `pico-args` is only needed for the cli so it's unnecessary to include with the tests.`default` is excluded because it added an extra unnecessary dimension to the features being run(as all default does is enable other features, it is not used in the code itself). Add the `--print-command-list` flag to your cargo hack call to see all of the commands that will be run and with what features.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -68,6 +68,8 @@ jobs:
       - name: Update
         run: cargo update
       - run: cargo build --bin lalrpop --features pico-args
+      - name: Build feature powerset
+        run: cargo hack build --workspace --feature-powerset --exclude-features pico-args,default --optional-deps
       - name: Run feature powerset check
         # test with minimal amount of features plus a few extra on regex/regex-syntax
         run: cargo hack test --workspace --feature-powerset --exclude-features pico-args,default --optional-deps
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,9 +30,6 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
-          components: rustfmt
-      - name: Install cargo-hack
-        uses: taiki-e/install-action@cargo-hack
       - name: Update
         if: matrix.cargo-update
         run: cargo update
@@ -59,3 +56,18 @@ jobs:
           toolchain: nightly
       - name: Check the minimum possible crate versions
         run: rm Cargo.lock && cargo +nightly build -Zdirect-minimal-versions
+  feature_powerset:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+            toolchain: stable
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
+      - name: Update
+        run: cargo update
+      - run: cargo build --bin lalrpop --features pico-args
+      - name: Run feature powerset check
+        # test with minimal amount of features plus a few extra on regex/regex-syntax
+        run: cargo hack test --workspace --feature-powerset --exclude-features pico-args --optional-deps
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -58,6 +58,8 @@ jobs:
         run: rm Cargo.lock && cargo +nightly build -Zdirect-minimal-versions
   feature_powerset:
     runs-on: ubuntu-latest
+    env:
+      CARGO_INCREMENTAL: 1
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@master
@@ -68,8 +70,6 @@ jobs:
       - name: Update
         run: cargo update
       - run: cargo build --bin lalrpop --features pico-args
-      - name: Build feature powerset
-        run: cargo hack build --workspace --feature-powerset --exclude-features pico-args,default --optional-deps
       - name: Run feature powerset check
         # test with minimal amount of features plus a few extra on regex/regex-syntax
         run: cargo hack test --workspace --feature-powerset --exclude-features pico-args,default --optional-deps
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,9 +7,6 @@ on:
     branches: ["master"]
   merge_group:
 
-env:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -7,9 +7,6 @@ on:
     branches: ["master"]
   merge_group:
 
-env:
-  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
```
dburgener, line 11: Just curious, do you have documentation you can link about rust-toolchain managing the crates.io lookup?  I was under the impression that rust-toolchain would handle the initial toolchain download and install, but this is referring to grabbing the crates later with cargo update.  I could easily be wrong, but I couldn't find any relevant rust-toolchain docs in a little bit of googling.

tools/ci.sh
```diff
@@ -6,8 +6,6 @@ export CARGO_INCREMENTAL=0
 export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
 cargo build --bin lalrpop --features pico-args
-# build with minimal amount of features to make sure dependencies can build too.
-cargo hack build --workspace --feature-powerset --exclude-features pico-args --optional-deps
```
dburgener: Did you do any measurement on how much time dropping the feature-powerset build saves in CI?  My impression is that test is the really expensive bit.  I'm inclined to agree with dropping the powerset test.  The really important thing is just getting the single binary the tests need, but we do get *some* value here, since "cargo build" tests a slightly different configuration that "cargo test" does.  It's probably not enough value to be worth the extra CI time, but I'm curious if we have any data to compare what it's costing us to leave this in.

(If we were to leave it in, it should maybe be reordered to before the regular build to ensure we run tests with a specified lalrpop binary, rather than whatever hack happened to decide to do last.)

.github/workflows/test.yaml
```diff
@@ -42,3 +39,16 @@ jobs:
         run: cargo fmt --all -- --check
       - name: Run clippy
         run: cargo clippy --all --all-targets -- -D warnings
+  feature_powerset:
```
dburgener, line 42: Just a thought.  Would we want to add "needs: lint" here?  It would lengthen the total time spent on PR testing, since this wouldn't start until after linting is done, but it would avoid doing this heavyweight check on a PR that's just failing on format or clippy anyways.

dburgener:
+1 to the general idea here.  Just left a few comments/questions in the PR about some of the specific details.

.github/workflows/test.yaml
Pat-Lafon: True the documentation is vague in saying "good defaults" but there are some pr's which show this. https://github.com/dtolnay/rust-toolchain/pull/27 https://github.com/dtolnay/rust-toolchain/pull/66 https://github.com/dtolnay/rust-toolchain/pull/54

.github/workflows/test.yaml
Pat-Lafon: Hmmm, what do you think compared to https://stackoverflow.com/questions/74690423/how-can-i-use-github-actions-concurrency-to-cancel-other-jobs-if-any-of-them-f ?

tools/ci.sh
Pat-Lafon: This seems to hold up in that it was only a 20s saving but probably close enough to some margin of error. 

`since "cargo build" tests a slightly different configuration that "cargo test" does` In what way? Maybe this is what I'm unclear about.

Pat-Lafon: hmm running with incremental but without cargo hack build has given the best time so far... but I'm not sure what my error margin is when doing this.

Pat-Lafon: (Inspired by https://github.com/dtolnay/rust-toolchain/issues/61)

.github/workflows/test.yaml
dburgener: That's good enough for me.  BTW, it looks like all of those environmental variables are also set at the top of tools/ci.sh.  Do we want to remove them there as well?

tools/ci.sh
dburgener: > since "cargo build" tests a slightly different configuration that "cargo test" does In what way? Maybe this is what I'm unclear about.

IIUC there are two major differences:

1. The final binary isn't linked together.  I doubt we care too much about this difference.  I'm not sure if there's anything that potentially could go wrong here that wouldn't be a rust bug.
2.  `#[cfg(test)]` blocks wouldn't be included in build.  This could potentially cause issues if a certain block of code was conditionally only needed under certain features, but was also marked `#[cfg(test)]`, I believe.

IMHO, running just the single build without a hack, and then running test on the full hack seems like a sensible compromise between CI usage and test coverage.  I don't think we're getting a ton of added value in doing the hack build, and I worry that we could end up causing unneeded CI failures by accidentally using the wrong built version during `cargo test`.

.github/workflows/test.yaml
dburgener: I tend to try to minimize my use of third party actions, since they're running potentially untrusted code in my CI pipeline with access to secrets, but that's above average on the paranoia level.  If I were the maintainer, I'd prefer the straight dependency to avoid adding the extra external action, but I won't pitch a fit if you/the maintainers prefer to go the other way.

Pat-Lafon: Fair, it's a trade off between "average" runtime usage, "worst-case" runtime usage, and additional outside dependencies. I don't have a strong stance here.

tools/ci.sh
Pat-Lafon: Yeah, I think those differences are unlikely to come up in practice. I also am in favor of removing `cargo hack build` purely on it being simpler for what is basically the same benefit(and then not needing to rely on cargo properly reusing build artifacts).

youknowone (approved):
Thank you!

@dburgener would you add a linter dependency to the new job later?

dburgener:
> would you add a linter dependency to the new job later?

Sure.  I'll send a PR later this week.

PR #796, opened 2023-05-02
[RFC] Don't run feature-powerset test until after linting

dburgener:
This is so that we don't waste resources on the heaviest test if we're just going to fail on format or similar.

It does lengthen the CI runtime by a little bit, but that's async anyways.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -58,6 +58,8 @@ jobs:
         run: rm Cargo.lock && cargo +nightly build -Zdirect-minimal-versions
   feature_powerset:
     runs-on: ubuntu-latest
+    # Don't do expensive test if we're just going to fail on linting
+    needs: lint
     env:
       CARGO_INCREMENTAL: 1
     steps:


dburgener:
You can watch the test sequencing for this PR run here: https://github.com/lalrpop/lalrpop/actions/runs/4864958522

I'm not 100% convinced on this change, but I'm putting it up for wider discussion.

After the work by @Pat-Lafon in #794, The feature-powerset test is nowhere near as heavy-weight as it was.  It takes a little longer than the other build/test runs, but not a lot.  Sequencing linting in front of it allows us to early-abort and not do it if linting fails, but it lengthens the full CI runtime on success by about 3 minutes.

IMO both options are pretty acceptable.  When this originally came up, feature-powerset was running for 12 minutes, and this seemed worth more.  We could also consider other options as discussed in this comment chain: https://github.com/lalrpop/lalrpop/pull/794#discussion_r1181293068

youknowone (approved):
Let's save the earth unless it hurts development process too much.

PR #797, opened 2023-05-03
Remove some debugging statements

Pat-Lafon:
I was poking around at the compile times and I realized there was an incredible amount of error printing going on. Seems like some debugging code got merged, whoops. Removing these saves at least a minute in `cargo hack test`.


Pat-Lafon:
There are most likely some sizeable wins in splitting up lalrpop's modules and lalrpop-tests modules into their own subcrates. (compiling these two crates is like 60-80% of the build time depending on flags I think, splitting into subcrates would give parallelism).

Also in #662 which I think is in part responsible for larlpop-test's build script taking 2 seconds before we even get to compilation.

Based on running `cargo clean && cargo build --timings`

youknowone (approved):
Thank you!

cc @nikomatsakis please check if this is intentionally added

PR #798, opened 2023-05-03
Describe lexer unicode support in the book

dburgener:
Clarify the new unicode feature, and add suggestions for avoiding the unicode dependency if it's not needed.  Some users may be depending on unicode without intending to, so give them migration tips to avoid the requirement.


youknowone (approved):
Thank you! It looks good :smile:

PR #799, opened 2023-05-07
Add deprecation notice for process_root_unconditionally

Pat-Lafon:
Given the comment on process_root_unconditionally, seems like there could be an additional annotation to help elevate this.


youknowone (approved):

youknowone:
Thank you! `#[deprecate]` must be a lot better than deprecated note only on document

PR #804, opened 2023-06-14
User error visibility

### LINKED ISSUE #519: Generated source contains `pub`
After generating a grammar with lalrpop 0.18.1 the resulting file contains a number of `pub` entires:

* `pub struct __StateMachine`
* `pub trait __ToTriple`
* `pub enum __Symbol`

These leak into my library API, which isn't great. But worse, because `__Symbol` contains all the AST nodes of my grammar, it forces those nodes to be public too. Should lalrpop only ever use `pub(crate)`? Or is there a configuration to avoid public symbols? I already use `pub(crate) TopLevel: ...` in my definitions.

arnaudgolfouse:
This reduces the visibility of various items internal to LALRPOP, with the goal of allowing a non-public custom error type.

I believe this also fixes #519.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -365,7 +365,19 @@ fn emit_recursive_ascent(
         exit(1);
     }
 
+    // Find a better visibility for some generated items.
+    // This will be the maximum of the visibility of all starting nonterminals.
+    let mut max_start_nt_visibility = pt::Visibility::Priv;
     for (user_nt, start_nt) in &grammar.start_nonterminals {
+        match (
+            &max_start_nt_visibility,
+            &grammar.nonterminals[start_nt].visibility,
+        ) {
+            (r::Visibility::Pub(None), _) | (_, r::Visibility::Priv) => {}
+            (v1, v2) if v1 == v2 => {}
+            (r::Visibility::Priv, v) => max_start_nt_visibility = v.clone(),
+            _ => max_start_nt_visibility = r::Visibility::Pub(None),
+        };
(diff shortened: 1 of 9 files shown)


arnaudgolfouse:
By the way, it looks like the contributing guidelines are slightly outdated : it suggests activating the feature `test`, but it doesn't exist.

lalrpop/src/build/mod.rs
```diff
@@ -365,7 +365,19 @@ fn emit_recursive_ascent(
         exit(1);
     }
 
+    // Find a better visibility for some generated items.
+    // We try to find a good enough common denominator for all starting nonterminals.
```
yannham, line 369: It seems what you do is to compute the max of the visibility of all non terminals. In this case, I would just spell this out simply, instead of "try to find a good enough common denominator", which is IMHO a bit vague.

yannham (approved):
Looks reasonable and implemented in a sensible way :+1: 

I understand that this shouldn't break compatibility in any meaningful way, because what you made private are internal types (in which case, if someone relied on it, it's at their own risk), and otherwise `__To_Triple` is still public if at least one non terminal is public. Is that correct?

arnaudgolfouse:
Yes, this only changes the visibility of types and functions that are prefixed by `__`, which I assume are all internal and should not be accessed anyway.
`__To_Triple` needs to be public if the starting nonterminals are public, because it is then a trait bound on a public function.

lalrpop/src/build/mod.rs
arnaudgolfouse: You're right, I changed the comment 🙂

PR #805, opened 2023-07-05
Fix dependency on is-terminal on specific version 0.4.7

jannic:
With is-terminal 0.4.8, the MSRV was changed from 1.48 to 1.63. This implicitly affects the MSRV of lalrpop, breaking existing builds.

I think a MSRV change should be considered a breaking change according to semver rules. But even with a less strict interpretation of semver, it's probably a bad idea to change MSRV without a version update at all, breaking existing build setups.
(Example:
https://github.com/rp-rs/rp-hal/actions/runs/5458734797/jobs/9934102344?pr=642)

As is-terminal has a policy of changing MSRV in a minor release, and actively suggests to pin a specific version to avoid that, I think lalrpop should do that and use exactly version 0.4.7. (https://github.com/sunfishcode/is-terminal/blob/v0.4.7/README.md?plain=1#L97-L99)

### DIFF
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -23,7 +23,7 @@ ascii-canvas = { version = "3.0", default_features = false }
 bit-set = { version = "0.5.2", default_features = false }
 diff = { workspace = true }
 ena = { version = "0.14", default_features = false }
-is-terminal = "0.4.2"
+is-terminal = "=0.4.7"
 itertools = { version = "0.10", default_features = false, features = ["use_std"] }
 petgraph = { version = "0.6", default_features = false }
 regex = { workspace = true }


yannham:
I see on master than LALRPOP has a MSRV defined to `1.64`, so I  guess you are talking about the released version.

Honestly, it's a bit of a headache. If we start to follow this approach, this means that we either have to pin everything at an exact version, or monitor the changes of all of our transitive dependencies. Note that the approach of not increasing the major version upon MSRV increase seems to be standard practice, or even recommended practice: https://github.com/rust-lang/api-guidelines/discussions/231.

On the other hand, I agree that breaking existing build without a version bump isn't great. Despite that shouldn't really "break existing builds", because you'll have to update `is-terminal` in the Cargo.lock for the build to break, right? So it's rather that if you start a fresh project with a dep on LALRPOP, your MSRV is now higher than before the `is-terminal` update. I personally wouldn't deviate from the community's approach and also I would be reluctant to implement a general policy that would either require to pin everything or to add a lot of work for maintainers. @youknowone @nikomatsakis any opinion on that?

jannic:
I see your point, and I'm also not happy with adding exact dependencies everywhere.
As far as I understand the situation, this is a fundamental issue, and every attempt to solve it in individual crates is necessarily an ugly workaround:
- if everybody agreed that an MSRV bump was a breaking change, it would be easy to keep things consistent
  - at the cost of a lot more version bumps, causing busy work throughout the whole ecosystem
- alternatively, if everybody ignored MSRV and just accepted that one might need to use latest rust, things would be easy as well
  - that would effectively void the concept of having a defined MSRV
- the combination is just not working. As soon as there is a dependency chain, things fall appart: If crate A depends on B which depends on C, B and C have policies that allow for MSRV bumps in patch releases, and C bumps its MSRV, the MSRV of B will change implicitly without a version change at all, and A can't avoid that MSRV bump. Note that at that point, even an explicit dependency on a specific version of B won't help.
  - if A is a binary crate, that can be solved using Cargo.lock
  - if A is a library crate, it is no longer possible to specify a well-defined MSRV for A.
    - (A could add an artificial direct, specific dependency on C. Like in https://github.com/rp-rs/rp-hal/pull/643. Not nice, but should work.)

Overall, this should probably be discussed in https://github.com/rust-lang/api-guidelines/discussions/231. So I'd understand if you didn't merge this PR for now.

youknowone:
I think @jannic made the point well.

In my opinion, it is logically not a bug for either case.
If bumping MSRV is not a breaking change, we don't need a release. Users can pin the version of is-terminal in that case.
If bumping MSRV is a breaking change, this is a bug of is-terminal, but not one of lalrpop.

Also, as @yannham pointed out, since MSRV of our master branch 1.64, this patch will not be helpful for next version users.

yannham:
Thanks @jannic, this is a detailed and compelling point. I would lend toward not merging this PR, for the previously mentioned reasons plus the practical side:

- pinning a dependency impacts downstream users as well. Say that I WANT to use `is-terminal` v0.4.8 because it works for my strange use-case, and I also use LALRPOP. I believe cargo will yell if I pin it now, because there are two semver-compatible but different pinned versions of `is-terminal`. Which is not great.
- following the previous point, users might just add `is-terminal = "=0.4.7"` to their specific project if they need to build but Rust < 1.63.

Although I don't have a very strong opinion, and accepting this change wouldn't be the end of the world.

jannic:
Given that there are also strong arguments against adding such a dependency, and that it doesn't apply to the master branch at all (it would only useful as part of a bugfix release), I retract the pull request.
In case there are other arguments in favor of this change, we can always re-open the PR.

Thanks for the constructive discussion!

PR #807, opened 2023-07-12
Continue PR #341 (Allow `_` in `match` to appear earlier on)

fpoli:
This is the same as https://github.com/lalrpop/lalrpop/pull/341 by @k0pernicus, with an additional commit to test overlapping regular expressions as asked in the review.

(I happened to need exactly this feature :smile:)


lalrpop/src/normalize/token_check/test.rs
```diff
@@ -169,6 +169,36 @@ fn match_catch_all() {
...
+            r"\w+"
+        }
+        pub Term = {
+            Num,
+            "(" <Term> ")",
+            r"\w+" => format!("Id({})", <>),
+        };
+        Num: String = r"[0-9]+" => <>.to_string();
+"#;
+    assert!(validate_grammar(&grammar).is_ok());
```
Pat-Lafon, line 192:
```suggestion
    assert!(validate_grammar(grammar).is_ok());
```

yannham (approved):
Many thanks for reviving that old work.

I'm not very knowledgable on this part of the code, but reading the previous issues, PRs, and @nikomatsakis messages, this looks good to me, as far as I can tell.

I'll let other maintainers take a look if they have the time to do so (pinging @youknowone as well), but otherwise, if no one complains, I'll go ahead and merge it at some point.

yannham:
Ah, there is a test failing though. Sounds like the grammar isn't accepted when all default features are disabled. Maybe something related to this unicode issue (e.g. #810)?

lalrpop/src/normalize/token_check/test.rs
```diff
@@ -169,6 +169,36 @@ fn match_catch_all() {
     assert!(validate_grammar(grammar).is_ok())
 }
 
+/// Test that a `catch-all` can be use in the first `match` arm.
+/// Before the pull request to close [issue 325](https://github.com/lalrpop/lalrpop/issues/325),
+/// the usage of the `catch-all` symbol was not allowed in the first arm of a `match` block.
+#[test]
+fn match_catch_all_in_first_arm() {
```
fpoli, line 176:
```suggestion
// This test requires regex's unicode case support
#[cfg_attr(not(feature = "unicode"), ignore)]
fn match_catch_all_in_first_arm() {
```

fpoli:
> Ah, there is a test failing though. Sounds like the grammar isn't accepted when all default features are disabled. Maybe something related to this unicode issue (e.g. #810)?

Thanks, based on that issue I added `#[cfg_attr(not(feature = "unicode"), ignore)]` and it now works (although I'm not sure why that feature is needed).

yannham:
> (although I'm not sure why that feature is needed)

I believe this is the point of the issue: we shouldn't require unicode by default in the lexer, but unfortunately, we currently do, which is an error. We might probably get rid of the attribute once #810 is closed.

youknowone:
@yannham If this patch itself doesn't have problem, can we merge it first and fix the unicode feature stuff later? Or will it affect to users with/without unicode feature?

yannham:
I think we can just remove the `#[cfg(not...]` now that the unicode part has been fixed, and it should be good to go?

youknowone:
I tried it but didnt work only with it. I didnt look in details.

dburgener:
I believe the issue here is the `\w` character matches unicode "word" characters, not just ASCII ones.

See the regex docs: https://docs.rs/regex/latest/regex/#perl-character-classes-unicode-friendly and the lalrpop docs: https://lalrpop.github.io/lalrpop/lexer_tutorial/001_lexer_gen.html#unicode-compatibility

Replacing "\w" with "[[:word:]]" fixes the issue (I believe "[0-9a-zA-Z_]" would work as well).  Demonstrated here: https://github.com/dburgener/lalrpop/pull/4

I'm happy to either push those two commits onto this branch, or let the original contributor change it.

As a side note, I wonder if we should do something like silently rewrite the unicode character classes to ASCII only ones if the unicode feature isn't enabled.  The fact that using "\s", "\w" "\d" etc implicitly requires unicode support feels pretty counter intuitive to me.

dburgener:
Since I haven't heard anything from the submitter, I went ahead and pushed my fix on this remote.  All the checks are passing now, without the unicode ignore.

yannham (approved):

Issue #811, opened 2023-07-20
Assertion instead of error message


fabi321:
When using a macro without any arguments, an assertion is triggered:
https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/normalize/prevalidate/mod.rs#L381

Example:
```
grammar;
Macro<Smth>: String = {
    Smth => <>.to_string()
}

pub Root: String = {
    Macro<>
}
```
This definition triggers the assertion, instead of raising an error message. Changing it to `Macro<"a">` fixes it.
Version: 0.20.0

Pat-Lafon:
Is https://github.com/Pat-Lafon/lalrpop/commit/23ee8f7eab9de1fddcbf6227a69782634d7c93a6 all that's needed to fix this?

fabi321:
looks fine

```
  processing file `~/tmp-lalrpop/src/tmp.lalrpop`
  ~/tmp-lalrpop/src/tmp.lalrpop:7:5: 7:11 error: macros must have atleast one argument

  --- stderr
        Macro<>
        ~~~~~~~
```

Pat-Lafon:
@fabi321 This issue can be closed now?

PR #812, opened 2023-07-20
patch #811

Pat-Lafon:
Convert trigger-able assertion to error. See https://github.com/lalrpop/lalrpop/issues/811

I'm not sure if there are tests for grammar files that are intended to error. Let me know if something like that should be added.

### DIFF
--- a/lalrpop/src/normalize/prevalidate/mod.rs
+++ b/lalrpop/src/normalize/prevalidate/mod.rs
@@ -378,7 +378,9 @@ impl<'grammar> Validator<'grammar> {
                 }
             }
             SymbolKind::Macro(ref msym) => {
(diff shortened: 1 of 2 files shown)


yannham (commented):
This looks fine. We indeed have test that are supposed to error, I've seen them inside LARLPOP modules themselves, see eg https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/normalize/prevalidate/test.rs. It's rudimentary but I guess this case could fit in such a small test case, which would be nice indeed. Ideally we would use something like `cargo insta`, but in the meantime...

Pat-Lafon:
Whoops, I missed that. Fixed

yannham (approved):

yannham:
Thanks!

PR #815, opened 2023-07-31
Drop mention of adding regex dependencies from book.

dburgener:
The regex dependency is not and should not be a dependency of crates depending on lalrpop (unless they need regex directly).  This was presumably added to work around lalrpop's previous feature underspecification, but is no longer needed.  Using the built-in lexer works just fine with the normal dependencies, indirectly pulling lalrpop's regex dependency.


youknowone (approved):
👍

PR #816, opened 2023-08-03
Modernize extern crate

Pat-Lafon:
Hello, I was looking through the docs again to write a parser and I realized they have some excess code in the form of `extern crate lalrpop;` and setting the build file to `build.rs` which is the default.

I've gone through and done my best to remove these throughout the code base.


youknowone (approved):
Thank you!

PR #817, opened 2023-08-05
Remove dependency on `is-terminal` crate

Kmeakin:
The same functionality as `is_terminal::is_terminal` is now provided by `std::io::IsTerminal::is_terminal()`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - "1.64"
+          - "1.70"
           - stable
           - beta
           - nightly
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,12 +47,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cc"
-version = "1.0.79"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50d30906286121d95be3d479533b458f87493b30a4b5f79a607db8f5d11aa91f"
-
-[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


Pat-Lafon:
So is this just waiting for 1.72 to come out to update the msrv to 1.70?

Kmeakin:
> So is this just waiting for 1.72 to come out to update the msrv to 1.70?

Yes

Pat-Lafon:
@Kmeakin I think you just need to update https://github.com/lalrpop/lalrpop/blob/3d5195b454ec8f64c911afd35ecd997c77231e52/Cargo.toml#L23 and https://github.com/lalrpop/lalrpop/blob/3d5195b454ec8f64c911afd35ecd997c77231e52/.github/workflows/test.yaml#L19

yannham (approved):
Thanks, one less dependency :slightly_smiling_face:  LGTM. @youknowone do we want to block on that a bit, as 1.70 is quite close to the current Rust stable version? But I suspect we'll see a version bump before we release anything, and `n-2` was one of the original suggestion as a MSRV. I'm fine either way, merging this right now or later

Pat-Lafon:
@yannham FWIW I don't think this should be blocked only because this waited until now to be within the msrv policy that was set with Niko's input. In https://github.com/lalrpop/lalrpop/issues/742 I believe we hashed out `N-2` as an appropriate amount of time to wait before moving up the msrv. There doesn't seem to be a reason to wait any longer here.

My understanding of trying to make that policy is that the maintainers didn't want to rehash this issue every time it comes up that the msrv value could be changed.

youknowone:
I don't have strong opinion.

My weak idea is this patch is not adding feature but keeping compatibility while bumping rust version. In other words, one more dependency to keep 6 months release compatibility looks good deal to me unlike making a new feature patch in pain way. Even more, it will be really easy to rebase later. So nothing to lose.

Still, I don't have strong opinion about it. Please go ahead if anyone think this is good to go right now.

yannham:
Thanks for your input both, I think it's reasonable to proceed.

PR #818, opened 2023-08-05
Move dependency on `diff` crate to `dev-dependecies`

Kmeakin:
The `diff` crate is only used for tests, so move it to `dev-dependencies`. This should mean end users have one less crate to download and build.


youknowone (approved):
Thank you!

PR #819, opened 2023-08-06
Remove `extern crate` since we don't longer support edition 2015

youknowone:


youknowone:
oh #816 already did it.

PR #821, opened 2023-08-07
Filter out NotFound errors

### LINKED ISSUE #808: Dangling symlinks anywhere in the crate cause obscure build failures
How to reproduce:
1. Create a new crate with `build.rs` file `fn main() { lalrpop::process_root().unwrap() }`
2. Create a dangling symlink: `ln -sf /file/does/not/exist dangling-symlink`
3. Run `cargo build`:

```shell
$ cargo build
   Compiling cargo-dangling-symlinks v0.1.0 (/home/liri/cargo-dangling-symlinks)
error: failed to run custom build command for `cargo-dangling-symlinks v0.1.0 (/home/liri/cargo-dangling-symlinks)`

Caused by:
  process didn't exit successfully: `/home/liri/cargo-dangling-symlinks/target/debug/build/cargo-dangling-symlinks-6e4fe4fd1c3bb576/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', build.rs:2:29
```

Full crate example: https://github.com/legeana/lalrpop-dangling-symlinks

This was quite frustrating to debug as the error message is extremely obscure. I had to resort to `strace -f target/.../build-script-build` to determine what file lalrpop was trying to open.

I believe lalrpop should do two things:
1. Improve error messages by indicating the path that doesn't exist.
2. Ignore dangling symlinks.

Pat-Lafon:
Closes #808 

This pr will ignore any erroneous symlinks when looking for lalrpop files and will not propagate up the error.

The related issue also mentions adding some kind of warning message. I'm not sure how/if such a thing should be added especially since we are otherwise skipping over this error(So it would mean using `eprintln` I guess?).

Better error messages would certainly be more helpful but in this case lalrpop is just propagating up whatever https://doc.rust-lang.org/std/io/type.Result.html gives us so poor error messages seem like a libs team issue https://github.com/rust-lang/rfcs/issues/2885.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -186,16 +186,17 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
             result.extend(lalrpop_files(&path)?);
         }
 
-        let is_symlink_file = || -> io::Result<bool> {
+        let is_symlink_file = || -> bool {
             if !file_type.is_symlink() {
-                Ok(false)
+                false
             } else {
-                // Ensure all symlinks are resolved
-                Ok(fs::metadata(&path)?.is_file())
+                // Ensure all symlinks are resolved to a file
+                // Or ignore eronious ones https://github.com/lalrpop/lalrpop/issues/808
+                fs::metadata(&path).map_or(false, |m| m.is_file())
             }
         };
 
-        if (file_type.is_file() || is_symlink_file()?)
+        if (file_type.is_file() || is_symlink_file())
             && path.extension().is_some()
             && path.extension().unwrap() == "lalrpop"
         {
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -186,17 +186,26 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
             result.extend(lalrpop_files(&path)?);
         }
 
-        let is_symlink_file = || -> bool {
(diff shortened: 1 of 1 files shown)


lalrpop/src/build/mod.rs
```diff
@@ -186,16 +186,17 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
             result.extend(lalrpop_files(&path)?);
         }
 
-        let is_symlink_file = || -> io::Result<bool> {
+        let is_symlink_file = || -> bool {
```
yannham, line 189: Nitpick: I wonder if this small lambda could reflect slightly better the new semantics, something like `is_valid_symlink_file`, or a better word? It's of course not very important, you can dismiss if you feel like it

yannham (commented):
Thanks for the contribution!

Mh, I guess `eprintln!` would indeed make sense (in most error reporting framework a warning is just one specific kind of error, I think). I think silently failing can also create new confusing situations (I have a `grammar.lalrpop` file but LALRPOP doesn't pick it?). Although this sounds a bit primitive, I'm tempted to say that a good old `eprintln!("Warning: ignoring dangling symlink {path}")` would at lest avoid this situation. And we can improve on it later if that proves to be limiting?

lalrpop/src/build/mod.rs
Pat-Lafon: Sure, seems reasonable

yannham (approved):

PR #822, opened 2023-08-10
miscellaneous clippy/cfg_attr rustfmt deprecations

Pat-Lafon:
I've procrastinated a little bit by seeing what happens if `clippy::all` was removed from the generated code.

In some cases it just meant making the lints more precise in what was being allowed(like `needless_lifetimes`, `just_underscores_and_digits`, etc).

It did catch that in a couple places `#[cfg_attr(rustfmt, rustfmt_skip)]` was being emitted which has been deprecated in favor of `#[rustfmt::skip]`. There were a bunch of unnecessary clones that have been removed.

It also suggests adding a `Default` impl to any parsers that are generated which might have a niche application with the more built-in support for default Rust has been getting.

There were two cases where trivial math was being generated like `* 1` and `1 - 1` that I removed(which are actually clippy errors not warnings).

Updating the `lrgrammar.rs` with `snap.sh` is really noisy but I guess that's what happens when you have a 40k line file checked in

In part this is motivated by #769 in trying to limit how much of clippy is allowed by default.

### DIFF
--- a/lalrpop/src/build/action.rs
+++ b/lalrpop/src/build/action.rs
@@ -120,7 +120,7 @@ fn emit_user_action_code<W: Write>(
         ]);
     }
 
-    rust!(rust, "#[allow(clippy::too_many_arguments)]");
+    rust!(rust, "#[allow(clippy::too_many_arguments, clippy::needless_lifetimes, clippy::just_underscores_and_digits)]");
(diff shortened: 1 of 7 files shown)


youknowone (approved):
👍

PR #824, opened 2023-08-24
Clippy 1.72 release fixes

Pat-Lafon:
With the new clippy release, I've gone through and either manually or `cargo clippy --fix`ed the triggered lints. I don't think anything here is particularly interesting/useful but may as well not let things slip.

### DIFF
--- a/lalrpop/src/lr1/build/test.rs
+++ b/lalrpop/src/lr1/build/test.rs
@@ -281,6 +281,7 @@ T: () = {
 fn issue_144() {
     let _tls = Tls::test();
 
+    #[allow(clippy::needless_raw_string_hashes)] // Fix merged for next clippy release, after 1.72
     let grammar = normalized_grammar(
         r##"
 grammar;
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -852,7 +852,19 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                 rust!(
                     self.out,
                     "{} => match {}token {{",
-                    indices.iter().map(|(index, _)| index).format(" | "),
+                    if indices.windows(2).all(|w| w[0].0 + 1 == w[1].0) {
+                        format!(
+                            "{}..={}",
+                            indices.first().unwrap().0,
+                            indices.last().unwrap().0
+                        )
+                    } else {
+                        indices
+                            .iter()
+                            .map(|(index, _)| index)
+                            .format(" | ")
+                            .to_string()
+                    },
                     self.prefix
                 );
                 rust!(
@@ -871,7 +883,19 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                 rust!(
                     self.out,
                     "{indices} => {p}Symbol::{variant_name}({p}token),",
-                    indices = indices.iter().map(|(index, _)| index).format(" | "),
+                    indices = if indices.windows(2).all(|w| w[0].0 + 1 == w[1].0) {
+                        format!(
+                            "{}..={}",
+                            indices.first().unwrap().0,
+                            indices.last().unwrap().0
+                        )
+                    } else {
+                        indices
+                            .iter()
+                            .map(|(index, _)| index)
+                            .format(" | ")
+                            .to_string()
+                    },
                     p = self.prefix,
                     variant_name = variant_name,
                 )
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -805,7 +805,11 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
             .emit()?;
         rust!(self.out, "{{");
 
-        rust!(self.out, "match {p}token_index {{", p = self.prefix,);
+        rust!(
+            self.out,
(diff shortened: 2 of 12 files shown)


lalrpop/src/lr1/build/test.rs
```diff
@@ -281,6 +281,7 @@ T: () = {
 fn issue_144() {
     let _tls = Tls::test();
 
+    #[allow(clippy::needless_raw_string_hashes)] // Fix merged for next clippy release, after 1.72
```
dburgener, line 284: I vote against the comment that will get stale really quickly.

lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -852,7 +852,11 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                 rust!(
                     self.out,
                     "{} => match {}token {{",
-                    indices.iter().map(|(index, _)| index).format(" | "),
```
dburgener, line 855: I guess clippy made a suggestion on the generated code, motivating this change?

It looks like the logic is equivalent iff the series of indices is ordered and always increasing by one.  It should be ordered, since it's the result of `.enumerate()`, but can't it possibly not be increasing by one if the None case occurs on line 823 in between two instances of the Some() case?

lalrpop/src/lr1/build/test.rs
Pat-Lafon: I'm confused? Yes it will get stale but the point is that, at that point, this allow can be removed. Clippy sometimes has false positives that get fixed in a subsequent version so we may as well keep track of false positives and when they can be removed.(Here the fix has already been merged into master)

lalrpop/src/lr1/codegen/parse_table.rs
Pat-Lafon: Yes, clippy is more aggressively linting on or patterns of incrementally increasing values.

That's an interesting point, is there a small test case that helps trigger this? All of the current tests are of the logically equivalent form.

Pat-Lafon: Ok, nervermind, found one in lrgrammar.rs

Pat-Lafon: I've attempted a fix for this to check the ordering first.

lalrpop/src/lr1/build/test.rs
dburgener: Oh, I misunderstood.  I thought you were saying this allow wasn't needed yet, but would be needed after the clippy update to 1.72.  It's exactly the other way around.

In that case, maybe the comment would be more clear if it specified the expected action of removing this after 1.72?

lalrpop/src/lr1/codegen/parse_table.rs
dburgener: The fix looks right.  The weird case is exactly 1 terminal, which looks like it ends up using the inclusive range `N..=N`, instead of the old `N`.  That's probably a really uncommon case though, and an inclusive range of length one is just weird, not wrong.

Is this worth the extra code complexity?  Is it just about making clippy happy, or is there a performance benefit to replacing "1 | 2 | 3 | 4" with "1..=4"?

Pat-Lafon: It compiles to the exact same code. Entirely just a clippy thing.

dburgener: Well, a maintainer should have a final say, but I feel as though going from 1 fairly straightforward line to 13 lines with a fair amount of logic to parse and think though isn't worth it to make clippy happy in the generated code.  I would vote for just silencing this one to keep the lalrpop code more readable.

Pat-Lafon: Certainly up to the maintainer, I can understand the code is more complex than before to handle this case.

However, it's a little disingenuous to say it went from 1 line to 13 lines. For example, the old case is still there but now takes 5 lines because of formatting... that part didn't suddenly become more complex.

dburgener: I'd just call that "lines of code is a bad metric".  But it's a convenient shorthand.  Personally, I think the fact that you've got two cases instead of 1 is an increase in complexity, and the logic in the window closure, particularly with indexing is certainly more involved than a simple map.

It's not a huge difference, but the "make clippy happy" I would call a near zero benefit in this case, since clippy is really analyzing the readability of the generated code, but our level of care about that is near-zero, and our level of care about the readability of the actual lalrpop code is much higher.

youknowone: I agree to @dburgener. I believe clippy suggests it for readability.
But we maintain code generator, not the generated code here. People will read more generator code than generated code. In my opinion, previous one looks more readable.

Pat-Lafon: Fair, I've reverted this

youknowone: @Pat-Lafon will putting clippy ignore to line 808 work? 
https://github.com/lalrpop/lalrpop/pull/824/files#diff-f6fa48d7b0092fbf2bbdf4a7413478e304b5b2127fd314a3848b64e022f7aa71R808

Pat-Lafon: Sure, lets try that

Pat-Lafon: Marking this as resolved

youknowone (approved):
@Pat-Lafon @dburgener Thank you so much!

Issue #710, opened 2023-03-04
Getting an error when compiling project


emm312:
```
Caused by:
  process didn't exit successfully: `./Hexagn-rust/target/debug/build/hexagn-rust-923f982ba359502d/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', build.rs:4:29
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

build.rs
```rust
extern crate lalrpop;

fn main() {
    lalrpop::process_root().unwrap();
}
```

grammar file:
```
use crate::ast::nodes::*;
use std::str::FromStr;

grammar;

pub ProgramParser: Vec<Node> = {
    Block
}

Block: Vec<Node> = {
    Statement*
}

Statement: Node = {
    ExprNode,
    FuncDef
}

NonBoxedExpr: Expr = {
    FullExpr => *<>
}

ExprNode: Node = {
    <expr: FullExpr> ";" => Node::ExprNode(*expr)
}

FullExpr: Box<Expr> = {
    Expr,
    Comparison
}

Expr: Box<Expr> = {
    Expr ExprOp Factor => Box::new(Expr::BiOpNode(<>)),
    Factor
}

Comparison: Box<Expr> = {
    <lhs: Expr> <op: ComparisonOp> <rhs: Expr> => Box::new(Expr::ComparisonNode(lhs, op, rhs)),
}

ComparisonOp: Comparison = {
    "==" => Comparison::EQ,
    "!=" => Comparison::NEQ,
    ">" => Comparison::GT,
    "<" => Comparison::LT,
    ">=" => Comparison::GTE,
    "<=" => Comparison::LTE
}

ExprOp: Operation = {
    "+" => Operation::Add,
    "-" => Operation::Sub
}

Factor: Box<Expr> = {
    <lhs: Factor> <op: FactorOp> <rhs: Term> => Box::new(Expr::BiOp(lhs, op, rhs)),
    Term
}

FactorOp: Operation = {
    "*" => Operation::Mul,
    "/" => Operation::Div,
    "%" => Operation::Mod
}

Term: Box<Expr> = {
    num => Box::new(Expr::Number(<>)),
    ident => Box::new(Expr::Ident(<>)),
    FuncCall,
    string => Box::new(Expr::Str(<>)),
    "(" <Expr> ")"
}

[...]

Issue #699, opened 2023-01-11
Compatibility behavior of changing the out_dir is incredibly confusing 


not-my-profile:
I was just getting incredibly confused why `lalrpop::Configuration().set_out_dir("out").process_current_dir()` wrote the `.rs` file to the current directory instead of the `out/` directory. I looked through [the documentation](https://docs.rs/lalrpop/0.19.8/lalrpop/index.html) and couldn't find any answer, I looked through the [`api/mod.rs` source code](https://docs.rs/lalrpop/0.19.8/src/lalrpop/api/mod.rs.html) and couldn't find any answer. Only when digging deeper into the source code I found the following:

https://github.com/lalrpop/lalrpop/blob/b777a6a884052d6283971f71b3ba453aa6d417ab/lalrpop/src/build/mod.rs#L91-L96

And `process_current_dir` apparently sets the in_dir to `./src` triggering this "compatibility" behavior.

This is so incredibly unintuitive that I think that behavior should either be removed or the `Configuration` API be deprecated in favor of something more sane.

youknowone:
@nikomatsakis do you have any idea about this issue?

nikomatsakis:
I don't remember much of the details here but I definitely think we should clean this up for 1.0. I was trying it out on a recent project and I also hit some issues with a `.lalrpop` file that was not in the root directory, I think the problem was more that I didn't know what to specify as the path in the `lalrpop_mod` macro, though.

Issue #690, opened 2022-10-12
`lalrpop_mod!` looks under build subdirectory instead of crate root


littlebenlittle:
Running `cargo build` results in `couldn't read target/debug/build/xmd-lalrpop-<id>/out/src/calculator1.lalrpop: no such file or directory`.

I can get the build script to finish if I use `lalrpop_mod!(pub calculator1, "/../../../../../src/calculator1.lalrpop")` but that really doesn't look like the intended way to use the macro.

Using `cargo -V` `cargo 1.60.0 (d1fd9fe 2022-03-01)`

Project layout based on calculator example example:

```
.
├── build.rs
├── Cargo.lock
├── Cargo.toml
└── src
   ├── lib.rs
   └── calculator1.lalrpop
```

src/lib.rs:

```rs
#[macro_use] extern crate lalrpop_util;

lalrpop_mod!(pub calculator1, "/src/calculator1.lalrpop");
```

marcospb19:
I just had the same error while following the tutorial.

an-inconspicuous-semicolon:
I found the reason for this. It is because in the macro `lalrpop_mod!` it looks for the file in `env!("OUT_DIR")/argument-given-to-macro`

The issue comes with the fact that cargo sets `OUT_DIR` on build. so when `OUT_DIR=foobar cargo build` is run, OUT_DIR gets set to the binary output dir, rather than `foobar`

the fix for this is to change the enviroment variable used inside `lalrpop_mod!` to something like `LALRPOP_OUT_DIR`

an-inconspicuous-semicolon:
Also note that this is expected behavior as of `v0.16.0`
[documentation](https://lalrpop.github.io/lalrpop/generate_in_source.html)

Issue #677, opened 2022-07-24
Surprising behavior of `Configuration`


martin-t:
- The build.rs recommended in the book doesn't call `emit_rerun_directives(true)`, thus wasting time recompiling the final binary after even completely unrelated changes in the project dir. It would be nice to document this behavior. If there's a downside to generating these rerun directives, it should also be documented.
- `process_current_dir` generates output in `./debug/build/fml-aac4247c81d0131d/out/fml.rs` (in my project). However, attempting to avoid rescanning the whole project by using `.process_dir("src")` causes the generated path to be `./debug/build/fml-aac4247c81d0131d/out/src/fml.rs`. This is surprising and there's no documented reason for this change. (It also leads to hard-to-debug issues because the original file might still exist so new changes to the grammar will have no effect.)
- `process_dir` seems to do a lot of extra stuff compared to `process_file` such as setting the output dir, which (again) is surprising because the difference is not documented anywhere.

PR #825, opened 2023-08-24
LALRPOP Configuration fixes

Pat-Lafon:
Hello. I would like to start addressing some of the issues raised around lalrpop's Configuration settings for setting paths and such. There are a couple relevant issues that I think should be address on the way to a potential 1.0(since they may involve/give the chance to making breaking changes).

Maybe there is a more radical change with configurations that should be made, but I don't have the high level overview to  make that call.

Here are the following issues I would like to address eventually, I'm open to suggestions/feedback. #545 #611 #677 #690 #699 #710 (maybe #539)

So far, this pr adds a little bit of documentation for the rerun directives and looks at identifying ill named Rust modules from file names. The latter would be a warning of the form `eprintln` except that cargo will suppress these as long as the build doesn't fail. I think having spaces in the lalrpop file name is rare enough to just raise an error(a potentially breaking change). It does seem that if lalrpop was a proc_macro, it could leverage some of the nightly diagnostics features instead for this purpose.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -96,7 +96,13 @@ fn gen_resolve_file(session: &Session, lalrpop_file: &Path, ext: &str) -> io::Re
         .unwrap()
         .contains(char::is_whitespace)
     {
-        eprintln!("The lalrpop file named `{}` contains whitespace which may make this file difficult to import.", lalrpop_file.display())
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!(
+                "LALRPOP file names cannot contain whitespace: {}",
+                lalrpop_file.display()
+            ),
+        ));
     }
 
     // If the lalrpop file is not in in_dir, the result is that the
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -91,9 +91,21 @@ fn gen_resolve_file(session: &Session, lalrpop_file: &Path, ext: &str) -> io::Re
     // But I don't think we want a full blown syn dependency unless fully converting to proc macros.
     if lalrpop_file
         .file_name()
-        .unwrap()
+        .ok_or(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!(
+                "LALRPOP could not extract a valid file name: {}",
+                lalrpop_file.display()
+            ),
+        ))?
         .to_str()
-        .unwrap()
+        .ok_or(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!(
+                "LALRPOP file names must be valid UTF-8: {}",
+                lalrpop_file.display()
+            ),
+        ))?
         .contains(char::is_whitespace)
     {
         return Err(io::Error::new(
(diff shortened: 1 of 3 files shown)


lalrpop/src/build/mod.rs
```diff
@@ -87,6 +87,24 @@ fn gen_resolve_file(session: &Session, lalrpop_file: &Path, ext: &str) -> io::Re
         in_dir
     };
 
+    // Ideally we do something like syn::parse_str::<syn::Ident>(lalrpop_file.file_name())?;
+    // But I don't think we want a full blown syn dependency unless fully converting to proc macros.
+    if lalrpop_file
+        .file_name()
+        .unwrap()
+        .to_str()
+        .unwrap()
```
dburgener, line 96: This unwrap panics if the filename contains invalid unicode.  I feel like adding a check for one sort of weird filename that introduces a panic on a different sort feels weird to me.  Maybe `unwrap_or_else("")` to skip the check if the conversion fails would be cleaner?

Pat-Lafon: Hmm, I certainly could but maybe it's not a bad thing that it will panic on invalid unicode? I don't think there is an expectation that any of this would work in that case... what name would the module have that the rust code would call? How would you supply the path to lalrpop_mod if it isn't a valid string?

dburgener: I tend to take the angle that if the failure is purely our internal bug, panicing is fine.  If it could possibly be triggered by external state, we should error instead.  So in my opinion erroring on invalid unicode instead of panicing would be better.

dburgener: To elaborate, maybe someone's build script wants to log and error and proceed to other tasks if lalrpop has an issue.  Returning an error instead of panicing gives them that option.

Pat-Lafon: I've replaced both of the panics with io Errors

lalrpop/src/build/mod.rs
```diff
@@ -87,6 +87,36 @@ fn gen_resolve_file(session: &Session, lalrpop_file: &Path, ext: &str) -> io::Re
...
+        .to_str()
+        .ok_or(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!(
+                "LALRPOP file names must be valid UTF-8: {}",
+                lalrpop_file.display()
+            ),
+        ))?
+        .contains(char::is_whitespace)
+    {
```
youknowone, line 110:
```suggestion
    let file_name = if let Ok(lalrpop_file) else {
        return Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            format!(
                "LALRPOP could not extract a valid file name: {}",
                lalrpop_file.display()
            )
        );
    };
    let file_name = if let Ok(file_name) = file_name.to_str() else {
        return Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            format!(
                "LALRPOP file names must be valid UTF-8: {}",
                lalrpop_file.display()
            ),
        ));
    };
    if file_name.contains(char::is_whitespace) {
```
how about this style to separate each validation steps?

dburgener: Just my two cents, but isn't the `ok_or()` style more idiomatic rust?  I personally find the version @Pat-Lafon has written to be more readable...

PR #826, opened 2023-08-24
Mention the requirement that state parameters implement Copy in the tutorial

### LINKED ISSUE #689: state requires copy trait
As per the instructions: https://lalrpop.github.io/lalrpop/tutorial/009_state_parameter.html

If I try to pass a state parameter that does not implement the copy trait, I get the error:

```
   Compiling robotica-node-rust v0.1.0 (/home/brian/tree/personal/robotica-node-rust)
error[E0507]: cannot move out of `self.context` which is behind a mutable reference
   --> /home/brian/tree/personal/robotica-node-rust/target/debug/build/robotica-node-rust-a07a1b4494bd2ade/out/scheduling/conditions.rs:399:17
    |
399 |                 self.context,
    |                 ^^^^^^^^^^^^ move occurs because `self.context` has type `Fields<T>`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0507`.
error: could not compile `robotica-node-rust` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `robotica-node-rust` due to previous error
```

I am trying to pass a type that is incompatible with the copy trait (clone would be OK however).

This happens in the reduce function:

```
        fn reduce(
            &mut self,
            action: i8,
            start_location: Option<&Self::Location>,
            states: &mut alloc::vec::Vec<i8>,
            symbols: &mut alloc::vec::Vec<__state_machine::SymbolTriple<Self>>,
        ) -> Option<__state_machine::ParseResult<Self>> {
            __reduce(
                self.context,
                self.input,
                action,
                start_location,
                states,
                symbols,
                core::marker::PhantomData::<(&(), T)>,
            )
        }
```

I am wondering if this parameter can be passed as a reference, so the copy is not required.

dburgener:
closes #689

### DIFF
--- a/doc/src/tutorial/009_state_parameter.md
+++ b/doc/src/tutorial/009_state_parameter.md
@@ -33,6 +33,8 @@ fn calculator8() {
 
 For a more practical example with a custom tree structure, check out [this parser][expr_arena] using [this structure][expr_arena_ast] to build the AST.
 
+Note: The state parameter must implement the Copy trait.  For types that don't implement Copy, you should pass them as a reference instead.
+
 
 [expr_arena]: https://github.com/lalrpop/lalrpop/blob/master/lalrpop-test/src/expr_arena.lalrpop
 [expr_arena_ast]: https://github.com/lalrpop/lalrpop/blob/master/lalrpop-test/src/expr_arena_ast.rs


youknowone (approved):
👍

PR #833, opened 2023-09-22
Bump petgraph to 0.6.4

nilehmann:
Petgraph 0.6.4 depends on indexmap v2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,6 +101,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "equivalent"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
+
+[[package]]
 name = "fixedbitset"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


dburgener:
I'm confused by this.

1. petgraph 0.6.4 is semver compatible with 0.6.  Which means we get 0.6.4 automatically via "cargo update" without needing to change Cargo.toml
2. It's not clear to me why lalrpop has any particular need for 0.6.4.

Pat-Lafon:
My expectation is that this would fail the min_version workflow action unless there was some breaking change.

dburgener:
> My expectation is that this would fail the min_version workflow action unless there was some breaking change.

Would it?  As I understand it the min version workflow doesn't tell us if we've overspecified versions in our Config.  It tries the lowest possible versions from our config and makes sure we still build.  So this change would bump the version of petgraph tested by min_version from 0.6.0 to 0.6.4, both of which I think should succeed.  The regular workflows would either test the locked version (0.6.3) or the latest (0.6.4) for the "cargo update" version.

My guess is that what the submitter really wants is just to get 0.6.4 in a lalrpop binary build, in which case just updating the Cargo.lock would be sufficient.

nilehmann:
Sorry for the noise. Now with a clear mind I see that the bump doesn't make any sense for what I want

Feel free to close this.

PR #840, opened 2023-10-10
Grammar file block comment support

### LINKED ISSUE #370: Allow multiline comments in grammar files
Is there any particular reason why `lalrpop` doesn't support multiline comments (`/* ... */`)? I just noticed it when trying to put a copy of my reference grammar in my `*.lalrpop` file.

For Example:

```rust
use ast::{Atom, BinOp, BinaryOp, Expr};

/*
Original Grammar
================

Expressions
-----------

Expression         ::= XOR_Expression {'OR' XOR_Expression}
XOR_Expression     ::= AND_Expression {'XOR' AND_Expression}
   ...
ST_FunctionInput   ::= [VariableName ':='] Expression
*/

grammar<'input>;

...
```

seanbright:
Add support for block comments including nesting.

Fixes #370

### DIFF
--- a/lalrpop/src/tok/mod.rs
+++ b/lalrpop/src/tok/mod.rs
@@ -26,6 +26,7 @@ pub enum ErrorCode {
     UnterminatedAttribute,
     UnterminatedCode,
     ExpectedStringLiteral,
+    UnterminatedBlockComment,
 }
 
 fn error<T>(c: ErrorCode, l: usize) -> Result<T, Error> {
@@ -353,6 +354,13 @@ impl<'input> Tokenizer<'input> {
                         self.take_until(|c| c == '\n');
                         continue;
                     }
+                    Some((_, '*')) => {
+                        self.bump(); // Skip over the *
+                        match self.block_comment(idx0) {
+                            Err(err) => Some(Err(err)),
+                            _ => continue,
+                        }
+                    }
                     _ => Some(error(UnrecognizedToken, idx0)),
                 },
                 Some((idx0, '-')) => match self.bump() {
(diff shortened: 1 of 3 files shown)


seanbright:
Apologies for the `rustfmt` issues. _Should_ be good now.

lalrpop/src/tok/mod.rs
```diff
@@ -564,6 +580,49 @@ impl<'input> Tokenizer<'input> {
...
+                }
+                State::Star if c == '/' => {
+                    depth -= 1;
+                    if depth == 0 {
+                        State::Complete
+                    } else {
+                        State::Initial
+                    }
+                }
+                _ => State::Initial,
```
Pat-Lafon, line 611: If you are in the star state, and see another star, shouldn't you stay in the star state?

Consider these three examples which are valid rust comments. Maybe elevate at least into a test.

```rust    
/***/
/*****/
/* **/
```

Pat-Lafon: Not that we need to be fully compatible, but here is the grammar for comments: https://doc.rust-lang.org/reference/comments.html

dburgener: Along the same lines, shouldn't you stay in the Slash state if you're in the Slash state and see another slash?

I believe that `/*//**/*/` is a valid comment that I don't think this would parse correctly.

I wonder if the whole thing would be a little cleaner and more idiomatic as something like:

```
match (state, c) {
    (State::Initial, '/') => State::Slash
    // etc
```

dburgener: I *think* with the above suggestion, you can write the final clause as `(_,c) if c != '/' && c != '*''`, and get a compiler guarantee that all the '/' and '*' cases are covered

seanbright: I was not able to get this to compile without a warning, so I reverted to just using the `_` match.

dburgener (approved):

dburgener:
This looks good to me, thanks!

I vote for delaying merge until after we finish up the 0.20.1 release to avoid additional churn there, since we're (hopefully) really close to getting that finished up but if we'd rather get it in for 0.20.1, that's fine.

dburgener:
I'm going to go ahead and merge this, even though the 0.20.1 release was yanked.  No reason to put it off IMO.

dburgener:
Merged.  Thanks for the contribution!

PR #829, opened 2023-08-31
Better default skip rule

QuarticCat:
Extract skip rule related updates from #814 as requested by @youknowone.

PR #841, opened 2023-10-10
add 'unicode' to tutorial import

IsakNyberg:
As addressed in issue #750 the regex breaks the build. In this case the current [tutorial](https://lalrpop.github.io/lalrpop/tutorial/001_adding_lalrpop.html) does not import unicode which causes the very first example form the tutorial to fail with the error. 

```
called 'Result::unwrap()' on an 'Err' value: Syntax(
regex parse error:
    ^(\s*)
      ^^
)
```

This is all because the unicode ~crate~ feature is not imported in the cargo.toml, (like it is in the [source file](https://github.com/lalrpop/lalrpop/blob/master/doc/calculator/Cargo.toml)) adding is there fixes the issue.

### DIFF
--- a/doc/src/tutorial/001_adding_lalrpop.md
+++ b/doc/src/tutorial/001_adding_lalrpop.md
@@ -34,7 +34,7 @@ edition = "2021"
 lalrpop = "0.20.0"
 
 [dependencies]
-lalrpop-util = { version = "0.20.0", features = ["lexer"] }
+lalrpop-util = { version = "0.20.0", features = ["lexer", "unicode"] }
 ```
 
 Cargo can run [build scripts] as a pre-processing step,


dburgener (approved):

dburgener:
This looks good.  A minor note is that it's the unicode "feature", not "crate" that this PR enables, but otherwise seems reasonable.

I'll merge once tests finish.

dburgener:
Merged, thanks for the contribution!

Pat-Lafon:
Isn't this not an issue on the master branch because of https://github.com/lalrpop/lalrpop/pull/829/files? At least testing locally by setting 
```
lalrpop-util = { version = "0.20.0", features = [
    "lexer",
], git = "https://github.com/lalrpop/lalrpop.git" }
```
on my project worked. Just want to check if this documentation won't just become obsolete when the next version is released.

dburgener:
Ooh, I think you're right.  I somehow got under the mistaken impression that the tutorial explicitly used a `\s` somewhere, but yeah, I don't see it.

My gut reaction is that it's probably best to leave the online docs in a working state compared to the latest published version.

Ideally, we should publish 0.20.1 ASAP.  #829 and the performance improvements are both pretty big deal changes to get out.  If we're ready for 0.20.1, we could do a revert on this right before we publish 0.20.1 to keep the docs accurate.

I imagine this means we can drop the unicode feature from the calculator example though, which would be awesome.  It's pretty painful to dance around feature unification trying to do builds without unicode support.

dburgener:
I take it back.  I tried dropping the unicode feature in the calculator and it failed.

It turned out calculator2b in the lexer precendence example uses unicode: https://lalrpop.github.io/lalrpop/lexer_tutorial/001_lexer_gen.html

Failing calculator PR run: https://github.com/dburgener/lalrpop/actions/runs/6473643550/job/17576878751?pr=5

Pat-Lafon:
In some sense it doesn't need to, it could use `r"[[:space:]]+"` instead I think.

Pat-Lafon:
Hmmm, we might need to be careful with this kind of rewriting because for better or worse this is the behavior of regex. I can see the ergonomic benefit, but it's worth saying that currently if people have these kinds of issues then they can go to regex for documentation. This would put more documentation burden on lalrpop to explain the parts of a `r""` that might get silently rewritten.

Without knowing your approach, I would lean towards trying to provide better error messages in this case to the user(Suggest either using the ascii-only version or in enabling the unicode feature).

dburgener:
I think that's very fair.  I'm hopefully close on the PR, so I'll plan on putting it up as an RFC and we can discuss if it's something we want.

PR #842, opened 2023-10-16
Release 0.20.1

Pat-Lafon:
This is an attempt to "kick-the-tires" on a new release based on https://github.com/orgs/lalrpop/discussions/823.

I did an initial draft of the release notes in `RELEASES.md` and ran `./version.sh` to update the version number in all the places.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,22 @@
+<a name="0.20.1"></a>
+## 0.20.1 (2023-10-**)
+
+#### Features
+
+* Dangling symlinks in crate no longer cause build failure (thanks to legeana for the report!)
+* Lalrpop now uses the ascii-aware space regex when the unicode feature is not enabled (thanks to QuarticCat!)
+* Lalrpop no longer depends on the `is-terminal` crate (thanks to Kmeakin!)
+* Better performance with the default lexer using the underlying `regex-automata` crate (thanks to QuarticCat!)
+* Allow the catch-all `_` case for token matching can now be set to a higher precedence in match (thanks to fpoli!)
+* Fewer clippy lints triggered in generated code
+
+#### Compatibility note
+
+* MSRV increased to `1.70` in accordance to the MSRV policy.
+* `process_root_unconditionally` now correctly lints as having been deprecated.
+* Internal types which lead with a `__` and should not be relied upon are no longer publicly exposed (thanks to arnaudgolfouse!)
+* Lalrpop files containing a space in their name now return an error.
+
 <a name="0.20.0"></a>
 ## 0.20.0 (2023-05-02)
 
@@ -67,7 +86,7 @@ This release fixes an incompatibility with regex.
 
 This release addresses backwards compatibility warnings and security alerts, along with a few smaller features.
 
-Special thanks to our newest maintainers, Yann Hamdaoui and YunWon Jeong for helping to coordinate this release. 
+Special thanks to our newest maintainers, Yann Hamdaoui and YunWon Jeong for helping to coordinate this release.
 
 #### Features
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -3,16 +3,19 @@
 
 #### Features
 
-* Dangling symlinks in crate no longer cause build failure (thanks to legeana for the report!)
-* Lalrpop now uses the ascii-aware space regex when the unicode feature is not enabled (thanks to QuarticCat!)
 * Lalrpop no longer depends on the `is-terminal` crate (thanks to Kmeakin!)
 * Better performance with the default lexer using the underlying `regex-automata` crate (thanks to QuarticCat!)
 * Allow the catch-all `_` case for token matching can now be set to a higher precedence in match (thanks to fpoli!)
 * Fewer clippy lints triggered in generated code
 
+#### Bugfixes
+
+* Lalrpop now uses the ascii-aware space regex when the unicode feature is not enabled (thanks to QuarticCat!)
+* Dangling symlinks in crate no longer cause build failure (thanks to legeana for the report!)
+
 #### Compatibility note
 
-* MSRV increased to `1.70` in accordance to the MSRV policy.
+* MSRV increased to `1.70`.
 * `process_root_unconditionally` now correctly lints as having been deprecated.
 * Internal types which lead with a `__` and should not be relied upon are no longer publicly exposed (thanks to arnaudgolfouse!)
 * Lalrpop files containing a space in their name now return an error.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,8 @@
 <a name="0.20.1"></a>
 ## 0.20.1 (2023-10-**)
 
+Special thanks to our newest maintainers, Daniel Burgener and Patrick LaFontaine for helping to coordinate this release.
+
 #### Features
 
 * Lalrpop no longer depends on the `is-terminal` crate (thanks to Kmeakin!)
--- a/doc/calculator/Cargo.toml
+++ b/doc/calculator/Cargo.toml
@@ -1,12 +1,12 @@
 [package]
 name = "calculator"
-version = "0.20.0"
+version = "0.20.1"
 authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 workspace = "../.." # <-- We added this and everything after!
 edition = "2021"
 
 [build-dependencies]
-lalrpop = { version = "0.20.0", path = "../../lalrpop" }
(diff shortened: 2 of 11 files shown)


RELEASES.md
```diff
@@ -1,3 +1,22 @@
+<a name="0.20.1"></a>
+## 0.20.1 (2023-10-**)
+
+#### Features
+
+* Dangling symlinks in crate no longer cause build failure (thanks to legeana for the report!)
```
dburgener, line 6: I might call this and the ascii space regex "bugfixes" rather than features.

RELEASES.md
```diff
@@ -1,3 +1,22 @@
...
+* Lalrpop now uses the ascii-aware space regex when the unicode feature is not enabled (thanks to QuarticCat!)
+* Lalrpop no longer depends on the `is-terminal` crate (thanks to Kmeakin!)
+* Better performance with the default lexer using the underlying `regex-automata` crate (thanks to QuarticCat!)
+* Allow the catch-all `_` case for token matching can now be set to a higher precedence in match (thanks to fpoli!)
+* Fewer clippy lints triggered in generated code
+
+#### Compatibility note
+
+* MSRV increased to `1.70` in accordance to the MSRV policy.
```
dburgener, line 15: The phrase "in accordance to the MSRV policy" makes it sound like the increase is due to some policy rather than the is_terminal dependency requiring it.  Maybe just end at "MSRV increased to 1.70"?

dburgener (commented):
Thanks for putting this together!  I think it looks mostly good from my perspective.  Just a couple of minor notes on the changelog.

dburgener:
The notes look good, however I gave it one final test, and I'm seeing problems.  If I build using the commit before #814, everything works, but #814 without unicode support is translating my regexes containing `^` into an enumeration of all the possible unicode characters except excluded ones, and then panicing at run time upon encountering the unicode.  I'm looking into a root cause and fix, but I'm not sure if I'll be able to get to it tonight.

dburgener:
I'm out of time for tonight unfortunately.  I'm getting weird failures both with and without unicode when trying to run tests in https://github.com/dburgener/cascade against lalrpop e2eb361cf507d5607f66c6b7ef4262d65770712d, but not against e2eb361cf507d5607f66c6b7ef4262d65770712d^.

When I run without unicode, I get the unexpected unicode on `^` patterns I mentioned above.   When I run with unicode, a few of my error tests for non-parsing input don't terminate in a reasonable amount of time.

I'm trying to get a simple reproducer that's more pure-lalrpop to narrow down what's wrong, but I haven't been successful at that yet.  It's not clear to me at this point why my code is failing but the lalrpop tests aren't.  Perhaps its user error on my end.  I'll take a look in the morning and follow up then.

doc/calculator/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "calculator"
-version = "0.20.0"
+version = "0.20.1"
```
yannham, line 3: Can't we use something like:

```
version.workspace = true
```

if this crate always follows the version of the main package of the workspace? This would save future version bumps.

doc/calculator/Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
 workspace = "../.." # <-- We added this and everything after!
 edition = "2021"
 
 [build-dependencies]
-lalrpop = { version = "0.20.0", path = "../../lalrpop" }
+lalrpop = { version = "0.20.1", path = "../../lalrpop" }
 
 [dependencies]
-lalrpop-util = { version = "0.20.0", path = "../../lalrpop-util", features = ["lexer", "unicode"] }
+lalrpop-util = { version = "0.20.1", path = "../../lalrpop-util", features = ["lexer", "unicode"] }
```
yannham, line 12: Similarly, can we do something like:

```
lalrpop-util = { workspace = true, features = [ "lexer", "unicode" ] }
```

here, and put in the main `Cargo.toml` something like:

```
lalrpop-util = { version = "0.20.1", path = "./lalrpop-util"}
```

Sorry, it might be out of scope of this PR, but I've been doing that in other projects (define workspace-level dependencies for workspace members, and a workspace-level version, and use that everywhere else so that you only have to bump each member's, and only if it doesn't inherit the workspace's main version) and I wonder if it could make the workflow simpler for LALRPOP. Feel free to tell me that we can see that later :slightly_smiling_face:

version.sh
```diff
@@ -3,6 +3,8 @@
 # A script to bump the version number on all Cargo.toml files etc in
 # an atomic fashion.
 
+set -e -o pipefail
```
yannham, line 6: Ah, I only see this script now. So maybe it's not really a hassle if it's automatized, although the scripts feel a bit ad-hoc.

yannham (commented):
LGTM, beside side suggestions. I don't approve to wait for @dburgener to sort his tests out before potentially merging this.

doc/calculator/Cargo.toml
Pat-Lafon: That's a fair point. It doesn't seem like any of these examples get rendered in the lalrpop book so it seems fine.

Actually on that point, maybe we should be pointing to these examples in the book instead of the tests when possible(referring to http://lalrpop.github.io/lalrpop/tutorial/index.html). Otherwise, I'm not sure how much visibility some of these get.

version.sh
Pat-Lafon: It is somewhat automated, in a kinda hacky way. I think we could do a little bit better here in terms of setting up the release process steps. Not sure what big crates do in this regard. Maybe something that includes a mini-crater run or cargo-semvar-checks to help catch any last minute issues.

doc/calculator/Cargo.toml
Pat-Lafon: I've spawned this off to it's own pr to be discussed/merged separately.

doc/calculator/Cargo.toml
Pat-Lafon: See #843

dburgener:
I *think* I have a solution for the issues I'm seeing.  I haven't tested it end to end yet, but limited testing I've done looks promising.

I believe the problem is that even before #814, when we use regex-syntax to parse our regexes, we didn't correctly disable unicode on non-unicode builds, which resulted in unicode regexes ending up in the generated parser.  Prior to #814 this seems to have not mattered, I *think* because internals of regex compile the unicode away anyways (based on [this](https://github.com/rust-lang/regex/issues/1088#issuecomment-1711727185) reply to a similar, but different issue.  The issue there was a bug in regex-syntax, but this is a bug in our code, that happens to hit a similar symptom).

Anyways, I'm going to tidy up my fix and do end to end testing.  If all goes well, I'll have a PR up in a bit.

dburgener:
Oh, the other thing is that I'm still kind of mystified as to why our existing lalrpop tests didn't catch this at all.  I'm going to prioritize getting the fix in to unblock the release, but then I'll try to loop back and identify the testing gap.

dburgener:
Okay, the second issue was user error in the sense of "why on earth would you do that, don't do that", and lalrpop error in the sense of "we should handle this bad input more sanely".

I had a skip pattern that could match zero length regexes.  #814 seems to have inadvertently dropped a check that returned "InvalidToken" if our longest match was a zero length skip pattern.

Note that the lalrpop book does recommend possibly zero length whitespace matches: https://lalrpop.github.io/lalrpop/lexer_tutorial/001_lexer_gen.html#customizing-skipping-between-tokens

I think the fix should be to restore the original check to return InvalidToken on a zero length skip match.  I'll submit a PR for that as well.

Pat-Lafon:
All this sounds good to me! Thanks for the detailed look.

dburgener:
https://github.com/lalrpop/lalrpop/pull/845 is up

Everything seems good as far as I can tell.  Once we get those merged, I'll retest.

youknowone:
I'd like to add credits for new maintainers, who contributed a lot not only for this version but for longer time.

youknowone:
is #844 the last piece of new release?

dburgener:
> is https://github.com/lalrpop/lalrpop/pull/844 the last piece of new release?

That plus this PR of course is all I'm tracking on.  It looks like everyone is okay with #844, so I'll go ahead and merge that later today.

dburgener:
#844 is merged.

I ran the same tests I did previous that uncovered the bugs against the latest master, and they passed this time.  This seems ready to go from a functional perspective in my opinion.

youknowone (approved):
Released 0.20.1 :+1:

youknowone:
https://crates.io/crates/lalrpop/0.20.1

PR #843, opened 2023-10-17
Include a pointer from the tutorial to a collection of examples

Pat-Lafon:
Following up on https://github.com/lalrpop/lalrpop/pull/842#discussion_r1362300955

### DIFF
--- a/doc/src/tutorial/index.md
+++ b/doc/src/tutorial/index.md
@@ -13,8 +13,9 @@ first.
 - [Error recovery](008_error_recovery.md)
 - [Passing state parameter](009_state_parameter.md)
 
-This tutorial is still incomplete. Here are some topics that I aim to
-cover when I get time to write about them:
+This tutorial is still incomplete. There are a few example grammars in the [doc section] of the repository.
+
+Here are some topics that I aim to cover when I get time to write about them:
 
 - Advice for resolving shift-reduce and reduce-reduce conflicts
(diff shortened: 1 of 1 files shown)


dburgener:
I'm hesitant on this.  The point of the doc content is for tutorials.  The tutorial is telling you how to use lalrpop in your own project, including Cargo.toml changes.  So to some extent, they are intended to function as examples for the users, rather than just maintaining our build system.  To that end, referencing workspaces, both complicates things for the new(er) rust user, and is likely not relevant to the user's environment.

A counter-argument could be that version and edition are not at all the salient factors here, and users will be managing their own of those independent of lalrpop.

yannham:
After taking a second look at this, I think I agree with @dburgener. I didn't realize those were example crates from the doc, and not proper subcrates of lalrpop. Also, I guess their version has no reason to be following LALRPOP (and `nobol` was indeed `0.1.0`). Sorry for the misleading suggestion.

dburgener:
If we do abandon the workspacification, let's grab 72c21035bd98e904282d53808b1fcd805f36ff7b by itself though.  That seems like a good change.

youknowone:
I agree that examples under `doc` is tutorial. I think we had similar discussion couples weeks ago. Leaving a comment on doc's Cargo.toml would be helpful

Pat-Lafon:
I guess I didn't remember the previous discussion on this. I think it still makes sense because none of these `/doc/` examples are drop in as they are. They load lalrpop via a relative path instead of via crates.io. I viewed them more as examples of grammar files(since there is a quickstart tutorial for how to set up your crate for lalrpop).

I'm not too attached to this pr though so we can just cherry-pick the second commit to raise awareness to these examples.

youknowone (approved):

PR #844, opened 2023-10-17
Only generate unicode tokens if the unicode feature is enabled

dburgener:
regex_syntax::parse() converts our regex strings into the HIR of the regex, part of that includes unpacking various metacharacters into a list of symbols.  In many cases, this expansion changes depending if it should expand into unicode or not.

Prior to #814, we were still outputting unicode regexes unconditionally, but regex internals seem to be compiling them away and avoiding errors. The switch in #814 caused these to be result in real errors.

Follow-up work will be needed to determine why existing tests didn't detect this.

### DIFF
--- a/lalrpop/src/lexer/re/mod.rs
+++ b/lalrpop/src/lexer/re/mod.rs
@@ -1,7 +1,7 @@
 //! A parser and representation of regular expressions.
 
 use regex_syntax::hir::Hir;
-use regex_syntax::{self, Error, Parser};
+use regex_syntax::{self, Error, ParserBuilder};
 
 #[cfg(test)]
 mod test;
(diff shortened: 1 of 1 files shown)


dburgener:
This fixes part of the issues I'm seeing.  With this change I no longer see "invalid unicode" errors when building without unicode.

However, I'm still seeing a couple of tests for parse failures running for a long time (regardless of the unicode feature).

I think this fix is independently correct/valuable, but I still need to investigate the second issue.

lalrpop/src/lexer/re/mod.rs
```diff
@@ -19,6 +19,11 @@ pub fn parse_literal(s: &str) -> Regex {
 
 /// Parse a regular expression like `a+` etc.
 pub fn parse_regex(s: &str) -> Result<Regex, RegexError> {
-    let expr = Parser::new().parse(s)?;
+    let enable_unicode = cfg!(feature = "unicode");
+    let expr = ParserBuilder::new()
+        .utf8(enable_unicode)
```
yannham, line 24: Shouldn't we always require regexp to match only utf8 tokens? And even if we want to allow, should this option really be tied to unicode support?

yannham (approved):
I see, thanks for the explanation! It's more subtle that it first seemed. It does sound reasonable, then

Issue #848, opened 2023-10-23
Release 0.20.1 not SemVer compatible?


shaobo-he-aws:
Hello,

Our project [Cedar](https://github.com/cedar-policy/cedar) depends on `lalrpop-util`. Updating to 0.20.1 causes our code to panic.  To reproduce,
```
git clone https://github.com/cedar-policy/cedar.git
cd cedar/cedar-policy-cli/sample-data/sandbox_a
cargo run evaluate \                               
    --principal 'User::"alice"' \
    --action 'Action::"listPhotos"' \
    --resource 'Album::"jane_vacation"' \
    --entities entities.json \
   "resource in Account::\"jane\""
```

The error message is,

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: BuildError { kind: NFA(BuildError { kind: Syntax(Translate(Error { kind: UnicodeNotAllowed, pattern: "(?:\"((?:(?:\\\\[\0-\t\u{b}-\u{10ffff}])|[\0-!\\#-\\[\\]-\u{10ffff}]))*\")", span: Span(Position(o: 19, l: 1, c: 20), Position(o: 23, l: 1, c: 21)) })) }) }', /private/tmp/cedar/target/debug/build/cedar-policy-core-894cc3fee656509a/out/src/parser/grammar.rs:57330:76
```

We need to enable `unicode` feature of `lalrpop-util` to avoid the crash, which seems to be a breaking change.

dburgener:
Thanks for the report!  I guess we should yank the release, @youknowone?

Looking at the issue, I think I know what's going on here (this was initially very surprising to me, since I thought we got this exact issue sorted out before the release (#844)).

Checking your `cargo tree -e features`, I see that the `unicode` feature is being enabled in lalrpop, but not lalrpop-util.  That's happening because it's in the default features for lalrpop, but not lalrpop util.  We should definitely get consistent in setting those default features.  I think that had we done so, you would have picked up the default feature in util and this wouldn't have broken you.

I think that the state today is that lalrpop and lalrpop-util need to be consistent in terms of whether they use the unicode feature (I suspect 0.20.1 introduced this implicit requirement, but I'm not sure).  I'm not aware of if Cargo gives us a way to enforce that, but at least setting the defaults the same should avoid the pitfall.

[...]

PR #849, opened 2023-10-23
Set unicode as a default feature in lalrpop-util

dburgener:
We get breakage if lalrpop and lalrpop-util are not consistent, but the defaults aren't consistent

### DIFF
--- a/lalrpop-util/Cargo.toml
+++ b/lalrpop-util/Cargo.toml
@@ -17,7 +17,7 @@ regex-automata = { workspace = true, optional = true }
 lexer = ["regex-automata/std", "std"]
 unicode = ["regex-automata?/unicode"]
 std = []
-default = ["std"]
+default = ["std", "unicode"]
 
 [package.metadata.docs.rs]
 features = ["lexer"]
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -35,7 +35,7 @@ unicode-xid = { version = "0.2", default_features = false }
 # library, disable it in your project by setting default_features = false.
 pico-args = { version = "0.5", default_features = false, optional = true }
 
-lalrpop-util = { path = "../lalrpop-util", version = "0.20.0" }
+lalrpop-util = { path = "../lalrpop-util", version = "0.20.0", default_features = false }
 
 [dev-dependencies]
 diff = { workspace = true }
@@ -43,7 +43,7 @@ rand = "0.8"
 
 [features]
 default=["lexer", "unicode", "pico-args"]
-unicode = ["regex/unicode", "regex-syntax/unicode"]
+unicode = ["regex/unicode", "regex-syntax/unicode", "lalrpop-util/unicode"]
 lexer = ["lalrpop-util/lexer"]
 
 [package.metadata.docs.rs]


dburgener:
See #848 for this inconsistency causing user breakage.

dburgener:
Tested this against the reproducer in #848, and the panic is no longer seen with this change in place.

Pat-Lafon:
Hmm, should https://github.com/lalrpop/lalrpop/blob/0c3599bd79d2f29252c165a13ffd5f7a71ac4af0/lalrpop/Cargo.toml#L46 have `"lalrpop-util/unicode"` as well?

Should we have a list of some of the larger lalrpop users for which we can test for these kinds of breaks before a release?

Pat-Lafon:
Like https://crates.io/crates/lalrpop/reverse_dependencies ?

dburgener:
Oh, perfect.  I will try to take a look at that this week, but no promises, I've got a lot on my plate at work the next two weeks, so it might be a while until I get to it depending on how things shake out.

Pat-Lafon:
We can certainly spin that off to it's own issue. I don't want to clog up merging this commit.

yannham (approved):
LGTM, as far as I can tell. Reading [cargo's doc](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), it _sounds_ like feature unification will take care of packages depending on both `lalrpop` and `lalrpop-util` and enabling unicode only on `lalrpop`. I wonder what happens if a user enables unicode only for `lalrpop-util` though? I would say this combination is fine (in that, it won't fail to build/panic?)

lalrpop-util/Cargo.toml
```diff
@@ -17,7 +17,7 @@ regex-automata = { workspace = true, optional = true }
 lexer = ["regex-automata/std", "std"]
 unicode = ["regex-automata?/unicode"]
 std = []
-default = ["std"]
+default = ["std", "unicode"]
```
youknowone, line 20: Should `unicode` be default of `lalrpop-util` too? Because `lalrpop/unicode` now requires `lalrpop-util/unicode`, wouldn't this be enabled by default from the default `lalrpop/unicode` feature?

If we are going to add `unicode` here, we also have to add `no-default-feature` flag to `lalrpop`'s `lalrpop-util` dependency. Otherwise `lalrpop-util/unicode` will never be disabled from `lalrpop` crate.

yannham: Ah, good point. Maybe the other change of this PR, which is that `larlpop/unicode` implies `lalrpop-util/unicode`, is sufficient alone?

dburgener: Thinking through this, I'm wondering about the following situation:

Library crate `foo` is a user of lalrpop.  They have `lalrpop` in their build dependencies, and `lalrpop-util` in their dependencies, using default features for both.  Another crate `bar` uses `foo` as a dependency.

I believe the situation here if we have `lalrpop/unicode` imply `lalrpop-util/unicode`, but don't set `unicode` as a default feature in `lalrpop-util` is that due to feature unification, `foo` will have the `unicode` feature set on their local builds, but when `bar` builds, they end up doing a build that never brings in `lalrpop`, just `lalrpop-util`, since `lalrpop was only a build time dependency, and so `lalrpop-util` doesn't inherit the `unicode` feature in `bar`'s build, which could of course cause problems if `foo` relies on unicode (possibly unknowningly).

That seems to me to be an argument in favor of adding `no-default-feature` to `lalrpop` as per the second paragraph of @youknowone's comment, but keeping the default on `lalrpop-unicode` to avoid the feature unification issue I just described?

This is all theory based on how I think feature unification works.  If it would be helpful, I can set up a collection of crates with the above structure and verify that that's how it works in practice.

youknowone: Adding no-default-feature makes sense to me for that cases

dburgener: Pushed in line with that.

youknowone (approved):
👍

PR #852, opened 2023-11-06
Traverse into symlinks to dirs

mbid:
The lalrpop_files function traverses recursively into subdirectories of the provided root directory and returns all .lalrpop files. However, lalrpop_files used to skip symlinks to directories because it used the DirEntry::file_type method to detect directories, which doesn't follow symlinks. Since lalrpop_files does follow symlinks to files, this old behavior for symlinks was inconsistent and unintuitive.

With this commit, we're now using the Path::is_dir method, which does follow symlinks, so now lalrpop_files traverses also into symlinks to directories.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -211,7 +211,7 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
 
         let path = entry.path();
 
-        if file_type.is_dir() {
+        if path.is_dir() {
             result.extend(lalrpop_files(&path)?);
         }
 
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -205,6 +205,35 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
     }
 }
 
+/// Handles a [walkdir::Error] if the root cause is a dangling symlink.
+///
+/// Returns `Ok` if the error could be handled, otherwise returns `Err(err)`.
+fn handle_dangling_symlink_error(err: walkdir::Error) -> Result<(), walkdir::Error> {
+    let is_not_found = err.io_error().map(|io_err| io_err.kind()) == Some(io::ErrorKind::NotFound);
+    if !is_not_found {
+        return Err(err);
+    }
+
+    // As of now on Linux, this is the path of the symlink (not where it points to) in case of a
+    // dangling symlink:
+    let path = match err.path() {
+        Some(path) => path,
+        None => {
+            return Err(err);
+        }
+    };
+
+    if !path.is_symlink() {
+        return Err(err);
+    }
+
+    eprintln!(
+        "Warning: ignoring dangling/erroneous symlink {}",
+        path.display()
+    );
+    Ok(())
+}
+
 fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
     let mut result = vec![];
 
@@ -213,7 +242,14 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
         // Use deterministic ordering:
         .sort_by_file_name();
     for entry in walkdir {
-        let entry = entry?;
+        let entry = match entry {
+            Ok(entry) => entry,
+            Err(err) => {
+                handle_dangling_symlink_error(err)?;
+                continue;
+            }
+        };
+
         // `file_type` follows symlinks, so if `entry` points to a symlink to a file, then
         // `is_file` returns true.
         if !entry.file_type().is_file() {
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -211,9 +211,7 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
     let walkdir = WalkDir::new(root_dir)
         .follow_links(true)
         // Use deterministic ordering:
-        .sort_by_file_name()
-        // Prevent loops:
-        .same_file_system(true);
+        .sort_by_file_name();
     for entry in walkdir {
         let entry = entry?;
         // `file_type` follows symlinks, so if `entry` points to a symlink to a file, then
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -17,7 +17,9 @@ use crate::util::Sep;
 use itertools::Itertools;
 use lalrpop_util::ParseError;
 use tiny_keccak::{Hasher, Sha3};
+use walkdir::WalkDir;
 
+use std::ffi::OsStr;
 use std::fs;
 use std::io::{self, BufRead, IsTerminal, Read, Write};
 use std::path::{Path, PathBuf};
@@ -205,42 +207,29 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
 
 fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
     let mut result = vec![];
-    for entry in fs::read_dir(root_dir)? {
+
+    let walkdir = WalkDir::new(root_dir)
+        .follow_links(true)
+        // Use deterministic ordering:
+        .sort_by_file_name()
+        // Prevent loops:
+        .same_file_system(true);
+    for entry in walkdir {
         let entry = entry?;
-        let file_type = entry.file_type()?;
+        // `file_type` follows symlinks, so if `entry` points to a symlink to a file, then
+        // `is_file` returns true.
+        if !entry.file_type().is_file() {
+            continue;
+        }
 
         let path = entry.path();
-
-        if path.is_dir() {
-            result.extend(lalrpop_files(&path)?);
+        if path.extension() != Some(OsStr::new("lalrpop")) {
+            continue;
         }
 
-        let is_valid_symlink_file = || -> bool {
-            if !file_type.is_symlink() {
-                false
-            } else {
-                // Ensure all symlinks are resolved to a file
-                // Or ignore erroneous ones https://github.com/lalrpop/lalrpop/issues/808
-                fs::metadata(&path).map_or_else(
-                    |_| {
-                        eprintln!(
-                            "Warning: ignoring dangling/erroneous symlink {}",
-                            path.display()
-                        );
-                        false
-                    },
-                    |m| m.is_file(),
-                )
-            }
-        };
-
-        if (file_type.is_file() || is_valid_symlink_file())
-            && path.extension().is_some()
-            && path.extension().unwrap() == "lalrpop"
-        {
-            result.push(path);
-        }
+        result.push(PathBuf::from(path));
     }
+
     Ok(result)
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,6 +176,7 @@ dependencies = [
  "term",
  "tiny-keccak",
  "unicode-xid",
+ "walkdir",
 ]
 
 [[package]]
@@ -424,6 +425,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ffc183a10b4478d04cbbbfc96d0873219d962dd5accaff2ffbd4ceb7df837f4"
 
 [[package]]
+name = "same-file"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93fc1dc3aaa9bfed95e02e6eadabb4baf7e3078b0bd1b4d7b6b0b68378900502"
+dependencies = [
+ "winapi-util",
+]
+
+[[package]]
 name = "scopeguard"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


dburgener:
This seems reasonable to me.  @Pat-Lafon touched this same function pretty recently in #821 to address #808.  It looks to me as though this is a reasonable additional clean-up in that code path.  Would love it if @Pat-Lafon could take a look too, since he was just in this code.

Pat-Lafon:
This makes sense to me, I would like lalrpop to feel more consistent/uniform. What stops lalrpop from entering into an infinite loop if the symlinks accidentally form a cycle? 

If we need it, I think cargo uses some kind of `WalkDir` iterator to do a more controlled traversal of the directory structure.

mbid:
Thanks for having a look Daniel and Patrick!

@Pat-Lafon: I haven't tried, but I'm pretty sure we're not terminating if there's a loop in the file system "tree". The problem already exists before my patch I believe, since you can use bind mounts on at least linux to create a loop without symlinks. But yes, if we follow symlinks as well, then we'll encounter the issue a lot more often.

Regarding `WalkDir`: Perhaps you have the [`walkdir`](https://docs.rs/walkdir/2.4.0/walkdir/) crate in mind? From a cursory glance, I think it has loop detection for symlinks, but doesn't try to detect loops created by bind mounts (which is fair, since it's rather obscure and a lot of other things will break in that situation). I'd be happy to rewrite my patch to use that crate instead.

Btw, is there an obvious test case that I can adapt to check the behavior with symlinks?

Pat-Lafon:
> @Pat-Lafon: I haven't tried, but I'm pretty sure we're not terminating if there's a loop in the file system "tree". The problem already exists before my patch I believe, since you can use bind mounts on at least linux to create a loop without symlinks. But yes, if we follow symlinks as well, then we'll encounter the issue a lot more often.

Ah, fair point that there is always a bigger fish. If we are encouraging people to use symlinks as a way to better organize their lalrpop files then I think cyclical symlinks are in scope(gracefully failing if someone incorrectly sets up their symlink target). But sure, don't want to feature creep this pr too much since it works as intended in most cases.

> Regarding `WalkDir`: Perhaps you have the [`walkdir`](https://docs.rs/walkdir/2.4.0/walkdir/) crate in mind? From a cursory glance, I think it has loop detection for symlinks, but doesn't try to detect loops created by bind mounts (which is fair, since it's rather obscure and a lot of other things will break in that situation). I'd be happy to rewrite my patch to use that crate instead.

I believe so. If that looks like a cleaner abstraction for lalrpop to use instead of manually traversing the directory structure then that would be a pretty useful refactoring. The nice thing here is that if there are corner cases like bind mounts that don't properly work, then now it's up to `walkdir` to fix them instead of us.

> Btw, is there an obvious test case that I can adapt to check the behavior with symlinks?

Hmm, not sure if the others like @dburgener have opinions on this. I'm not sure that we want to clutter `lalrpop-test` (since then all the tests fail if something goes wrong). The `doc` examples could be appropriate in that you might actually want to use symlinks when you have multiple lalrpop files being compiled. Alternatively, we could take a dev-dependency on `tempfile` and test just `lalrpop_files` in isolation.

Maybe the `doc` examples path if we want to raise visibility to lalrpop's support of symlinked files/directories. The `tempfile` path if we don't want to add unnecessary complexity to the current examples/tests.

dburgener:
My two cents on the bind mount loop problem is that I think the typical solution is to not traverse across file system boundaries on this sort of search.  Walkdir seems to provide a config knob for that: https://docs.rs/walkdir/2.4.0/walkdir/struct.WalkDir.html#method.same_file_system

> Hmm, not sure if the others like @dburgener have opinions on this. I'm not sure that we want to clutter lalrpop-test (since then all the tests fail if something goes wrong). The doc examples could be appropriate in that you might actually want to use symlinks when you have multiple lalrpop files being compiled. Alternatively, we could take a dev-dependency on tempfile and test just lalrpop_files in isolation.

I'm working on a small test harness to run lalrpop as an external library against various crates (mostly real-life users).  I've been delayed due to some other priorities in both work and personal life, but I did make some decent progress last week, and hope to finish it up in a few days.  While the primary point is for real-world applications, it seems reasonable to also include some test cases that aren't appropriate for lalrpop-test, such as symlink loops there.  Hard to make promises on timing unfortunately, but I really hope to have a first draft of the test harness repo up this week (without symlink loop support, but that won't be too heavy of an add).

mbid:
Ok, I've pushed a commit that rewrites the `eqlog_files` function using the `walkdir` crate. Could you have another look?

As you suggested @dburgener, I set `same_file_system` to `true` to prevent loops. I was a bit hesitant to do so, since this will break some legitimate (i.e. hierarchical) project structures that span several file system, but maybe those are esoteric enough that we can ignore them.

lalrpop/src/build/mod.rs
```diff
@@ -205,42 +207,29 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
...
 fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
     let mut result = vec![];
-    for entry in fs::read_dir(root_dir)? {
+
+    let walkdir = WalkDir::new(root_dir)
+        .follow_links(true)
+        // Use deterministic ordering:
+        .sort_by_file_name()
+        // Prevent loops:
+        .same_file_system(true);
```
Pat-Lafon, line 216: "Currently, this option is only supported on Unix and Windows. If this option is used on an unsupported platform, then directory traversal will immediately return an error and will not yield any entries." -> Is this something to be worried about?

lalrpop/src/build/mod.rs
```diff
@@ -205,42 +207,29 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
...
+        // `is_file` returns true.
+        if !entry.file_type().is_file() {
+            continue;
+        }
 
         let path = entry.path();
-
-        if path.is_dir() {
-            result.extend(lalrpop_files(&path)?);
+        if path.extension() != Some(OsStr::new("lalrpop")) {
```
Pat-Lafon, line 226: It's kind of a micro-optimization to ask if this check could have been implemented with https://docs.rs/walkdir/latest/walkdir/struct.IntoIter.html#method.filter_entry . More of a thought than any thing else, might not be worth changing.

lalrpop/src/build/mod.rs
```diff
@@ -205,42 +207,29 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
...
+            continue;
         }
 
-        let is_valid_symlink_file = || -> bool {
-            if !file_type.is_symlink() {
-                false
-            } else {
-                // Ensure all symlinks are resolved to a file
-                // Or ignore erroneous ones https://github.com/lalrpop/lalrpop/issues/808
```
Pat-Lafon, line 223: This implementation will probably regress on this issue?

If it's easy to skip over bad links then maybe we should keep that behavior. At the very least, I would like to make sure that if lalrpop does error out on a bad link, it contains the name.

Pat-Lafon (changes requested):
Overall I like this design much better than what it was previously. This should be much more maintainable/less error prone! I'm looking for clarification on at least two of my comments.

lalrpop/src/build/mod.rs
mbid: Yes, I saw this function as well. However, it looks like `filter_entry` is there mostly to skip traversing into subdirectories, not to skip individual files. We don't want to skip directories, so the lambda we pass to `filter_entry` would need to return `true` for directories. But then we'd consume the (now filtered) `WalkDir` iterator, we'd need to skip over directories. I think the current approach is simpler and not less efficient than an approach using `filter_entry`.

lalrpop/src/build/mod.rs
mbid: Perhaps it's best to just accept non-termination in case of loops resulting from bind mounts then, since `same_file_system` would also break some legitimate (i.e. directed) mount setups? I've removed the `same_file_system` call for now, let me know if I should put it in again.

lalrpop/src/build/mod.rs
mbid: You're right, this was a regression. I've added some handling code that should restore the current behavior: We print a warning about the dangling symlink and otherwise ignore it.

mbid:
Good points @pat-lafon! I tried to address them, could you have a look again?

lalrpop/src/build/mod.rs
Pat-Lafon: I was thinking about this because the files in a directory are sorted first, then we ignore everything that isn't a lalrpop file. It's probably not a bottleneck and as you say, probably not worth it over the simpler solution.

mbid: Ah, right, if the sorting happens after filtering, then a solution using `filter_entry` might actually be faster. I don't think we'll notice a difference in almost all cases though.

lalrpop/src/build/mod.rs
Pat-Lafon: It's something I'm fine with, if `walkdir` ever addresses this in a better way then we can add it back in. @dburgener can weigh in if he has other ideas.

lalrpop/src/build/mod.rs
```diff
@@ -205,6 +205,35 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
     }
 }
 
+/// Handles a [walkdir::Error] if the root cause is a dangling symlink.
+///
+/// Returns `Ok` if the error could be handled, otherwise returns `Err(err)`.
+fn handle_dangling_symlink_error(err: walkdir::Error) -> Result<(), walkdir::Error> {
+    let is_not_found = err.io_error().map(|io_err| io_err.kind()) == Some(io::ErrorKind::NotFound);
```
Pat-Lafon, line 212: Do we also want to gracefully handle symlink cycles here with [loop_ancestor](https://docs.rs/walkdir/latest/walkdir/struct.Error.html#method.loop_ancestor)?

lalrpop/src/build/mod.rs
Pat-Lafon: Your solution here looks reasonable to me. Just one last think about symlink cycles.

lalrpop/src/build/mod.rs
mbid: I believe no; I haven't tried but I think we'll print the error message according generated by `walkdir` and exit. Tbh I'm not sure I want to think too much about something we could do in case of cycles; wouldn't we need to recreate the cyclic symlinks also in the target directory so that every `.lalrop` file in the source directory is accessible under the same relative paths also in the target directory? I suggest we refuse compiling such crates for now, and if somebody has a real need for this we can think about it again.

lalrpop/src/build/mod.rs
dburgener: > Is this something to be worried about?

I did some digging to try to determine what "unix" meant in this case.  The reason it's exclusive to "unix" and "windows" is these calls to WalkDir's private "device_num" function https://docs.rs/crate/walkdir/latest/source/src/util.rs.  That is conditionally compiled on `target_family` `unix` or `windows`, and errors otherwise.

The `target_family` config is extremely poorly documented.  Most of the links point to https://doc.rust-lang.org/reference/conditional-compilation.html, which gives "unix", "windows", and "wasm" as examples, but doesn't tell you how OSs map to families, or whether that list of examples is exclusive.

Eventually I found this stackoverflow post: https://stackoverflow.com/questions/66845716/where-are-rusts-built-in-target-specifications-defined which gives a way to query the architecture to `target_family` mapping on your system.  Running that command locally confirms that `unix`, `windows` and `wasm` are the only three options for built in target specs on my system.

So my understanding is that that "unix" in this context means all unix-like operating systems (eg Linux, BSD (including Mac), android, etc).  So "unix" and "windows" is pretty close to covering everything.  The only other mainstream option is "wasm".  Since this is code to detect lalrpop files at build time, that would break someone who's building directly in a wasm environment (which might actually be the situation for the rust playground?).  Of course, users aren't limited to the built-in options.  In theory a user could set target_family to an arbitary string, which would probably happen in specific embedded case.  Again though, this is called from build.rs, so if you're cross-compiling to a custom target, you wouldn't hit this code.

In other words, I suspect that use cases without "unix" or "windows" here, while technically possible are probably rare (the rust playground might be the big exception though).

One option could be to do a `#[cfg(any(unix, windows))]` ourselves, just to be certain, with the behavior of setting the same_file_system option if we're on one of those families, and otherwise not.

All that said, maybe if you've created a looped directory structure via bind mounts inside your source tree, you only have yourself to blame when things go wrong, so it may not be worth it.

lalrpop/src/build/mod.rs
mbid: Or do you mean by "gracefully" handling it that we should print a reasonable error, and not that we should try to proceed with lalrpop code generation?

lalrpop/src/build/mod.rs
Pat-Lafon: No strong opinions, but I think your final point is accurate. Lalrpop should do the reasonable thing which works for the most amount of people and if your file system is cursed enough then it's up to you to figure out a work around.

lalrpop/src/build/mod.rs
Pat-Lafon: Hmm, fair enough. It doesn't look like normal `io::Error`'s can be converted into `walkdir::Error` so it's a bit more work to propagate this information up the stack.

We could raise an warning and try to keep compiling, but your right that that could get wonky.

Pat-Lafon (approved):
I'm content with the current state of this pr. Setting up the proper regression tests is non-trivial and probably large enough to be its own thing.

dburgener (approved):
LGTM too

mbid:
LGTM as well. Thanks @Pat-Lafon and @dburgener for the great review!

PR #855, opened 2023-11-13
Add symlinks to license files

### LINKED ISSUE #600: please include LICENSE-* files in all published crates
The Apache-2.0 and MIT License terms require that the license text is made available with any redistributed sources, which also means that the LICENSE-* files should be included in crates published to crates.io.

To have "cargo publish" pick the files up, adding symlinks `LICENSE-MIT → ../LICENSE-MIT` and `LICENSE-APACHE → ../LICENSE-APACHE` in each of the workspace member directories (at least those that are actually published) is enough, if you're running "cargo publish" from Linux or mac OS (Windows does not really support symlinks, so there actual copies of the files are necessary).

Pat-Lafon:
Closes #600 

I don't fully understand the license implications here but from my reading, the correct solution from cargo has not fully manifested yet. In looking what other projects like tokio/regex do, they include the license files in each project so it seems reasonable to follow that strategy.

See https://github.com/bytecodealliance/wasmtime/pull/4664 https://github.com/tokio-rs/tracing/pull/842 https://github.com/rust-lang/cargo/issues/8537 for some discussion on these issues.

### DIFF
--- a/lalrpop-util/LICENSE-APACHE
+++ b/lalrpop-util/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
--- a/lalrpop-util/LICENSE-MIT
+++ b/lalrpop-util/LICENSE-MIT
@@ -1,0 +1,1 @@
+../LICENSE-MIT
--- a/lalrpop/LICENSE-APACHE
+++ b/lalrpop/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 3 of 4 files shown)


dburgener (approved):

dburgener:
I read the associated links, and it seems like this is the right thing to do.  Ideally, cargo would have support for copying this automatically, but since they've opted against doing that, it seems like this is what we need to do in order to meet our obligations under the apache and MIT licenses.

The biggest problem with duplication in software repos IMO is the risk of missing a copy when changing things.  In the case of symlinks, this is only a problem if we move or rename our license files, which I imagine is something that would happen very infrequently.

Issue #862, opened 2024-01-11
fail to enable lalr parser


AphrosDelos532:
I have read the tutorial. There are something wrong when I try to use lalr.

I have added `set_var` in `build.rs`，and marked grammar with `#[lalr]` attribute.

```
// build.rs
use std::path::Path;

fn main() {
    std::env::set_var("LALRPOP_LANE_TABLE", "disabled");
    lalrpop::process_root().unwrap();
    let mut config = lalrpop::Configuration::new();
    config.always_use_colors();
    config
        .process_file(Path::new("src/parser/parser.lalrpop"))
        .unwrap();
}


// key part of lalrpop file
...
use crate::ast::pool::ContextPool;
#[lalr]
grammar(pool: &ContextPool);
...
```

Finally, I got the follow
```
PATH/TO/PROJECT/src/parser/parser.lalrpop:8:3: 8:6 error: unrecognized annotation `lalr`

  --- stderr
    #[lalr]
      ~~~~
```

How can I fix it?

AphrosDelos532:
Some detail in `Cargo.toml`
```toml
[build-dependencies]
lalrpop = "0.20.0"

[dependencies]
lalrpop-util = { version = "0.20.0", features = ["lexer"] }
```

patrickzbhe:
Tutorial is wrong, needs to be #[LALR] not #[lalr]

PR #863, opened 2024-01-15
Consistent LALR docs and clippy

Pat-Lafon:
Addresses the documentation confusion with #862 and some of the new clippy lints.


dburgener (approved):
This all seems reasonable to me.

Issue #387, opened 2018-06-30
Listing all "features" 


rustonaut:
I think it would be a nice think to have a list of all features with a one sentence description and
a link to a part of the tutorial explaining them.  Or more specific start with the list and
add links once the tutorial covers them.

The reason for this is two fold:

1. features not covered by the tutorial are hard to find or to  know what they are if you 
   stumble across them in the wild
2. the tutorial is a general tutorial, but sometimes this makes it not easy to find the description of an
    specific feature in them

A start could be along the line of:

| name | snipped | description | tutorial |
|---|---|---|---|
| position | `<@L> T <@R>` | uhm, I think it gives the index of the parser in the input | - | 
| error_recovery | `! => { ... }` | recover from parser errors | link | 
| grammar_parameter | `grammar(scale: isize);` | input parameters usable in the generated parser | - | 
| ?? | `Num => func(<i32>)` | maybe automatic number parsing?? | - | 
| custom_error | `"e" =>? ParseError::User { error: "an error" }` | make an action failable | - | 
| custom_macros | `Comma<T> =  { ... }` | allows a non-terminal to be generic about some parameters | link  |
| qunatifier_macros | `<Num?> <Num*> <Num+>` |  a non-terminal which can appear 0..1, 0+, 1+ times | link |
| tuple_macro | `<a:(<Num> ",")*>` | when used with a qunatifier allows crating lists | link |
| extern | `extern { }` | allows to override some parts of the generated parser | - |

Also an extra table for `external`:

| name | snipped | description | tutorial |
|---|---|---|---|
| extern_error | `type Error = MyError;` | sets the error to use in the `ParseError::User` variant | - |
| extern_location | `type Location = MyLoc;` | set the type to for locations instead of `usize` | - |
| extern_tok | `enum Tok { }` | set the type to use for tokens?? | - |

- The idea behind `name` is to have a name to speak about them in issues, name
tests etc.
- The idea of snipped is that if someone sees a unknown feature in the wild he/she
   can likely recognize it based on the snipped
- The description should just give some general idea, it doesn't have to be specific 
  of exhaustive
- Even if some thinks are documented together (e.g. all macro thinks) it's worth listing
   them separately as this can make it easier to recognize what you are looking for

It could be consider to add an additional column if there is a reference documentation
which is with less fluff and example but more exhaustive (e.g. it covers more about the
interaction of `custom_error` and `error_recovery`)

[...]

PR #867, opened 2024-02-08
Kick the tires on a documentation cheatsheet

Pat-Lafon:
I think the idea of a feature list "cheatsheet" is a great idea as suggested in #387. I've done a pretty minimal adaptation of the table elements suggested in that list with their corresponding links.

I don't have a clear idea of the right way to present this, so open to all suggestions.

### DIFF
--- a/doc/src/cheatsheet.md
+++ b/doc/src/cheatsheet.md
@@ -10,7 +10,7 @@ functionality, use this table to jump to the right section.
 | error_recovery | `! => { ... }` | recovers from parser errors | [Error recovery](tutorial/008_error_recovery.md) |
 | grammar_parameter | `grammar(scale: isize);` | input parameters usable in the generated parser | [Passing state parameter](tutorial/009_state_parameter.md) |
 | ?? | `Num => func(<i32>)` | maybe automatic number parsing?? | - |
-| custom_error | `"e" =>? ParseError::User { error: "an error" }` | make an action failable | [Fallible actions](tutorial/007_fallible_actions.md) |
+| custom_error | `"e" =>? Err(ParseError::User { error: "an error" })` | makes an action fallible | [Fallible actions](tutorial/007_fallible_actions.md) |
 | custom_macros | `Comma<T> =  { ... }` | allows a non-terminal to be generic about some parameters | [Macros](tutorial/006_macros.md) |
 | quantifier_macros | `<Num?> <Num*> <Num+>` |  a non-terminal which can appear 0..1, 0+, 1+ times | [Macros](tutorial/006_macros.md) |
 | tuple_macro | `<a:(<Num> ",")*>` | when used with a qunatifier allows crating lists | [Macros](tutorial/006_macros.md) |
--- a/doc/src/cheatsheet.md
+++ b/doc/src/cheatsheet.md
@@ -18,6 +18,6 @@ functionality, use this table to jump to the right section.
 | extern_error | `type Error = MyError;` | sets the error to use in the `ParseError::User` variant | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
 | extern_location | `type Location = MyLoc;` | sets the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
 | extern_tok | `enum MyToken { }` | declares the type of lexer tokens to be consumed by the generated parser  | [Using tokens with references](lexer_tutorial/004_token_references.md) |
-| auto_parameters | `<>` | in an action it turns into a list of parameters | [Type inference](tutorial/003_type_inference.md) |
+| auto_parameters | `<>` | refers to all the parameters of the non-terminal as a tuple | [Type inference](tutorial/003_type_inference.md) |
 |conditional actions | `<T> if T == "a" => (),` | - | - |
 |precedence| `#[precedence(level="0")]` | - | [Handling full expressions](tutorial/004_full_expressions.md) |
--- a/doc/src/cheatsheet.md
+++ b/doc/src/cheatsheet.md
@@ -17,7 +17,7 @@ functionality, use this table to jump to the right section.
 | extern | `extern { }` | allows to override some parts of the generated parser | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
 | extern_error | `type Error = MyError;` | sets the error to use in the `ParseError::User` variant | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
 | extern_location | `type Location = MyLoc;` | sets the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
-| extern_tok | `enum Tok { }` | set the type to use for tokens?? | [Using tokens with references](lexer_tutorial/004_token_references.md) |
+| extern_tok | `enum MyToken { }` | declares the type of lexer tokens to be consumed by the generated parser  | [Using tokens with references](lexer_tutorial/004_token_references.md) |
 | auto_parameters | `<>` | in an action it turns into a list of parameters | [Type inference](tutorial/003_type_inference.md) |
 |conditional actions | `<T> if T == "a" => (),` | - | - |
 |precedence| `#[precedence(level="0")]` | - | [Handling full expressions](tutorial/004_full_expressions.md) |
--- a/doc/src/cheatsheet.md
+++ b/doc/src/cheatsheet.md
@@ -13,7 +13,7 @@ functionality, use this table to jump to the right section.
 | custom_error | `"e" =>? Err(ParseError::User { error: "an error" })` | makes an action fallible | [Fallible actions](tutorial/007_fallible_actions.md) |
 | custom_macros | `Comma<T> =  { ... }` | allows a non-terminal to be generic about some parameters | [Macros](tutorial/006_macros.md) |
 | quantifier_macros | `<Num?> <Num*> <Num+>` |  a non-terminal which can appear 0..1, 0+, 1+ times | [Macros](tutorial/006_macros.md) |
-| tuple_macro | `<a:(<Num> ",")*>` | when used with a qunatifier allows crating lists | [Macros](tutorial/006_macros.md) |
+| tuple_macro | `<a:(<Num> ",")*>` | applies a quantifier to a group of matches | [Macros](tutorial/006_macros.md) |
 | extern | `extern { }` | allows to override some parts of the generated parser | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
 | extern_error | `type Error = MyError;` | sets the error to use in the `ParseError::User` variant | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
 | extern_location | `type Location = MyLoc;` | sets the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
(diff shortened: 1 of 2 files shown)


doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
+# Cheatsheet
+
+Users of Lalrpop has compiled the following cheatsheet table as a quick way to
+look up useful Lalrpop-isms. If you are looking for a specific piece of
+functionality, use this table to jump to the right section.
+
+| name | snipped | description | tutorial |
+|---|---|---|---|
+| position | `<@L> T <@R>` | uhm, I think it gives the index of the parser in the input | - |
```
yannham, line 9: Not that the original example doesn't work, but let's

```suggestion
| position | `<left: @L> T <right: @R>` | captures the byte offset of the first byte and the offset of the last byte plus one (as `left` and `right` respectively) | - |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+| error_recovery | `! => { ... }` | recover from parser errors | [Error recovery](tutorial/008_error_recovery.md) |
+| grammar_parameter | `grammar(scale: isize);` | input parameters usable in the generated parser | [Passing state parameter](tutorial/009_state_parameter.md) |
+| ?? | `Num => func(<i32>)` | maybe automatic number parsing?? | - |
+| custom_error | `"e" =>? ParseError::User { error: "an error" }` | make an action failable | [Fallible actions](tutorial/007_fallible_actions.md) |
```
yannham, line 13:
```suggestion
| custom_error | `"e" =>? Err(ParseError::User { error: "an error" })` | makes an action fallible | [Fallible actions](tutorial/007_fallible_actions.md) |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+| custom_macros | `Comma<T> =  { ... }` | allows a non-terminal to be generic about some parameters | [Macros](tutorial/006_macros.md) |
+| quantifier_macros | `<Num?> <Num*> <Num+>` |  a non-terminal which can appear 0..1, 0+, 1+ times | [Macros](tutorial/006_macros.md) |
```
yannham, line 15: I don't think those are really called macros somewhere in LALRPOP's documentation, are they?

```suggestion
| quantifier | `<Num?> <Num*> <Num+>` |  a non-terminal which can appear 0..1, 0+, 1+ times | [Macros](tutorial/006_macros.md) |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+| tuple_macro | `<a:(<Num> ",")*>` | when used with a qunatifier allows crating lists | [Macros](tutorial/006_macros.md) |
```
yannham, line 16:
```suggestion
| tuple_macro | `<a:(<Num> ",")*>` | applies a quantifier to a group of matches | [Macros](tutorial/006_macros.md) |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+| extern | `extern { }` | allows to override some parts of the generated parser | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_error | `type Error = MyError;` | sets the error to use in the `ParseError::User` variant | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_location | `type Location = MyLoc;` | set the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_tok | `enum Tok { }` | set the type to use for tokens?? | [Using tokens with references](lexer_tutorial/004_token_references.md) |
```
yannham, line 20: Nitpick, but just to make it clear that `Tok` isn't a magic symbol and we can actually use whatever custom token type we've defined elsewhere in our custom lexer.

```suggestion
| extern_tok | `enum MyToken { }` | declare the type of lexer tokens to be consumed by the generated parser  | [Using tokens with references](lexer_tutorial/004_token_references.md) |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+Users of Lalrpop has compiled the following cheatsheet table as a quick way to
+look up useful Lalrpop-isms. If you are looking for a specific piece of
+functionality, use this table to jump to the right section.
+
+| name | snipped | description | tutorial |
+|---|---|---|---|
+| position | `<@L> T <@R>` | uhm, I think it gives the index of the parser in the input | - |
+| error_recovery | `! => { ... }` | recover from parser errors | [Error recovery](tutorial/008_error_recovery.md) |
+| grammar_parameter | `grammar(scale: isize);` | input parameters usable in the generated parser | [Passing state parameter](tutorial/009_state_parameter.md) |
+| ?? | `Num => func(<i32>)` | maybe automatic number parsing?? | - |
```
yannham, line 12: What does this one do, exactly? Do you have a complete example? Because I'm not sure to know the feature you're referring to

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+| custom_error | `"e" =>? ParseError::User { error: "an error" }` | make an action failable | [Fallible actions](tutorial/007_fallible_actions.md) |
+| custom_macros | `Comma<T> =  { ... }` | allows a non-terminal to be generic about some parameters | [Macros](tutorial/006_macros.md) |
```
yannham, line 14:
```suggestion
| custom_macros | `Comma<T> =  { ... }` | makes a non-terminal generic in other non-terminals | [Macros](tutorial/006_macros.md) |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+| quantifier_macros | `<Num?> <Num*> <Num+>` |  a non-terminal which can appear 0..1, 0+, 1+ times | [Macros](tutorial/006_macros.md) |
+| tuple_macro | `<a:(<Num> ",")*>` | when used with a qunatifier allows crating lists | [Macros](tutorial/006_macros.md) |
+| extern | `extern { }` | allows to override some parts of the generated parser | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_error | `type Error = MyError;` | sets the error to use in the `ParseError::User` variant | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_location | `type Location = MyLoc;` | set the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_tok | `enum Tok { }` | set the type to use for tokens?? | [Using tokens with references](lexer_tutorial/004_token_references.md) |
+| auto_parameters | `<>` | in an action it turns into a list of parameters | [Type inference](tutorial/003_type_inference.md) |
```
yannham, line 21:
```suggestion
| auto_parameters | `<>` | refers to all the parameters of the non-terminal as a tuple | [Type inference](tutorial/003_type_inference.md) |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+|conditional actions | `<T> if T == "a" => (),` | - | - |
```
yannham, line 22: Although conditional macro are currently undocumented, I believe the idea is to have a macro parameter (ab)used as a string constant, but usually you don't use this parameter directly in the alternative (see https://gist.github.com/nikomatsakis/5fa3bd8291841b853144)

```suggestion
|conditional actions | `Expr<I> = ... , <T> if I == "a" => (),` | Conditional definition of a macro's alternative | - |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
...
+| ?? | `Num => func(<i32>)` | maybe automatic number parsing?? | - |
+| custom_error | `"e" =>? ParseError::User { error: "an error" }` | make an action failable | [Fallible actions](tutorial/007_fallible_actions.md) |
+| custom_macros | `Comma<T> =  { ... }` | allows a non-terminal to be generic about some parameters | [Macros](tutorial/006_macros.md) |
+| quantifier_macros | `<Num?> <Num*> <Num+>` |  a non-terminal which can appear 0..1, 0+, 1+ times | [Macros](tutorial/006_macros.md) |
+| tuple_macro | `<a:(<Num> ",")*>` | when used with a qunatifier allows crating lists | [Macros](tutorial/006_macros.md) |
+| extern | `extern { }` | allows to override some parts of the generated parser | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_error | `type Error = MyError;` | sets the error to use in the `ParseError::User` variant | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
+| extern_location | `type Location = MyLoc;` | set the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
```
yannham, line 19:
```suggestion
| extern_location | `type Location = MyLoc;` | sets the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
```

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,23 @@
+# Cheatsheet
+
+Users of Lalrpop has compiled the following cheatsheet table as a quick way to
+look up useful Lalrpop-isms. If you are looking for a specific piece of
+functionality, use this table to jump to the right section.
+
+| name | snipped | description | tutorial |
+|---|---|---|---|
+| position | `<@L> T <@R>` | uhm, I think it gives the index of the parser in the input | - |
+| error_recovery | `! => { ... }` | recover from parser errors | [Error recovery](tutorial/008_error_recovery.md) |
```
yannham, line 10:
```suggestion
| error_recovery | `! => { ... }` | recovers from parser errors | [Error recovery](tutorial/008_error_recovery.md) |
```

yannham (commented):
This is a great idea! I made a first quick pass.

doc/src/cheatsheet.md
Pat-Lafon: The documentation linked says: 
```
In fact, LALRPOP comes with four macros builtin: `*`, `?`, `+`, and `(...)`.
```

I am happy with just calling them quantifiers though to be short and clear.

yannham: Oh, I missed that. Then it makes sense to stay consistent here :+1:

doc/src/cheatsheet.md
Pat-Lafon: Yeah, I'm not sure what the original poster meant by this the more I look into it. The closest I was able to find was something like:
```
extern {
    enum Tok<'input> {
        "(" => Tok::LParen,
        ")" => Tok::RParen,
        "-" => Tok::Minus,
        "+" => Tok::Plus,
        "*" => Tok::Times,
        "/" => Tok::Div,
        Num => Tok::Num(<i32>),
        Fraction => Tok::Fraction(<i32>, <i32>), // Regression test for #179
    }
}
```

Pat-Lafon:
@yannham Thank you for looking at this. I've largely adopted your feedback. Two features don't have documentation pages but are mentioned in the index so I've added links to there for consistency. I've also added a description to the precedence example.

We should probably cut the "automatic number parsing" feature, it's not what I expected it to be.

doc/src/cheatsheet.md
yannham: Oh, I see. I believe it's the usual way to map terminals to lexer tokens when using a custom lexer. You have to specify the type of data that the token carries (and that you want to extract in the parser): this just means that `Fraction` is defined as `enum Tok { ..., Fraction(i32, i32), ... }` in the lexer, which will correspond to the non-terminal `Fraction` that will have type `(i32, i32)` in the parser.

yannham (approved):
Right, I think what you're referring to is normal custom lexer business, and probably doesn't belong here, at least for a first iteration. Beside that, it's a good first version :+1:

doc/src/cheatsheet.md
```diff
@@ -0,0 +1,22 @@
+# Cheatsheet
+
+Users of Lalrpop has compiled the following cheatsheet table as a quick way to
```
yannham, line 3:
```suggestion
Users of Lalrpop have compiled the following cheatsheet table as a quick way to
```

yannham:
@Pat-Lafon if you're happy with this state, I think you can go ahead and merge :slightly_smiling_face: we should probably add a visible link to the cheatsheet in the README as well once it's published

Issue #869, opened 2024-02-16
Update (external library) lexer example for Logos 0.13+


airstrike:
[Since 0.13](https://github.com/maciejhirsz/logos/releases/tag/v0.13), Logos no longer requires the `#[error]` variant, yet the example in the tutorial still uses it:

https://github.com/lalrpop/lalrpop/blob/0b244afd19f6da8f5c095e8c780a48f80abe21bf/doc/src/lexer_tutorial/005_external_lib.md?plain=1#L94-L100

Pat-Lafon:
This got addressed in #874

Issue #766, opened 2023-04-22
Document `lexer` feature


youknowone:
> Ideally we would go over the docs to clarify the role of this feature and add intructions to the "custom lexer" section for the way to remove it.

https://github.com/lalrpop/lalrpop/pull/711#discussion_r1174264272

PR #870, opened 2024-02-19
First step to add a lexer example in tree

Pat-Lafon:
I'm interested in addressing #869. Before upgrading the version of logos, I would first like to add a testable version of the logos lexer example. This involves extracting out the example into it's own `doc/lexer` repo(Which is an unfortunate duplication given how documentation is generated). However in doing so I've run across some trivial issues in the documentation that I've addressed.

One part I'm not sure how to address is `the trait bound `Token: __ToTriple` is not satisfied` because `trait `__ToTriple` is not implemented for `Token``. Maybe I'll get around to figuring it out at some point. Until then, I'm leaving this pr here as a first step.(If someone knows the answer or wants to give it a try, feel free to suggest a edit).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
         "doc/nobol",
         "doc/pascal/lalrpop",
         "doc/whitespace",
+        "doc/lexer",
         "lalrpop-test",
         "lalrpop-util",
         "lalrpop",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ rust-version = "1.70"
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
 regex = { version = "1.3", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.8", default_features = false }
+regex-syntax = { version = "0.8.2", default_features = false }
 regex-automata = { version = "0.4", default_features = false, features = [
         "perf",
         "syntax",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "beef"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3a8241f3ebb85c056b509d4327ad0358fbbba6ffb340bf388f26350aeda225b1"
+
+[[package]]
 name = "bit-set"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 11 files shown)


Pat-Lafon:
This is maybe a step towards #766

dburgener:
Don't you need to pass a Lexer to parse(), rather than a Token?

This seems to at least compile for me:

```
diff --git a/doc/lexer/src/main.rs b/doc/lexer/src/main.rs
index 6a78b14..9484bae 100644
--- a/doc/lexer/src/main.rs
+++ b/doc/lexer/src/main.rs
@@ -1,5 +1,5 @@
 use lexer::grammar::ScriptParser;
-use lexer::token::Token;
+use lexer::lexer::Lexer;
 use logos::Logos;
 
 fn main() {
@@ -9,7 +9,7 @@ var b = 23;
 # a comment
 print (a - b);";
 
-    let lexer = Token::lexer(&source_code[..]);
+    let lexer = Lexer::new(&source_code[..]);
     let parser = ScriptParser::new();
     let ast = parser.parse(lexer).unwrap();
```

Pat-Lafon:
> Don't you need to pass a Lexer to parse(), rather than a Token?
 
Ah, whoops. Thanks. I've made your change and then cleaned things up. It now runs as expected and caught a few small documentation fixes.

Something of note is that pulling in logos 1.14 raised the minimum version of regex_syntax to 0.8.2

Cargo.toml
```diff
@@ -25,7 +26,7 @@ rust-version = "1.70"
 [workspace.dependencies]
 diff = { version = "0.1.12", default_features = false }
 regex = { version = "1.3", default_features = false, features = ["std"] }
-regex-syntax = { version = "0.8", default_features = false }
+regex-syntax = { version = "0.8.2", default_features = false }
```
dburgener, line 29: Do we actually need to specify this here?  If the `.2` support is just a dependency of logos, I think specifying only `0.8` on our end will result in feature unification demanding the right version when it build logos.  The advantage to only specifying `0.8` being that if something about logos were to change, we get that automatically, and could in theory support (for example) 0.8.1 environments if logos were to drop their requirement.

On my box, only specifying 0.8 works fine and pulls in the 0.8.2 version as expected.

dburgener (approved):
I think this looks good besides the one comment about specifying the regex_syntax version.

Cargo.toml
Pat-Lafon: It got caught up in the min_versions workflow check. Unfortunately, we can't have dev-dependency members or something like that in the workspace(since we really only want the 0.8.2 version for testing). I think we could: disable that check, remove the lexer example form the workspace members and run it manually, ask the logos people to use the minimum version on their end or just stick with 0.8.2 since it's a patch release which should be trivial upgradable for someone in a 0.8.1 environment(and if it isn't, that's a regex bug).

dburgener: Yuck.  I'm inclined to call this a bug in -Zdirect-minimal-versions, personally.  (I found someone trying to report it here: https://github.com/rust-lang/cargo/issues/5657#issuecomment-1914446399, and it didn't seem like epage agreed that it was a bug, although I don't think the reporters desired behavior was what I was expecting either).

Regardless, I imagine it's messy/undoable to get minimal versions and dependency resolution playing nicely (I'd expect it to ultimately resolve to 0.8.2 here, since that what's actually minimal required *after* feature resolution), so this is probably not a battle worth fighting. I'd rather just set our version to 0.8.2 than any of the workaround you mention.

PR #875, opened 2024-03-16
Add 'OUT_DIR' to quick start and tutorial

galexite:
Tried to make it clearer that the generated files are by default saved to `OUT_DIR`, and gave an example of `include!()`-ing the generated files.

### DIFF
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -31,10 +31,22 @@ fn main() {
 (If you already have a `build.rs` file, you should be able to just
 call `process_root` in addition to whatever else that file is doing.)
 
-That's it! Note that `process_root` simply uses the default settings, which
-means generated Rust code will reside in `OUT_DIR`, which you can `include!()`.
-If you want to configure how LALRPOP executes, see the
-[advanced setup](advanced_setup.md) section.
+In this case, `process_root` simply uses the default settings, which takes
+files in `src/` ending with the `.lalrpop` extension, and generates
+corresponding Rust source files with the same name in `OUT_DIR`. If you want to
+configure how LALRPOP executes, see the [advanced setup](advanced_setup.md)
+section.
+
+Finally, you can add any of the generated files as modules in your Rust code
+using the [`lalrpop_mod!`][lalrpop_mod] macro. For example, if your grammar is
+called `grammar.lalrpop`, to use the generated `grammar.rs` as a Rust module
+simply called `grammar`, add the following:
+
+```rust
+lalrpop_mod!(grammar)
+```
+
+[lalrpop_mod]: https://docs.rs/lalrpop-util/latest/lalrpop_util/macro.lalrpop_mod.html
 
 #### Running manually
 
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -3,7 +3,7 @@ For getting started with LALRPOP, it's probably best if you read
 to the syntax of LALRPOP files and so forth.
 
 But if you've done this before, or you're just the impatient sort,
-here is a quick 'cheat sheet' for setting up your project.  First, add
+here is a quick 'cheat sheet' for setting up your project. First, add
 the following lines to your `Cargo.toml`:
 
 ```toml
@@ -31,7 +31,8 @@ fn main() {
 (If you already have a `build.rs` file, you should be able to just
 call `process_root` in addition to whatever else that file is doing.)
 
-That's it! Note that `process_root` simply uses the default settings.
+That's it! Note that `process_root` simply uses the default settings, which
+means generated Rust code will reside in `OUT_DIR`, which you can `include!()`.
 If you want to configure how LALRPOP executes, see the
 [advanced setup](advanced_setup.md) section.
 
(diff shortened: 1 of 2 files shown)


doc/src/tutorial/001_adding_lalrpop.md
```diff
@@ -61,11 +61,19 @@ fn main() {
...
+all `lalrpop` files into `rs` files, and saving them to `OUT_DIR`. It is smart
+enough to check timestamps and do nothing if the `rs` file is newer than the
+`lalrpop` file, and to mark the generated `rs` file as read-only. It returns an
 `io::Result<()>`, so the `unwrap()` call just asserts that no
 file-system errors occurred.
 
-*NOTE:* On Windows, the necessary APIs are not yet stable, so
+You can include the generated code somewhere in your Rust code. For example, if
+the source grammar were to be called `grammar.lalrpop`, you could include the
+generated `grammar.rs` file like so:
```
yannham, line 72:
```suggestion
the source grammar is in `grammar.lalrpop`, you can include the
generated `grammar.rs` file like so:
```

doc/src/tutorial/001_adding_lalrpop.md
```diff
@@ -61,11 +61,19 @@ fn main() {
...
+
+```rust
+include!(concat!(env!("OUT_DIR"), "/grammar.rs"));
```
yannham, line 75: What is the purpose of including the generated gammar like this? The wording sounds like this is the normal way of using the generated grammar, while one should probably prefer `lalrpop_mod!` instead.

Pat-Lafon: Could also link to the https://lalrpop.github.io/lalrpop/generate_in_source.html section as a way to get similar behavior I think.

galexite: This is the very first time I've hearing of `lalrpop_mod!`. The documentation search shows it appears in a few code samples without explicitly mentioning what it does: https://lalrpop.github.io/lalrpop/index.html?search=Lalrpop_mod

(Only one example references it in text, saying 'it's not useful' when generating to the source directory.)

galexite: (So yeah, this should probably appear more often in the docs. I'll update my PR when I have the chance.)

yannham: Oh, I thought `lalrpop_mod` was part of the documentation in a clear way...thanks @galexite, mentioning `lalrpop_mod` is a most wanted addition then. I just assumed `lalrpop_mod` was known and that you wanted to achieve something different, but I couldn't really figure out what.

doc/src/quick_start_guide.md
```diff
@@ -31,9 +31,22 @@ fn main() {
...
+section.
+
+Finally, you can add any of the generated files as modules in your Rust code
+using the [`lalrpop_mod!`][lalrpop_mod] macro. For example, if your grammar is
+called `grammar.lalrpop`, to use the generated `grammar.rs` as a Rust module
+simply called `grammar`, add the following:
+
+```rust
+lalrpop_mod!(grammar)
+```
```
yannham, line 47: I found the part "include the generated code" to be slightly misleading, so here is a proposal for a more high-level (focused on what users _want to achieve_, and not how it works under the hood) and hopefully more direct-style wording. I also copy-paster the suggestion below because I think the two texts should be the same. Feel free to amend/reject if you don't think it's a good suggestion

```suggestion
The [`lalrpop_mod!`][lalrpop_mod] macro generates a wrapper module in your
crate so that you can use the generated parser from your code. For example,
if the source grammar is located in `grammar.lalrpop`, adding the following line
to `lib.rs` will create a corresponding `grammar` submodule (note that you can
also add this line to a `foo.rs` module definition instead, which will then
create a submodule `foo::grammar`):

```rust
lalrpop_mod!(grammar);
```
```

File: doc/src/tutorial/001_adding_lalrpop.md [hunk 4]
```diff
@@ -61,11 +61,21 @@ fn main() {
...
+all `lalrpop` files into `rs` files, and saving them to `OUT_DIR`. It is smart
+enough to check timestamps and do nothing if the `rs` file is newer than the
+`lalrpop` file, and to mark the generated `rs` file as read-only. It returns an
 `io::Result<()>`, so the `unwrap()` call just asserts that no
 file-system errors occurred.
 
-*NOTE:* On Windows, the necessary APIs are not yet stable, so
+You can include the generated code somewhere in your Rust code. For example, if
+the source grammar is in `grammar.lalrpop`, you can add the generated
+`grammar.rs` file as a module using the [`lalrpop_mod!`][lalrpop_mod] macro:
```
yannham, line 72:
```suggestion
The [`lalrpop_mod!`][lalrpop_mod] macro generates a wrapper module in your
crate so that you can use the generated parser from your code. For example,
if the source grammar is located in `grammar.lalrpop`, adding the following line
to `lib.rs` will create a corresponding `grammar` submodule (note that you can
also add this line to a `foo.rs` module definition instead, which will then
create a submodule `foo::grammar`):
```

galexite:
Looks fine to me! Suggestions committed.

yannham (approved):

yannham:
The lint failure seems entirely unrelated, and is probably due to bumping the Rust version and having a stricter clippy. As this just touches documentation, I'm merging this as it is. Thanks again @galexite !

PR #877, opened 2024-03-23
Remove outdated note?

Pat-Lafon:
I saw this note in the docs and it surprised me. I tracked this back to a commit over 9 years ago: 0983cec. In looking through the issue tracker, this might have been fixed as early as 2 weeks later in #30. Regardless, Lalrpop now uses cargo's `rerun-if-changed` directive so this is outside of Lalrpop's scope.


dburgener:
It looks to me like #30 had some issues, but this issue was later corrected in #31, but as you point out, rerun-if-changed fixes the problem as well, so yes, this comment looks obsolete.

dburgener (approved):

PR #894, opened 2024-05-06
Switch out the tiny_keccak hashing library for the sha3 library

dburgener:
The sha3 library seems more widely used and actively maintained.

All we do with these hashes is save the hash in the output file and compare to determine if we need a rebuild.


Pat-Lafon (approved):
Makes a lot of sense to me!

Issue #892, opened 2024-05-05
warning: unexpected `cfg` condition name: `rustfmt`


asomers:
The latest rustc compiler tries to detect the use of unknown `#[cfg()]` conditions.  It's triggering in lalrpop-generated code.  The easiest way to reproduce the problem is to simply run `cargo check -p nobol` in lalrpop's own git repo.  Or, it can be reproduced by trying to compile a file like this:
```
grammar;

pub Num: i16 = r"[0-9]+" => i16::from_str(<>).unwrap();
```
Which results in the warning
```
warning: unexpected `cfg` condition name: `rustfmt`
  --> /usr/home/somers/src/rust/bfffs/target/debug/build/bfffs-555042f4e29eb538/out/x.rs:12:12
   |
12 | #[cfg_attr(rustfmt, rustfmt_skip)]
   |            ^^^^^^^
   |
   = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(rustfmt)");` to the top of the `build.rs`
   = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration

warning: unexpected `cfg` condition name: `rustfmt`
   --> /usr/home/somers/src/rust/bfffs/target/debug/build/bfffs-555042f4e29eb538/out/x.rs:303:12
    |
303 | #[cfg_attr(rustfmt, rustfmt_skip)]
    |            ^^^^^^^
    |
    = help: consider using a Cargo feature instead or adding `println!("cargo::rustc-check-cfg=cfg(rustfmt)");` to the top of the `build.rs`
    = note: see <https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg> for more information about checking conditional configuration
```
This is the rust version I'm currently using:
```
> rustc --version -v
rustc 1.80.0-nightly (e82c861d7 2024-05-04)
binary: rustc
commit-hash: e82c861d7e5ecd766cb0dab0bf622445dec999dc
commit-date: 2024-05-04
host: x86_64-unknown-freebsd
release: 1.80.0-nightly
LLVM version: 18.1.4
```

Pat-Lafon:
This just landed on nightly right? I'm interested in fixing this(I think it exposed some other cfg issues that went undetected with `cfg(feature = "test")`. Ironically though, the example you included has just been fixed on the latest nightly(because of issues with rustfmt).

[...]

PR #895, opened 2024-05-07
Whoops, I don't think a test feature flag exists

Pat-Lafon:
I was poking at #892 and I noticed this warning pop up. I had to double check(and I see this is a common confusion it seems https://stackoverflow.com/questions/72908829/what-is-the-difference-between-cfgtest-and-cfgfeature-test).

Given the code in question, I think it just always pulled from `mod lrgrammar;` without issue?


Pat-Lafon:
It seems like this was a feature flag until it was removed in #786. This currently then fails to work because the build script expects to find the `"CARGO_FEATURE_TEST"` environment variable which is no longer set.
https://github.com/lalrpop/lalrpop/blob/f1060df4c9620d83ad8cc7ce2f6994608abfdda0/lalrpop/build.rs#L35
Thus no lalrpop grammar gets built

dburgener:
Thanks for the deep investigation!  Took me a while to wrap my head around, but I guess the idea is that for non-test builds, we just use the bootstrapped binary, but for test, we use that bootstrapped binary to build the existing grammar and run tests against that.  That seems to make sense.

So you probably just want to turn the build.rs part into gating on `#[cfg(test)]` instead of the feature as well, right?  It does seem a little weird to me that the setup code is done unconditionally.  I'm not sure any of it is needed if we're not doing the test rebuild?  I'd say move that in the block while you're there unless I'm missing something.

Pat-Lafon:
> So you probably just want to turn the build.rs part into gating on `#[cfg(test)]` instead of the feature as well, right? It does seem a little weird to me that the setup code is done unconditionally. I'm not sure any of it is needed if we're not doing the test rebuild? I'd say move that in the block while you're there unless I'm missing something.

Yeah that all makes sense to me. Unfortunately there is a wrinkle that `#[cfg(test)]` doesn't work in build.rs scripts! See https://github.com/rust-lang/cargo/issues/1581 and https://github.com/rust-lang/cargo/issues/13294. It seems we can either add back the `test` feature flag I guess or follow the suggestion in issues to isolate the code you want to run in your build script(which is all of it for us) into a new crate that you use as a dev dependency.

dburgener:
> Unfortunately there is a wrinkle that #[cfg(test)] doesn't work in build.rs scripts!

Ah, that probably explains the original choice to add a test feature flag.

I think just restoring the test feature flag seems like the least bad option to me.  The whole "make a new crate as a dev-dependency" is a lot of complexity.

I guess the other question is: do we really need this functionality?  We check that we've updated the grammar on PRs, so we're not at risk of the latest grammar ending up untested.  It's a potential trap for local iteration though, which is a downside.  I'm not sure the cost of the test feature flag is enough to justify that tradeoff.

CONTRIBUTING.md
```diff
@@ -37,20 +35,8 @@ generate your own `lrgrammar.rs` parser. That's how you do it:
...
-$ cargo build/test -p lalrpop --all-features
-```
-
-When this flag is passed, cargo uses not `lalrpop/src/parser/lrgrammar.rs` but the newly generated
-`lrgrammar.rs` instead (generating it if needed). From now on you use this command.
-
-Once you're done with your work, all the tests are passed, and you are ready to finally commit
-you changes run
+Once you're done with your work, make sure that you are running against an
+updated version of the grammar.
```
dburgener, line 39: Maybe mention the update_grammar.sh script here as well?

Pat-Lafon:
I was thinking about it and it seems best to remove the special test path. It adds some extra complexity and potentially means that you are not syncing what you are testing with what you are trying to ship. More so, it's just kinda a bad user experience to have this non-standard hack... I think I've just always mindlessly used `cargo test` without thinking I needed additional flags

CONTRIBUTING.md
Pat-Lafon: That is the next line, it just doesn't appear in the diff?

dburgener:
Yeah, I want to give this a second read with more focus on the details, but I really like the approach you have now after a quick first pass.

CONTRIBUTING.md
dburgener: Oh, yep.  My bad.

dburgener (approved):

Issue #884, opened 2024-04-13
Memory leakage when applying recursive combinators


A4-Tacks:
```lalrpop
grammar;
A<I> = { "x" I "y" I "z", A<("." I)> }
P = A<()>;
```

lalrpop version: 0.20.0 ~ 0.20.2

dburgener:
I've been looking at this a bit.  The "leak" looks to me more like an infinite recursion, and it's during parser generation, rather than an issue in the generated parser.

Just dropping the provided grammar in a file, and running `lalrpop` on it should reproduce the issue.

It seems to be something about the fact that the recursive macro expansion isn't just passing through the args.  Replacing `A<("." I)>` in the above grammar with `A<I> "q"` for example seems to work fine, as expected.

I haven't looked at the macro expansion code much in the past, so I'm still trying to get my head around what we do there to figure out what's going wrong.

dburgener:
I think this is a fundamental limitation of how lalrpop implements macros.

lalrpop macros resolve the generic arguments into the actual callers, then treats every resolved generic as a grammar nonterminal.  So, for example `A<()>` becomes a nonterminal after macro expansion in the above grammar.  But we also need to resolve the generic in the recursive call.  So we need to add nonterminals for `A<("." ())>`, `A<("." ("." ()))>`, and so on infinitely.

Common recursive macros, such as the ones http://lalrpop.github.io/lalrpop/tutorial/006_macros.html terminate because once we resolve the generics in the macro definition, we end up with a nonterminal we already have (for example the macro def for `Tier<Op, NextTier>` references `Tier<Op, NextTier>`, which resolves the generics to the callers `Tier<ExprOp, Factor>` and `Tier<FactorOp, Term>`, but we already already have those nonterminals from resolving the macro name, so we don't need to keep resolving them.

I *think* that rewriting the grammar to treat the generic as a nonterminal would work.  Something like:

```
grammar;
I = { ("." I), () }
A = { "x" I "y" I "z" }
P = A;
```

That errors, because the original grammar is invalid totally besides this issue, but it doesn't recurse infinitely.

I'm not really sure about the following two questions:

1. Is detecting this case as simple as seeing if there's a macro call in a macro definition that would need to resolve as a new non-terminal  (ie the resolve call didn't occur outside a macro definition)?  Or does that miss something? (I haven't thought about the case of multiple macros calling each other yet.  That is definitely something to consider...)
[...]

PR #896, opened 2024-05-10
Detect deep recursion when expanding macros and error out

dburgener:

### DIFF
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -158,6 +158,12 @@ impl Configuration {
         self
     }
 
+    /// Set the max macro recursion depth.  As lalrpop is resolving a macro, it may discover new macros uses in the macro definition to resolve.  Typically deep recursion indicates a recursive macro use that is non-resolvable.  The default resolution depth is 200.
+    pub fn set_macro_recursion_limit(&mut self, val: u16) -> &mut Configuration {
+        self.session.macro_recursion_limit = val;
+        self
+    }
+
     /// Sets the features used during compilation, disables the use of cargo features.
     /// (Default: Loaded from `CARGO_FEATURE_{}` environment variables).
     pub fn set_features<I>(&mut self, iterable: I) -> &mut Configuration
--- a/lalrpop/src/normalize/macro_expand/mod.rs
+++ b/lalrpop/src/normalize/macro_expand/mod.rs
@@ -53,7 +53,9 @@ impl MacroExpander {
     }
 
     fn expand(&mut self, items: &mut Vec<GrammarItem>) -> NormResult<()> {
-        let mut counter = 0;
+        let mut counter = 0; // Number of items
+        let mut loop_counter = 0;
+
         loop {
             // Find any macro uses in items added since last round and
             // replace them in place with the expanded version:
@@ -61,10 +63,29 @@ impl MacroExpander {
                 self.replace_item(item);
             }
             counter = items.len();
+            loop_counter += 1;
 
             // No more expansion to do.
             if self.expansion_stack.is_empty() {
                 return Ok(());
+            }
+
+            if loop_counter > 200 {
+                // Too much recursion
+                match self.expansion_stack.pop() {
+                    Some(sym) => {
+                        return_err!(
+                            sym.span,
+                            "Exceeded recursion cap while expanding this macro."
+                        );
+                    }
+                    None => {
+                        // This should be impossible, since we just checked is_empty(), but if the
+                        // expansion stack is empty, we're done, so I guess return Ok instead of
+                        // panicing?
+                        return Ok(());
+                    }
+                }
             }
 
             // Drain expansion stack:
--- a/lalrpop/src/normalize/macro_expand/mod.rs
+++ b/lalrpop/src/normalize/macro_expand/mod.rs
@@ -72,21 +72,13 @@ impl MacroExpander {
 
             if loop_counter > recursion_limit {
                 // Too much recursion
-                match self.expansion_stack.pop() {
-                    Some(sym) => {
-                        return_err!(
+                // We know unwrap() is safe, because we just checked is_empty()
+                let sym = self.expansion_stack.pop().unwrap();
+                return_err!(
                             sym.span,
                             "Exceeded recursion cap ({}) while expanding this macro.  This typically is a symptom of infinite recursion during macro resolution.  If you believe the recursion will complete eventually, you can increase this limit using Configuration::set_macro_recursion_limit().",
                             recursion_limit
                         );
-                    }
-                    None => {
-                        // This should be impossible, since we just checked is_empty(), but if the
-                        // expansion stack is empty, we're done, so I guess return Ok instead of
-                        // panicing?
-                        return Ok(());
-                    }
-                }
             }
 
             // Drain expansion stack:
--- a/lalrpop/src/normalize/macro_expand/mod.rs
+++ b/lalrpop/src/normalize/macro_expand/mod.rs
@@ -15,7 +15,7 @@ use string_cache::DefaultAtom as Atom;
 #[cfg(test)]
 mod test;
 
-pub fn expand_macros(input: Grammar) -> NormResult<Grammar> {
+pub fn expand_macros(input: Grammar, recursion_limit: u16) -> NormResult<Grammar> {
     let input = resolve::resolve(input)?;
 
     let items = input.items;
@@ -32,7 +32,7 @@ pub fn expand_macros(input: Grammar) -> NormResult<Grammar> {
         .collect();
 
     let mut expander = MacroExpander::new(macro_defs);
-    expander.expand(&mut items)?;
+    expander.expand(&mut items, recursion_limit)?;
 
     Ok(Grammar { items, ..input })
 }
@@ -52,7 +52,7 @@ impl MacroExpander {
         }
     }
 
-    fn expand(&mut self, items: &mut Vec<GrammarItem>) -> NormResult<()> {
+    fn expand(&mut self, items: &mut Vec<GrammarItem>, recursion_limit: u16) -> NormResult<()> {
         let mut counter = 0; // Number of items
         let mut loop_counter = 0;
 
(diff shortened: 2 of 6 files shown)


dburgener:
Fixes #884, as per @Pat-Lafon's suggestion to just cap recursion depth.

Two things I'm not really sure about and would like feedback on:

1. The 200 recursion depth is pretty arbitrary.  When I was experimenting, I was noticing a delay and hitting ctrl+c around 8000 depth, give or take, and I imagine 200 should be plenty more than intentional recursion would hit?
2. I'd like it if the error message was a little more explicit about the problem and suggested a fix, but I'm not really sure about what suggestion to make.  Any thoughts on a better error message?

lalrpop/src/normalize/macro_expand/mod.rs
```diff
@@ -53,20 +53,41 @@ impl MacroExpander {
...
+                        return_err!(
+                            sym.span,
+                            "Exceeded recursion cap while expanding this macro."
+                        );
+                    }
+                    None => {
+                        // This should be impossible, since we just checked is_empty(), but if the
+                        // expansion stack is empty, we're done, so I guess return Ok instead of
+                        // panicing?
+                        return Ok(());
```
Pat-Lafon, line 86: Hmm, we could use the `unreachable!()` macro though I wouldn't be against unwrapping the option since self.expansion_stack must have at least one element

lalrpop/src/normalize/macro_expand/mod.rs
```diff
@@ -53,20 +53,41 @@ impl MacroExpander {
...
             }
             counter = items.len();
+            loop_counter += 1;
 
             // No more expansion to do.
             if self.expansion_stack.is_empty() {
                 return Ok(());
             }
 
+            if loop_counter > 200 {
```
Pat-Lafon, line 73: One way to somewhat alleviate the magic value here would for this to be the default, but have this be override-able via `Configuration`. What do you think?

Pat-Lafon (commented):
One thing to be mindful of is that this is technically a breaking change no? I would imagine that nobody would hit it(But if we knew that for sure, we wouldn't be considering the right value for loop iterations).

For the error message, I'm not 100% sure but if we did provide a form of Configuration for this, then we could say, "You have hit the macro expansion limit set in Lalrpop to conservatively avoid an infinite loop. If you believe that the macro expansion process will terminate, set a higher limit via `Configuration::set_macro_expansion_limit`.

lalrpop/src/normalize/macro_expand/mod.rs
dburgener: Yeah, this is a good idea.  I've added the config knob and pushed.

lalrpop/src/normalize/macro_expand/mod.rs
dburgener: Yeah, I think just unwrap() here does make sense after all.  Pushed a fix for that.

dburgener:
In terms of the default, I poked around a bit at some of our top downloaded dependencies, and I found [this grammar](https://github.com/osohq/oso/blob/main/polar-core/src/polar.lalrpop) which makes pretty extensive use of macros and nesting macros.  So I instrumented a build to count the loops in `macro_expand()`, and it took 10 loops to resolve all the macros.  That seems to me like a lot, I imagine most grammars only need 2-3 at most.

All that said, I don't think there's much downside to a higher value.  The major downside of continuing to run is growing ram usage.  I can do some more experiments, but I have a feeling based on my earlier experimentation like we can safely go up at least into the 1000s (or higher) before having noticeable ram consumption issues on modern machines.  So maybe a higher default value gives a little more wiggle room in case someone has extremely deeply nested macros, with the only downside being mildly worse performance in an error path.

I can do some more performance measurements and experimenting next week probably if we want to understand the tradeoffs on the default value better.

lalrpop/src/session.rs
```diff
@@ -58,6 +58,8 @@ pub struct Session {
     /// this value if we so choose.
     pub max_errors: usize,
 
+    pub macro_recursion_limit: u16,
```
Pat-Lafon, line 61: This commit looks great! Thanks. A minor nit is that we can say something like `/// Macro limit ensures that compilation terminates in a finite number of steps` since most of the other fields have public documentation.

Pat-Lafon:
Thanks for all the details. For sure I am interested in the other's input(@yannham @youknowone) since it is possibly breaking.

My initial thoughts are that we don't want to unnecessarily break people. This actually seems like a change that I wish we could leverage Crater for but I am having a hard time thinking of how someone would go over the 200 macro expansion limit. Hopefully for 99%+ of users this limit doesn't matter anyways.

Lalrpop is pre-1.0 so I think we should be more comfortable with listing these kinds of changes as possibly breaking even if it isn't in practice. I don't think the social cost is too high and we can take the opportunity to give users a heads up. (A random thought is to include a request for grammars if this change does break a user, I think we would want to add it to the test suite as a higher complexity use case.)

So lets bump the version number and I am also happy to raise the limit if there is a desire to do so.

Pat-Lafon:
> Of course, just calling it breaking and moving on is the less risky approach. If we want to go that way, we should consider if there's anything else breaking we are thinking about that would make sense to get in at the same time?

Is there something you had in mind? I'm not opposed, but given some of my reasoning above I'm pretty happy to just ship this. Being pre-1.0 means there is a little less pressure to package breaking changes together and the change we are making is probably barely breaking to begin with.

Pat-Lafon (approved):

lalrpop/src/api/mod.rs
```diff
@@ -158,6 +158,12 @@ impl Configuration {
         self
     }
 
+    /// Set the max macro recursion depth.  As lalrpop is resolving a macro, it may discover new macros uses in the macro definition to resolve.  Typically deep recursion indicates a recursive macro use that is non-resolvable.  The default resolution depth is 200.
```
yannham, line 161: It's really a small nitpick, but do we want to hardcode the default value in the comment? That's one more things to make sure to keep in sync, while I feel like it doesn't add much to the understanding of this recursion limit. One can always go to the definition of the `macro_recursion_limit` or hopefully grep their way to find the current default value set.

```suggestion
    /// Set the max macro recursion depth.  As lalrpop is resolving a macro, it may discover new macros uses in the macro definition to resolve.  Typically deep recursion indicates a recursive macro use that is non-resolvable.
```

lalrpop/src/normalize/macro_expand/mod.rs
```diff
@@ -52,21 +52,35 @@ impl MacroExpander {
...
+            if loop_counter > recursion_limit {
+                // Too much recursion
+                // We know unwrap() is safe, because we just checked is_empty()
+                let sym = self.expansion_stack.pop().unwrap();
+                return_err!(
+                            sym.span,
+                            "Exceeded recursion cap ({}) while expanding this macro.  This typically is a symptom of infinite recursion during macro resolution.  If you believe the recursion will complete eventually, you can increase this limit using Configuration::set_macro_recursion_limit().",
+                            recursion_limit
+                        );
+            }
```
yannham, line 82: I don't know if it's worth it, but note that you can get rid of the unwrap by grouping both checks in one pattern matching, something like:

```rust
match self.expansion_stack.pop() {
  // No more expansion to do.
  None => return Ok(()),
  // Too much recursion
  Some(sym) if loop_counter > recursion_limit => { ...code path of the if...},
  _ => (),
}
```

yannham (approved):
Thanks for taking care of this, @dburgener. My personal stance on backward compatibility is that I usually lean on the side of "we don't break backward compatibility if we fix a behavior that was entirely accidental and that nobody should sanely rely on (and wasn't advertised in the documentation)". Of course the line isn't always entirely clear (insert obligatory XKCD here), but otherwise our hands are too tied and the price to pay is to carry technical debt for no clear benefits.

If we would have been post-1.0, I think I would be ok with not incrementing the major version for this change (in balance with the other too costly alternatives: do nothing and preserve the old behavior or increment the major version for something that won't actually break any real world program).

However, as @Pat-Lafon says, we don't have this issue here - it's cheap to increment the version in pre-1.0, so I'm fine with bumping the minor (but thus making it a breaking change in pre-1.0) version.

lalrpop/src/api/mod.rs
dburgener: Listing the default is consistent with all the other config functions that take an argument.  I'm definitely sympathetic to the "single point of truth" case to omit it, but documentation is much more visible to users than grepping through source files, and I think including the defaults in the docs add value.

lalrpop/src/normalize/macro_expand/mod.rs
dburgener: Yeah, I did consider that, but I do think that the code is more readable keeping these as separate.  In my mind at least, these are logically two separate checks.

1. Are we all done?
2. Have we exceeded the recursion limit?

The fact that check 2 implicitly needs info about check 1 is really just a detail, not a fundamental bit of code logic.

Also, I think the code as you've written it has a bug - if `pop()` returns a value, and we haven't exceeded the limit, then this would discard that value, when instead we want to use it in the ensuing `while` loop.  It's a fixable bug, but fixing it would further complicate the code.

lalrpop/src/session.rs
dburgener: Good catch, thanks.  Pushed.

dburgener:
Regarding the breaking aspect, yeah, I think the "we're pre-1.0" case is convincing enough.  I'm fine with shipping this as breaking just in case.

I didn't have in mind anything else breaking we're looking to do, no.  I just like batching breaking changes if we can/it makes sense, so wanted to flag the option if anyone else had anything in mind.

lalrpop/src/normalize/macro_expand/mod.rs
yannham: I agree that the current code is more readable, however, sometimes people like to minimize the number of `unwrap` by helping the compiler with proving that indeed this doesn't need to be there. So it's really a trade-off. I personally have no strong opinion, so I'll let you pick the solution you think is better :slightly_smiling_face:

lalrpop/src/api/mod.rs
yannham: Fair enough. Works for me, then

lalrpop/src/normalize/macro_expand/mod.rs
dburgener: My opinion remains unchanged from above :)  Since you left it to me, I'm going to go ahead and merge this as is.  Thanks for the review!

PR #899, opened 2024-05-18
Update to itertools 0.13

dburgener:
Switch deprecated group_by() to chunk_by()

The change in name was made because chunk_by() better implies that the function groups *consecutive* elements (rather than searching through the iterator and funding *all* matching elements).  This raises the question of whether the intention in the original group_by() was correct, since the change was made to avoid subtle bugs around assuming that this grouped all entries rather than consecutive entries.  In the two instances we have of this function call, the first one has a comment specifically calling out a desire to do this consecutively.  In the second, the iterator was previously sorted by the chunking value, making consecutive equivalent to all, so I believe we are free of this potential bug.

### DIFF
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -601,7 +601,7 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                 .map(|k2| state_lookup(k, k2))
                 .enumerate()
                 // Group consecutive indices so we can compress then as a..=b
-                .group_by(|(_, (next_state, _))| *next_state);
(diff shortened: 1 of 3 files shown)


lalrpop/src/lr1/codegen/parse_table.rs
```diff
@@ -601,7 +601,7 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
                 .map(|k2| state_lookup(k, k2))
                 .enumerate()
                 // Group consecutive indices so we can compress then as a..=b
```
yannham, line 603:
```suggestion
                // Group consecutive indices so we can compress them as a..=b
```

yannham (approved):

Issue #856, opened 2023-11-17
Replace term with anstream?


Pat-Lafon:
Not a high priority but I noticed that a more up-to-date `Cargo.lock` brings in a second version of `bitflags` because of the dependency on `term`. This got me looking into this crate which generally doesn't seem to be well maintained(https://github.com/Stebalien/term/issues/93) and as since been replaced in the ecosystem afaik(`Termcolor` was the original replacement, but maybe people are moving away from that for `anstream` https://github.com/rust-lang/cargo/pull/12751).

It would probably benefit Lalrpop to refactor the cli to use `anstream` instead of `term` and it might even clean up that area of the code base a little bit with it's hacks like `FakeTerminal`

dburgener:
I started briefly poking at this.  Part of the challenge here is that our usage of term is also tied into our ASCII canvas dependency (https://github.com/nikomatsakis/ascii-canvas).  Ascii canvas is also written by Niko like lalrpop, and it looks like it was originally part of lalrpop, but since it had independent use, he spun it out into its own separate crate.  The term dependency is used across the API between lalrpop and ASCII canvas.  Probably the strategy here needs to be to first go update ASCII canvas (I imagine we can ping Niko and get that change in even though he's presumably not actively managing the project).  Then that could get released as a breaking change, since the API would move away from term.  Then we can make a change here to update our ASCII canvas version while also dropping the term dependency.

I might work on this at some point, but it seems like a pretty large task, so I'm not able to commit the time right now.

nikomatsakis:
Whoops. I'd happily pull ascii-canvas into lalrpop org, I doubt it has many independent users.

Pat-Lafon:
> I started briefly poking at this. Part of the challenge here is that our usage of term is also tied into our ASCII canvas dependency (https://github.com/nikomatsakis/ascii-canvas). Ascii canvas is also written by Niko like lalrpop, and it looks like it was originally part of lalrpop, but since it had independent use, he spun it out into its own separate crate. The term dependency is used across the API between lalrpop and ASCII canvas. Probably the strategy here needs to be to first go update ASCII canvas (I imagine we can ping Niko and get that change in even though he's presumably not actively managing the project). Then that could get released as a breaking change, since the API would move away from term. Then we can make a change here to update our ASCII canvas version while also dropping the term dependency.
> 
> I might work on this at some point, but it seems like a pretty large task, so I'm not able to commit the time right now.

[...]

PR #901, opened 2024-05-26
First pass at update

Pat-Lafon:
This is a first pass at #856 though it is not ready to merge until https://github.com/lalrpop/ascii-canvas/pull/7 is settled and maybe a new version is pushed. The goal of this pr is to replace `term` with `anstream`.

If possible, I am also trying to remove this `fake_term.rs` usage. I think `StripStream` is the moral equivalent but I haven't done too much investigation.


dburgener:
I'm testing this out, and I have two questions:

1. We don't actually seem to use color much?  I guess I haven't really paid attention too much in the past, but playing with it now, I see that we bold "Conflict detected" on conflicts, and that's all I've found.  Is there other coloring I should be looking at?
2. With this PR, writing to a normal pty, the bold isn't terminating after "Conflict detected".  The beginning ("processing file" etc) is normal, then at "Conflict detected" the entire rest of the output is bolded.

Pat-Lafon:
Sorry, I've been busy. I'll see if I can get around to this and page this back in. Thanks for being on top of things!

PR #903, opened 2024-05-30
Cut a 0.21.0 release

Pat-Lafon:
As has been noted in other issues, there have been two nice bug fixes that probably warrant a new release. We can wait on #901(though I don't think that's a breaking change) to bundle everything together or we can release what we currently have now and then have a point release when #901 gets merged.

Feel free to make any changes to the release notes as needed.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,29 @@
+<a name="0.21.0"></a>
+## 0.21.0 (2024-05-30)
+
+Since the last release, a fair number of the commits have been focussed on cleaning
+up and improving LALRPOP's documentation. Shout out to Yudai Takada, George
+White, and Dinu Blanovschi.
+
+#### Bugfixes
+
+* A long-standing bug where LALRPOP would throw a "no entry found for key"
+  exception when trying to handle a particular reduce/reduce conflict has been
+  resolved.
+* LALRPOP will stop expanding macros infinitely during build time via a new
+  `macro_expansion_limit`.
+
+#### Compatibility note
+
+Adding a limit to the number of times that LALRPOP will attempt to expand a
+macro is technically a breaking change. However, the default limit of `200`
+should be more than enough for the grammars we are currently aware of(which
+almost always need a limit of less than 5). This limit is customizable via `Configuration::set_macro_expansion_limit`.
+
+If you have a grammar that uses a significant amount of macro expansion steps,
+we would be very interested in a pr that adds it to the test suite.
+
+
 <a name="0.20.2"></a>
 ## 0.20.2 (2024-02-**)
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -5,6 +5,14 @@ Since the last release, a fair number of the commits have been focused on cleani
 up and improving LALRPOP's documentation. Shout out to Yudai Takada, George
 White, and Dinu Blanovschi.
 
+#### Features
+
+* LALRPOP now throws an error in more cases where it would previously just write
+  out an error message and exit.
+* `lalrpop::process_src` is now the recommended function to use in `build.rs`
+  files. Previously the documentation incorrectly suggested that
+  `lalrpop::process_root` looked in `./src` instead of `.`
+
 #### Bugfixes
 
 * A long-standing bug where LALRPOP would throw a "no entry found for key"
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,7 +1,7 @@
 <a name="0.21.0"></a>
 ## 0.21.0 (2024-05-30)
 
-Since the last release, a fair number of the commits have been focussed on cleaning
+Since the last release, a fair number of the commits have been focused on cleaning
 up and improving LALRPOP's documentation. Shout out to Yudai Takada, George
 White, and Dinu Blanovschi.
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -16,7 +16,7 @@ White, and Dinu Blanovschi.
 
 Adding a limit to the number of times that LALRPOP will attempt to expand a
 macro is technically a breaking change. However, the default limit of `200`
-should be more than enough for the grammars we are currently aware of(which
+should be more than enough for the grammars we are currently aware of (which
 almost always need a limit of less than 5). This limit is customizable via `Configuration::set_macro_expansion_limit`.
 
 If you have a grammar that uses a significant amount of macro expansion steps,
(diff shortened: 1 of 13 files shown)


RELEASES.md
```diff
@@ -1,3 +1,29 @@
...
+
+Since the last release, a fair number of the commits have been focussed on cleaning
+up and improving LALRPOP's documentation. Shout out to Yudai Takada, George
+White, and Dinu Blanovschi.
+
+#### Bugfixes
+
+* A long-standing bug where LALRPOP would throw a "no entry found for key"
+  exception when trying to handle a particular reduce/reduce conflict has been
+  resolved.
```
dburgener, line 12:
```suggestion
  exception when trying to handle certain grammars has been resolved.
```

I don't love losing specificity, but I think reduce/reduce conflict is misleading.  This is a reduce/reduce conflict from a LALR perspective that may or may not be a reduce/reduce conflict after we split the states to full LR(1).  So from a user perspective, they could hit this error when their grammar actually has a reduce/reduce conflict, or with a valid grammar that is LR(1) but not LALR.  I'm not sure how to capture all that without going into too much detail, but I think as originally written it implies that every time we hit that panic, the end result should have been a reduce/reduce conflict error, which isn't the case.

dburgener:
I'm good either way on whether we get #901 in.  The main advantage I see to including it in a breaking release, even though it isn't breaking, is that these sorts of user interface sorts of changes often miss a lot of the regression testing paths, so I have less confidence in our testing than I normally would.  Since we're doing a breaking release anyways, having user eyes on the upgrade is sort of nice to have.  That said, it's not a big concern, so I don't think we need to hold the release up too long.

I do hope to do a full pass on #901 tomorrow if everything goes as planned, so my vote would be to revisit timing after that.  If we think #901 will take a while longer, we can send the release, but if we can just merge it tomorrow or over the weekend, maybe it makes sense to hold up this release another couple of days to get it in?

youknowone (approved):
👍

dburgener (approved):

dburgener:
Having looked a little more at #901, and thought more over the weekend, I think my concerns about putting #901 in a non-breaking release are overblown.  I say lets ship this and bring #901 in at our leisure.

Pat-Lafon:
#906 and #902 will make it in. Is it accurate to say that https://github.com/lalrpop/lalrpop/blob/master/publish.sh doesn't get used anymore?

youknowone:
I didn't use the script for the last release

RELEASES.md
```diff
@@ -1,3 +1,28 @@
...
+
+#### Compatibility note
+
+Adding a limit to the number of times that LALRPOP will attempt to expand a
+macro is technically a breaking change. However, the default limit of `200`
+should be more than enough for the grammars we are currently aware of(which
+almost always need a limit of less than 5). This limit is customizable via `Configuration::set_macro_expansion_limit`.
+
+If you have a grammar that uses a significant amount of macro expansion steps,
+we would be very interested in a pr that adds it to the test suite.
```
yannham, line 23:
```suggestion
we would be very interested in a PR that adds it to the test suite.
```

yannham (approved):

RELEASES.md
```diff
@@ -1,3 +1,28 @@
+<a name="0.21.0"></a>
+## 0.21.0 (2024-05-30)
+
+Since the last release, a fair number of the commits have been focussed on cleaning
```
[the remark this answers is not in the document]
Pat-Lafon: british english I think, whoops

dburgener (approved):

dburgener:
Just pinging on this.  I'm assuming the reason we're waiting is clarifying the process in #914?

Given a lack of additional comments there, it seems to me as though sending this out with `cargo-release` and then manually pushing a tag makes sense.  Perhaps that will turn up other considerations in #914.  Unless there's something else you want to hold the release for?

Pat-Lafon:
> Just pinging on this. I'm assuming the reason we're waiting is clarifying the process in #914?
> 
> Given a lack of additional comments there, it seems to me as though sending this out with `cargo-release` and then manually pushing a tag makes sense. Perhaps that will turn up other considerations in #914. Unless there's something else you want to hold the release for?

Just had to rebase and update the version for lexer-modes as we previously discussed. Might be useful to try out cargo-release for this version and see what we run into.

Most of the changes since last rebase seem to be documentation related/not actively visible anyways. Lets release 🚀

dburgener:
@Pat-Lafon Are you planning on running the release?  Any hold-ups/blockers at this point?

Pat-Lafon:
I looked through the history since July 10th and added two more items to the changelog.

I looked at `cargo-release`. It seems their workspace flag also tries to release the `doc/*` examples. We could in the future use this tool to also update our version numbers which might be convenient(though one would have to worry about the version numbers in the documentation.

Looking at something like:
```sh
cargo release --dry-run --package lalrpop-util --tag-name "{{version}}"
cargo release --dry-run --package lalrpop --tag-name "{{version}}"
```

dburgener:
All sounds good to me.

Pat-Lafon:
Ok, lets try to make this happen then

Pat-Lafon:
I have released v0.21.0 and I'll make a quick discussion post about it. 2 things to note about the above. First the commands wait until the crate is available on crates.io so no need to have a sleep. Second, you can't create a tag twice, but it will also try to make a default tag if you drop this flag... whoops.

dburgener:
> Second, you can't create a tag twice, but it will also try to make a default tag if you drop this flag... whoops.

What's the impact of this?  I feel like I'm missing something about the scenario.  Maybe I'm just being dense.  You want cargo-release to create the tag, and you want to run cargo-release exactly once without dry-run, right?

Pat-Lafon:
To clarify, if you try to publish lalrpop-util and then lalrpop with the commands above, then the second will fail because the tag for that version(tag name `0.21.0`) already exists. So I re-ran while dropping the `tag-name` part which then defaults to creating a tag called `lalrpop-0.21.0` which I had to go delete.

youknowone:
Thank you! 😆

dburgener:
> To clarify, if you try to publish lalrpop-util and then lalrpop with the commands above, then the second will fail because the tag for that version(tag name 0.21.0) already exists. So I re-ran while dropping the tag-name part which then defaults to creating a tag called lalrpop-0.21.0 which I had to go delete.

Thanks for clarifying.  It looks like in #914 you added the `--no-tag` option in the second one.  That should resolve it, so this is pretty much a non-issue moving forward, right?

Pat-Lafon:
Yes, this should hopefully be resolved going forward.

PR #907, opened 2024-06-14
Fix clippy issues for rust 1.81

dburgener:


dburgener:
Definitely no need to get this in for 0.21.  Should be low impact either way.

Pat-Lafon (approved):

PR #910, opened 2024-06-18
Document $attr in lalrpop_mod!

### LINKED ISSUE #909: Document  `$attr`  in `lalrpop_mod!`
The rust doc for `lalrpop_mod!` does not show an example in which the `$attr` is used.

A relevant example might be something like

```rust
lalrpop_mod!(#[allow(clippy::vec_box)] pub calculator);
```

I do understand why this is left out of the tutorial, but more fully documenting `lalrpop_mod!` will help those who do run clippy on the tutorial.

jpgoldberg:
Resolves #909  

This is a documentation only change. All tests pass on using

```
stable-aarch64-apple-darwin (default)
rustc 1.78.0 (9b00956e5 2024-04-29)
```

To see effect, run `cargo doc` and then look at `./target/doc/lalrpop_util/macro.lalrpop_mod.html`

### DIFF
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -182,7 +182,7 @@ pub struct ErrorRecovery<L, T, E> {
 ///
 /// // define a public module
 /// lalrpop_mod!(pub parser);
-/// 
+///
 /// // specify attribute for the generated module
 /// // to suppress clippy vec_box warnings
 /// lalrpop_mod!(#[allow(clippy::vec_box)] parser);
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -182,6 +182,10 @@ pub struct ErrorRecovery<L, T, E> {
 ///
 /// // define a public module
 /// lalrpop_mod!(pub parser);
+/// 
+/// // specify attribute for the generated module
+/// // to suppress clippy vec_box warnings
+/// lalrpop_mod!(#[allow(clippy::vec_box)] parser);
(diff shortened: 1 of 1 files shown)


jpgoldberg:
- rustfmt test should now pass after commit https://github.com/lalrpop/lalrpop/pull/910/commits/db2f6732837d4ec93b96d07d423fe1958645ba3f
- I am not sure exactly why GitHub is still considering my commits unverified. Perhaps I uploaded the wrong public key.

lalrpop-util/src/lib.rs
```diff
@@ -182,6 +182,10 @@ pub struct ErrorRecovery<L, T, E> {
 ///
 /// // define a public module
 /// lalrpop_mod!(pub parser);
+///
+/// // specify attribute for the generated module
+/// // to suppress clippy vec_box warnings
+/// lalrpop_mod!(#[allow(clippy::vec_box)] parser);
```
Pat-Lafon, line 188: The attribute support is more generic than just suppressing clippy warnings, so I don't love the "to suppress clippy" line. Maybe something more generic "to add onto the generated code".

You aren't just limited to one attribute iirc.

Pat-Lafon (approved):
Thanks for the pr! As you mention in #909, it might be worth adding to the tutorial as well. More documentation can't hurt.

We might nit on the exact wording but I like this change.

jpgoldberg:
Please change the text as you see fit. My text was specific to my particular example, but as you note it implies a far too limited view of that that argument can be.

Issue #908, opened 2024-06-18
PermissionDenied when unwritable subdirectories contain *.lalrpop


dkg:
https://bugs.debian.org/1069212 shows a problem i'm having on debian systems, where a common build practice is to have a link farm within `./debian/cargo_registry` that points back to a bunch of directories in `/usr/share/cargo/registry`.

lalrpop's `build.rs` seems to want to traverse down through the current directory and try to tinker with any file named `*.lalrpop`.

Since the normal user can't write to anything in `/usr/share/cargo/registry`, this yields a PermissionDenied error.

I'm not sure what the right fix is, but if lalrpop would instead hunt for `*.lalrpop` files in `./src/` instead of all of `./` it would make the problem go away for me for now ☺

I'm open to other suggestions.

dkg:
@decathorpe observed that `build/mod.rs.html` lines 56-62 are where the recursive directory walk starts in `.` and not in `./src`

decathorpe:
Yes. It is slightly confusing to me is why session.in_dir and root_dir are different paths at all:
https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/build/mod.rs#L55-L60

This results in the recursive directory walk finding all .lalrpop files under $PWD instead of $PWD/src, which seems to be the actual intention?

dburgener:
A few initial thoughts.  Curious what others think on this one:

* The term "root_dir" in `build/mod.rs` `process_dir()` is a little misleading.  It's the "root" in the sense of "root to search from" rather than "project root", whereas in `api/mod.rs` "root" generally means "project root".  The behavior of `build::process_dir()` is to check the path passed in from the parent, which in the common case comes from `api::process_root()`, which uses the current dir (ie not src)
* The "current dir" behavior is both implied by the name `process_root()` and is specified in the [lalrpop documentation](https://docs.rs/lalrpop/0.20.2/lalrpop/fn.process_root.html)
* Users (such as sequoia) that prefer to process in the src dir should be able to achieve that today with `Configuration::new().use_cargo_dir_conventions().process()` in their `build.rs`
[...]

PR #911, opened 2024-06-19
Add a new process_src() public helper to look in the source directory.

dburgener:
Update the docs to recommend this as the top choice for setting up lalrpop.

Prior to this commit, the book incorrectly stated that `process_root()` looked in `./src`, while the documentation comments in the code correctly identified it as parsing the current directory.  Switching the book to recommend `process_src()` addresses this inconsistency as well.

### DIFF
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -245,10 +245,21 @@ impl Configuration {
 
 /// Process all files in the current directory, which -- unless you
 /// have changed it -- is typically the root of the crate being compiled.
+/// If your project only builds one crate and your files are in a ./src directory, you should use
+/// `process_src()` instead
 ///
 /// Equivalent to `Configuration::new().process_current_dir()`.
 pub fn process_root() -> Result<(), Box<dyn Error>> {
     Configuration::new().process_current_dir()
+}
+
+/// Process all files in ./src.  In many cargo projects which build only one crate, this is the
+/// normal location for source files.  If you are running lalrpop from a top level build.rs in a
+/// project that builds multiple crates, you may want `process_root()` instead.
+pub fn process_src() -> Result<(), Box<dyn Error>> {
+    Configuration::new()
+        .set_in_dir(Path::new("./src"))
+        .process()
 }
 
 /// Deprecated in favor of `Configuration`.
--- a/doc/calculator/build.rs
+++ b/doc/calculator/build.rs
@@ -1,3 +1,3 @@
 fn main() {
-    lalrpop::process_root().unwrap();
+    lalrpop::process_src().unwrap();
 }
--- a/doc/lexer-modes/build.rs
+++ b/doc/lexer-modes/build.rs
@@ -1,3 +1,3 @@
 fn main() {
-    lalrpop::process_root().unwrap();
+    lalrpop::process_src().unwrap();
 }
--- a/doc/lexer/build.rs
+++ b/doc/lexer/build.rs
@@ -1,3 +1,3 @@
 fn main() {
-    lalrpop::process_root().unwrap();
+    lalrpop::process_src().unwrap();
 }
--- a/doc/nobol/build.rs
+++ b/doc/nobol/build.rs
@@ -1,3 +1,3 @@
 fn main() {
-    lalrpop::process_root().unwrap();
+    lalrpop::process_src().unwrap();
 }
--- a/doc/src/advanced_setup.md
+++ b/doc/src/advanced_setup.md
@@ -5,11 +5,11 @@ like this:
 
 ```rust
 fn main() {
-    lalrpop::process_root().unwrap();
+    lalrpop::process_src().unwrap();
 }
 ```
 
-This `process_root()` call simply applies the default configuration:
+This `process_src()` call simply applies the default configuration:
 so it will transform `.lalrpop` files into `.rs` files *in-place* (in
 your `src` directory), and it will only do so if the `.lalrpop` file
 has actually changed. But you can also use the
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -26,18 +26,26 @@ file that looks like:
 
 ```rust
 fn main() {
-    lalrpop::process_root().unwrap();
+    lalrpop::process_src().unwrap();
 }
 ```
 
 (If you already have a `build.rs` file, you should be able to just
-call `process_root` in addition to whatever else that file is doing.)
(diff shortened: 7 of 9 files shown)


dburgener:
closes #908

dburgener:
Adding a new public API element isn't breaking, but would require bumping the minor version in post-1.0.  If we merge this, we might want to include it in a minor version bump for easy cargo version specification for users relying on the new api.

Pat-Lafon:
Do you think there is a goal of deprecating `process_root` in favor of `process_src`? If I understand this, `process_src` is the happy/typical path, and if you want any path tweaking, you can use `Configuration`?

dburgener:
> Do you think there is a goal of deprecating process_root in favor of process_src? If I understand this, process_src is the happy/typical path, and if you want any path tweaking, you can use Configuration?

I went into this thinking that would be the goal, but now I think they're both independently useful.  Namely, `process_root()` is helpful in a case like lalrpop, where multiple crates are built in the same workspace (so there's no `src` at the top level).  I think having both, but recommending `src` in the common case makes sense to me, because 1. src will hard fail (not find the lalrpop file) if someone should have used root, but using root instead of src will have subtler issues, as seen in the report.  2. I suspect that single-crate workspaces are (much?) more common.

youknowone:
> 2.  I suspect that single-crate workspaces are (much?) more common.

single-crate workspace will be less common. But probably multi-crate workspaces which compile the same lalrpop file in multiple crate will be even less common. Multi-crate workspace which contains a single crate compiling a single lalrpop file still will use process_src.

Pat-Lafon:
One more question. I wonder how much benefit there is in having this separation/indirection between users and `Configuration`.

The main reason for this in my mind would be if we wanted to make enough changes to `Configuration` such that we were deprecating it or some of it's methods in favor of a new api. Then we could make the swap in the implementation of `process_src` with less real world breakage.

On the other hand, I don't think a `Configuration::new().set_in_dir("./src").process().unwrap()` is the worst default. It does make it a little easier to understand why lalrpop may not be finding your grammars(if they are in `grammars` instead of `src` as a convoluted example). It would be a little more consistent in leading users to more advanced features like [emit_rerun_directives](https://docs.rs/lalrpop/latest/lalrpop/struct.Configuration.html#method.emit_rerun_directives) or [generate_in_source_tree](https://docs.rs/lalrpop/latest/lalrpop/struct.Configuration.html#method.generate_in_source_tree). (I don't really use the rest of the options but I guess they are there for a reason)

lalrpop/src/api/mod.rs
```diff
@@ -245,12 +245,23 @@ impl Configuration {
...
 pub fn process_root() -> Result<(), Box<dyn Error>> {
     Configuration::new().process_current_dir()
 }
 
+/// Process all files in ./src.  In many cargo projects which build only one crate, this is the
+/// normal location for source files.  If you are running lalrpop from a top level build.rs in a
+/// project that builds multiple crates, you may want `process_root()` instead.
+pub fn process_src() -> Result<(), Box<dyn Error>> {
+    Configuration::new()
+        .set_in_dir(Path::new("./src"))
```
Pat-Lafon, line 261: Nit: Calling `Path::new` is a little unnecessary as it will be done automatically for `&str`.

Pat-Lafon (approved):
None of my comments should be construed as blocking this pr. Please move forward as you/the others see fit.

dburgener:
> The main reason for this in my mind would be if we wanted to make enough changes to Configuration such that we were deprecating it or some of it's methods in favor of a new api. Then we could make the swap in the implementation of process_src with less real world breakage.

I think of it as mainly a convenience function.  As a user, I like to grab a default and just get it working in early development.  If the default works for me, I can keep it, and if not, I can always upgrade to the full `Configuration` version once I have my bearings in the project a little more.

> On the other hand, I don't think a Configuration::new().set_in_dir("./src").process().unwrap() is the worst default.

Yeah, it's fine.  If we want to go that way, I'd be happy to submit a PR updating the book's rec to this.  As I said above, I do think that recommending `src` as the default location seems like a better strategy than recommending the root as the default.  Youknowone pushed back on the asserting that single-crate workspaces are more common, and I don't have data to back that up, but either way, preferring `src` as the default fails hard if they got it wrong, and preferring `.` as the default may fail subtly, or simply perform worse.

If we're recommending `src` as the default, it's a little weird to only have the convenience function `process_root()`, and not `process_src()` as well, and as I said above, deprecating `process_root()` seems to remove something that does have a lot of value.

I guess in terms of moving forward, we should probably agree on:

1. What is our recommendation to new users in the quick start guide?  As stated earlier, I think we should recommend `src` as a first choice, but state both options.
2. Do we want convenience functions like `process_root()` and `process_src()`?  If so, which ones?  (And if not `process_root()`, do we want to deprecate, or just deemphasize?)

I'm happy to hear push-back.  If we add to the public API, and then change our minds about it later, that has costs, so getting it right initially is definitely worth some time and back and forth.

Pat-Lafon:
> If we're recommending `src` as the default, it's a little weird to only have the convenience function `process_root()`, and not `process_src()` as well, and as I said above, deprecating `process_root()` seems to remove something that does have a lot of value.

Sure, I've both sorta come around on this. If it makes users lives easier then I'm willing to support the extra convenience function since it is simple enough. I think my point can be mostly addressed by having more documentation pointing from these functions to `Configuration`.

> I guess in terms of moving forward, we should probably agree on:
> 
>     1. What is our recommendation to new users in the quick start guide?  As stated earlier, I think we should recommend `src` as a first choice, but state both options.

Recommend `src` since it probably is the right default. Let users know the other two options to reach for when they fall off of this happy path.

>     2. Do we want convenience functions like `process_root()` and `process_src()`?  If so, which ones?  (And if not `process_root()`, do we want to deprecate, or just deemphasize?)

I guess there isn't a whole lot to be gained from deprecating `process_root` if this is what most users have. But encouraging people to use `process_src` seems worthwhile to me.

> I'm happy to hear push-back. If we add to the public API, and then change our minds about it later, that has costs, so getting it right initially is definitely worth some time and back and forth.

👍

dburgener:
@Pat-Lafon I know your approval is here already, but we did have some back and forth and discussion since then.  Does this seem ready to merge from your perspective?  Would you like to wait for any additional chime-ins from other maintainers?

If we feel good about this, it might be nice to get the public API addition in a release that increments the minor version anyways, so we could merge it before 0.21.

Pat-Lafon:
Yup lets merge this. I'll make a small doc suggestion in a separate pr.

PR #913, opened 2024-06-23
Make clippy happy in generated lookaround action code

wuxianucw:
A missing case of #822

### DIFF
--- a/lalrpop/src/build/action.rs
+++ b/lalrpop/src/build/action.rs
@@ -148,6 +148,7 @@ fn emit_lookaround_action_code<W: Write>(
     _defn: &r::ActionFnDefn,
     data: &r::LookaroundActionFnDefn,
 ) -> io::Result<()> {
+    rust!(rust, "#[allow(clippy::needless_lifetimes)]");
     rust.fn_header(
         &r::Visibility::Priv,
         format!("{}action{}", grammar.prefix, index),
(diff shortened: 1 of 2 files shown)


Pat-Lafon:
Hey, thanks for the pr! Since you are modifying the outputted grammar, you will need to run [./update_lrgrammar.sh](https://github.com/lalrpop/lalrpop/blob/master/update_lrgrammar.sh) to get the tests to pass.

dburgener:
Thanks for the PR!  This seems initially to make sense.  Do you happen to have a publicly viewable grammar that triggers the clippy violation here?  I'm confused as to why we aren't seeing this clippy warning locally when we build the lalrpop grammar.

The CI failure can be fixed by running the `./update_lrgrammar.sh` script and adding the updated lrgrammar.rs to your PR.

wuxianucw:
> Thanks for the PR! This seems initially to make sense. Do you happen to have a publicly viewable grammar that triggers the clippy violation here? I'm confused as to why we aren't seeing this clippy warning locally when we build the lalrpop grammar.

Yes. I followed [this guide](https://lalrpop.github.io/lalrpop/generate_in_source.html) to generate a `grammar.rs` in source tree for learning reasons. I did not add the `#[allow(clippy::all)]` annotation when declaring `mod grammar`. Since LALRPOP has such annotations (`rustfmt::skip`, `dead_code` and `clippy::all`), this clippy warning should not be seen.

dburgener (approved):

dburgener:
Thanks for the explanation.  I think this all makes sense as far as I'm concerned.  Thanks for the PR!

PR #914, opened 2024-06-23
Remove outdated script and it's reference

Pat-Lafon:
As mentioned in https://github.com/lalrpop/lalrpop/pull/903#issuecomment-2156725838, I think `publish.sh` is no longer in use. Maybe the release process is obvious to others(like just `cargo-release`?).


dburgener:
I was waiting for @youknowone to chime in since he's done the past few releases, but here are my few cents:

1. I think getting rid of the script makes sense.  Doing the version update via PR seems cleaner than doing it via direct push as the script seems to want
2. Assuming we want a tag in the local repo (which we have historically) I think we still need to direct push the tag.
3. The biggest gotcha with `cargo publish` in my experience is that it isn't git aware, so if you have a not clean working directory it will publish any local files along with the package.  I haven't used `cargo-release` before, but the docs sound like it checks for this, which is great.
4. The `publish.sh` script makes sure to publish `lalrpop-util` first, waits 25 seconds and then publishes `lalrpop`.  I'm not super clear on why that's necessary.  I would think that in situations where `lalrpop-util` and `lalrpop` have sequencing concerns that should be breaking?  Or do we have some requirement somewhere that users must keep `lalrpop-util` >= `lalrpop`?  Despite my lack of clarity on why it's necessary, I'm inclined to have the directions say to publish `lalrpop-util` first, on the basis of [Chesterton's Fence](https://fs.blog/chestertons-fence/)
5. I would mention in the updated instructions that after you tag the release locally you need to direct push that tag to github

Pat-Lafon:
>     1. I think getting rid of the script makes sense.  Doing the version update via PR seems cleaner than doing it via direct push as the script seems to want

+1

>     2. Assuming we want a tag in the local repo (which we have historically) I think we still need to direct push the tag.

I think this kind of book-keeping is what I would want to automate. If youknowone already has a process then maybe it can just be codified.

>     3. The biggest gotcha with `cargo publish` in my experience is that it isn't git aware, so if you have a not clean working directory it will publish any local files along with the package.  I haven't used `cargo-release` before, but the docs sound like it checks for this, which is great.

In the absence of adding in the current process, moving to cargo-release or on of the variation in the ecosystem might make a lot of sense. Bringing in outside code for the release process should be done with caution, but I'm a lot more worried about missing an easily automate-able step. Plus, it would mean getting ecosystem level developments like semvar-checks and other tools.

>     4. The `publish.sh` script makes sure to publish `lalrpop-util` first, waits 25 seconds and then publishes `lalrpop`.  I'm not super clear on why that's necessary.  I would think that in situations where `lalrpop-util` and `lalrpop` have sequencing concerns that should be breaking?  Or do we have some requirement somewhere that users must keep `lalrpop-util` >= `lalrpop`?  Despite my lack of clarity on why it's necessary, I'm inclined to have the directions say to publish `lalrpop-util` first, on the basis of [Chesterton's Fence](https://fs.blog/chestertons-fence/)

I was actually wondering if this would no longer be needed now that lalrpop captures all of this into a workspace https://github.com/rust-lang/cargo/issues/1169. It does seem to be a real issue, but also might be handled with the above mentioned tools https://github.com/rust-lang/cargo/issues/9507#issuecomment-1054288944

>     5. I would mention in the updated instructions that after you tag the release locally you need to direct push that tag to github

Done

youknowone:
I am sorry. I forgot to check this issue. 

When I was publishing it, I didn't recognize there was a script. I would try it if I knew that. But some part of the scripts are not necessary anymore.

> 3. The biggest gotcha with `cargo publish` in my experience is that it isn't git aware, so if you have a not clean working directory it will publish any local files along with the package.  I haven't used `cargo-release` before, but the docs sound like it checks for this, which is great.

`cargo publish` checks if the working directory is dirty or not. If we don't pass `--allow-dirty`, it should stop there. 

> 4. The `publish.sh` script makes sure to publish `lalrpop-util` first, waits 25 seconds and then publishes `lalrpop`.  I'm not super clear on why that's necessary.  I would think that in situations where `lalrpop-util` and `lalrpop` have sequencing concerns that should be breaking?  Or do we have some requirement somewhere that users must keep `lalrpop-util` >= `lalrpop`?  Despite my lack of clarity on why it's necessary, I'm inclined to have the directions say to publish `lalrpop-util` first, on the basis of [Chesterton's Fence](https://fs.blog/chestertons-fence/)

`cargo publish` these days automatically wait until the published package is fully accessible. Waiting 25 seconds would be a historical workaround. We can safely remove it now. Before cargo have the feature, the next release occasionally failed because its dependency is not fully published but the dependent required it.

Though lalrpop only have 2 crates, publishing multi-crate projects still have common concerns. 
1. Are the all crates have version consistency? (is this script checking it?)
2. Are they guaranteed to be published by dry-run? So, not to start publishing one crate but stuck by another one is broken. passing `--dry-run` will be helpful if directly using `cargo publish`. If cargo-release supports workspace validation, it will be even better.

Pat-Lafon:
I've successfully used `cargo-release` to publish v0.21.0. I have added a list of steps/commands for future reference. I ended up not using `cargo-release`'s workspace support only because I didn't delve too deep into learning how to ignore the members of the workspace(we don't want to publish lalrpop-test and doc/*). Two commands is sufficient given that you only try to create the tag in one of them.

> 
    Are the all crates have version consistency? (is this script checking it?)
    Are they guaranteed to be published by dry-run? So, not to start publishing one crate but stuck by another one is broken. passing --dry-run will be helpful if directly using cargo publish. If cargo-release supports workspace validation, it will be even better.


There are some version/consistency checks which can all be verified by doing a dry-run(by not including the `--execute` flag. We could also sort out the workspace feature if we want the additional automation.

> Do we still need the docs version update script?

This script also updates the lalrpop versions listed in the book which is useful. Not sure that it can be replaced just yet.

CONTRIBUTING.md
```diff
@@ -58,5 +58,17 @@ So this approach has been abandoned a while ago.
```
dburgener, line 58:
```suggestion

#### Without Cargo-release

```

dburgener (approved):

Issue #802, opened 2023-05-28
Custom lexer needs context information from the parser?


mccolljr:
This might be related to #195?

I want to parse a grammar where the same sequence of characters can mean different things in different contexts. The custom lexer I am using can handle this, but I can't tell it which mode to use from the grammar

#### Example:

- A (custom) lexer understands the individual token `">"`.
- Inside an expression context, `">" ">"` should parse as the right-shift operator, but `">" SPACE+ ">"` should parse as the `>` operator followed by an error indicating that `">"` is not a valid start token for the right-side expression.
- Outside of an expression context, `">" ">"` and `">" SPACE+ ">"` should both parse to two consecutive closing `>` braces.

The lexer can't decide between emitting two `">"` tokens or a single `">>"` token without some hint from the parser.

I can see two possible solutions:
1. Write my grammar in such a way that whitespace is a named token and explicitly encoded in the rules
2. Inform the lexer that it should NOT generate `">>"` right-shift tokens outside of an expression context

I'm currently using the option 1, but the downside is that I have to manually place whitespace productions all over the grammar, then manually filter `SPACE`, `SPACE_OR_NEWLINE`, etc out of error messages.

It would be nice to have the ability to do something like option 2. Ideally, I would have the ability to modify the lexer state when entering/exiting certain rules so I could push and pop `LexMode` values from a stack, informing the lexer which tokens it may legally generate in that context. I could definitely be wrong, but it doesn't seem like there's a way to accomplish this via lalrpop right now.

Is there a way to accomplish this, or something like it, with the current version lalrpop?

youknowone:
I had similar problem with soft keywords processing.

mccolljr:
As I continu to work on this grammar, I am realizing that the "explicit whitespace" approach fails. A grammar that is LR(1) when whitespace is implicit can become non-LR(1) when whitespace is explicit. An example of this:

This grammar is LR(1) (whitespace does not generate tokens)
```
ListOfBounds: ... = { ... }

NameWithBounds: ... = {
 <l:@L> <name:Name> <r:@R> => ...,
 <l:@L> <name:Name> ":" <bounds:ListOfBounds> => ...,
}

ListOfNamesWithBounds: Vec<ast::NameWithBounds> = {
  "(" <mut list:(<NameWithBounds> ",")> <NameWithBounds> ")" => ...
}
```

This grammar is not LR(1) (whitespace generates tokens)
```
ListOfBounds: ... = { ... }

[...]

Issue #706, opened 2023-02-17
Stateful Lexing Support


solomon-b:
The custom lexer page of the Lalrpop tutorial mentions:

> Tokens that require tracking internal lexer state

But does not give any say if Lalrpop supports stateful lexers. 

I'm coming from Haskell's Alex lexer which has a feature called "start codes" for this. You can annotate lexer rules with start codes so that they only apply when the lexer's state matches that start code. You can then push and pop start codes from a stack to set the state of the lexer.

This is a really helpful in certain situations such as lexing string templates. Does Lalrpop supports anything like this?

Marwes:
See #195, LALRPOP does not support this right now. I prototyped a feature for this in https://github.com/lalrpop/lalrpop/pull/673 but I have no plans to finish that anymore.

PR #916, opened 2024-06-26
Add an example showing how the parser can control the lexer's mode.

nwalfield:
In some languages, it is necessary for the parser to control the lexer's mode.  This is possible, but non-obvious.

This commit adds a simple example that shows how to parse the "list of length-value" language in which values are encoded as a length, followed by a colon, and then the literal number of bytes, e.g., `2:hi`.  To parse this, the parser need to tell the lexer to return the next `n` literal bytes when it sees the length prefix.

See #802.

### DIFF
--- a/doc/lexer-modes/Cargo.toml
+++ b/doc/lexer-modes/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "lexer-modes"
+version = "0.1.0"
+edition = "2021"
+authors = ["Neal H. Walfield <neal@sequoia-pgp.org>"]
+
+[build-dependencies]
+lalrpop = { version = "0.20.2", path = "../../lalrpop" }
+
+[dependencies]
+lalrpop-util = { version = "0.20.2", path = "../../lalrpop-util" }
--- a/doc/lexer-modes/build.rs
+++ b/doc/lexer-modes/build.rs
@@ -1,0 +1,3 @@
+fn main() {
+    lalrpop::process_root().unwrap();
+}
--- a/doc/lexer-modes/src/grammar.lalrpop
+++ b/doc/lexer-modes/src/grammar.lalrpop
@@ -62,7 +62,7 @@ LiteralCount: () = {
 Decimal: usize = {
     N_0 => 0,
     <x:LeadingDecimalDigit> <y:DecimalDigit*> =>? {
-        let count = std::iter::once(x).chain(y.into_iter())
+        let count = std::iter::once(x).chain(y)
             .map(|t| t.as_bytes()[0] as char)
             .collect::<String>();
         let count = count.parse::<usize>()
--- a/doc/lexer-modes/src/grammar.lalrpop
+++ b/doc/lexer-modes/src/grammar.lalrpop
@@ -1,0 +1,116 @@
+// -*- mode: Rust; -*-
+//
+// This grammar parses a list of `length:value` pairs.
+//
+// In order to parse the `value` part, we have to know the length.
+// The value part can't be parsed by the grammar, but it can be lexed
+// by the lexer.  The complication is that only the grammar can detect
+// that the lexer need to change modes.  So we need some way for the
+// parser to communicate that to the lexer.  This is possible, albeit
+// a bit awkward, using some shared state.
+
+//  When we observe the length, the parser changes the lexer to
+// "literal parsing" mode.  When next called, the lexer returns a
+// token containing the next `n` bytes, and then returns to normal
+// lexing mode.
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use crate::Value;
+use crate::lexer;
+use crate::lexer::LexerMode;
+use crate::lexer::LexicalError;
+
+// Because `mode` has to be `Copy`, we have to pass a reference to the
+// `Rc<RefCell<LexerMode>>`.
+grammar<'input, 'mode>(mode: &'mode Rc<RefCell<LexerMode>>);
+
+// Zero or more whitespace separated values.
+pub List: Vec<Value> = {
+    Whitespace* <ValueWhitespace*>
+};
+
+Whitespace = {
+    SPACE, HTAB, VTAB, CR, LF, FORMFEED,
+};
+
+ValueWhitespace = {
+    <Value> Whitespace*
+}
+
+// A value is of the form: `n:bytes` where `n` is a count and bytes is
+// exactly `n` literal bytes.
+Value: Value = {
+    LiteralCount COLON <literal:LITERAL> => {
+        Value {
+            value: literal.as_bytes().to_vec()
+        }
+    }
+}
+
+// We factor this production out of the Literal production so that we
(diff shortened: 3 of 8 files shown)


nwalfield:
quick format done.  I also updated the commit message to reference #706.

doc/lexer-modes/src/grammar.lalrpop
```diff
@@ -0,0 +1,116 @@
...
+// the literal data.
+LiteralCount: () = {
+    <count:Decimal> => {
+        // Change the lexer to literal parsing.
+        mode.borrow_mut().literal = Some(count);
+    }
+}
+
+// A Number.
+Decimal: usize = {
```
dburgener, line 62: I don't know that it really matters too much for the purpose of documentation, but is there a reason to lex the counts by digit and then reconstruct numbers at the parser level rather than when the lexer isn't in literal mode just taking all the digits up to the colon as a single token?

nwalfield: This example is a distilled version of a [sexp parser](https://datatracker.ietf.org/doc/html/draft-rivest-sexp-09), [which I wrote](https://gitlab.com/sequoia-pgp/sequoia/-/commit/464d289daf0888cd4eeaa02a23ec1cbbecbd5686).  As far as I can tell, when parsing sexp expressions, the lexer doesn't have enough information to decide how to parse the digits.  In this example, it does, and if you want to change it, that's fine by me.

doc/lexer-modes/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "lexer-modes"
+version = "0.1.0"
+edition = "2021"
+authors = ["Neal H. Walfield <neal@sequoia-pgp.org>"]
+
+[build-dependencies]
+lalrpop = { version = "0.20.2", path = "../../lalrpop" }
```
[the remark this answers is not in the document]
nwalfield: The other examples have the same issue:

```
$ grep 'lalrpop =' $(find -name Cargo.toml)
./calculator/Cargo.toml:lalrpop = { version = "0.20.2", path = "../../lalrpop" }
./pascal/lalrpop/Cargo.toml:lalrpop = { version = "0.20.2", path = "../../../lalrpop" }
./nobol/Cargo.toml:lalrpop = { version = "0.20.2", path = "../../lalrpop" }
./lexer-modes/Cargo.toml:lalrpop = { version = "0.20.2", path = "../../lalrpop" }
./lexer/Cargo.toml:lalrpop = { version = "0.20.2", path = "../../lalrpop" }
```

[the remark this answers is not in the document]
dburgener: My point is that [an open PR](https://github.com/lalrpop/lalrpop/pull/903/files) bumps these numbers for all the other examples.  Whichever of this PR or #903 merges second will need to update to sync up, since #903 obviously doesn't currently bump them for this.

doc/lexer-modes/src/grammar.lalrpop
dburgener: Okay, I'm fine with leaving it as is.  We can always update it post-merge if desired.

dburgener:
The rust code is hitting some clippy warnings, so you'll need to clean those up before merge.

doc/lexer-modes/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    lalrpop::process_root().unwrap();
```
Pat-Lafon, line 2: Pinging #911

Pat-Lafon (approved):
This looks good to me. Since it's a documentation change, I think the benefit would be to merge this.

doc/lexer-modes/src/lexer.rs
```diff
@@ -0,0 +1,205 @@
...
+pub type Spanned<Token, Loc, LexicalError> = Result<(Loc, Token, Loc), LexicalError>;
+
+// The type of the parser's input.
+//
+// The parser iterators over tuples consisting of the token's starting
+// position, the token itself, and the token's ending position.
+pub(crate) type LexerItem<Token, Loc, LexicalError> = Spanned<Token, Loc, LexicalError>;
+
+#[derive(Debug, Clone, PartialEq)]
+#[allow(non_camel_case_types)]
```
yannham, line 48: Nitpick: beside history, I don't see a good reason why token should be spelled in upper case.

yannham (approved):

doc/lexer-modes/src/lexer.rs
nwalfield: In Sequoia, we choose this convention to make the distinction between tokens and productions in the grammar clearer.

nwalfield: To be clear: I'm not saying that you or anyone else should adopt this convention. I'm explaining that I have a good reason to use this convention.

dburgener (approved):

doc/lexer-modes/src/lexer.rs
yannham: But you need to write a mapping in the `grammar.lalrpop`, don't you? So you could do:

```
        LPAREN => lexer::Token::LParen,
```

And keep the Rust-side camel case. Anyway, as I said, it's a nitpick - I'm fine keeping it like that, and I surely don't want to block this PR on that :slightly_smiling_face:

nwalfield: I could, but I don't 🙂.  This isn't my project. I'm happy for you to modify the code to suit your tastes.

yannham: Sure, it's just that

> In Sequoia, we choose this convention to make the distinction between tokens and productions in the grammar clearer.

My point was that you can still keep the distinction between tokens and productions in the grammar clear (as in the example above), while using camel case for the Rust enum. As often, I suspect the actual answer is mostly that it's a matter of taste, and that's ok :slightly_smiling_face: 

Honestly this is quite unimportant, so let's merge like this. Thanks again for the contribution!

Pat-Lafon:
Before I go wrongly closing a bunch of issues, I'll just list them here in case I'm mistaken(or if anyone sees those issues since this pr might help them solve their issue) : #706, #195, #673, #802

dburgener:
My reading of #706, #195, and #673 is that they are more about implementing this as a direct feature in lalrpop's lexer.  The approach here requires a custom lexer, and then the parser can exchange state between itself and that custom lexer to control behavior.  I haven't understood #673 at any sort of real level, but my impression is that the high level approach there is to make the built-in lexer aware of what tokens would be permitted at a given parse state, and then disambiguate lexing using that information.

So I would say that those three seem to be fundamentally about achieving this *without* a custom lexer, which is still a feature that lalrpop does not have today.

#802 does seem to me to just be a question of "is this achievable at all?" and the answer provided by this PR is "yes, and here's an example of how to do it using a custom lexer".  So I'd think that #802 can be closed with a pointer to this example.

Issue #860, opened 2023-12-23
Matching order of external tokens is not preserved during generation


js2xxx:
There is a case where multiple terminal strings map to the same variant of the token but with patterns of different exhaustiveness like this:
```lalrpop
extern {
    enum Token {
        // Keywords
        "while" => Token::Ident { symbol: kw::WHILE, is_raw: false },
        ...
        // Non-keywords
        "ident" => Token::Ident { symbol: <Symbol>, is_raw: <bool> }, // Will be checked in a production rule.
        ...
    }
}
```

However, due to the current implementation, the terminal string will be reordered alphabetically alongside the target token pattern, thus preventing patterns of less exhaustiveness but greater alphabetic order like `"while"` from being matched before those that are more exhaustive but alphabetically less like `"ident"` (`'i' < 'w'` in this case).

Users could add prefixes to those terminal strings to preserve order manually, but I think changing the implementation is the optimal solution.

If the current behavior is intended and not going to be changed, an explanation of that behavior should be placed in the manual to avoid confusion.

Pat-Lafon:
Hmmm, I think this is related to https://github.com/lalrpop/lalrpop/issues/671 and #670. From reading those issues, I think the sorting behavior is removable? It's still a little unclear.

js2xxx:
Oh, I should have checked those issues.

I wonder what the purpose of the sorting behavior is in the first place. It seems unable to make the following process more efficient but makes the external token conversion somehow unpredictable. Therefore, it should be removed in my opinion.

dburgener:
Fixed in #918

Issue #671, opened 2022-06-20
Feature request: With custom lexers, allow for terminals to alias to token specializations


Destroyerrrocket:
I'm truly sorry for the slightly undescriptive title, what I mean is this:
```
extern {
  ...
  enum PreToken {
  ...
    LParen => OperatorPunctuator("("),
    RParen => OperatorPunctuator(")"),
    Comma => OperatorPunctuator(","),
    OperatorPunctuator => OperatorPunctuator(<&'static str>),
  ...
  }
}
```
You'll find that if you test this out, It *kinda* works: It is accepted by the lalrpop parser, and it compiles, but due to line lalrpop/src/normalize/lower/mod.rs:189, with its "all_terminals.sort();", the rules are scrambled up and the match generation does not work. I'm currently working with a version of lalrpop that does not sort the terminals, all tests pass, and I haven't encountered any bugs, but I'd be nice to know if it has some purpose. Also, I'd love for a specialization for the case where "OperatorPunctuator" above has a String, that would make things significantly simpler later on :)

This would be immensely useful, as being able to specialize some of the tokens in my grammars depending on the context I need would aid me a lot for the project I'm working on. In C++ there are things that, depending on the context, may or may not be a token as the preprocessing rules change situation by situation. (for instance, defined is a preprocessing token only when inside of an if preprocessing directive). This would simply make things a lot easier! :D

Destroyerrrocket:
Btw, I think that the best way to allow for easy matching of strings is if you allow for the addition of a conditional to each terminal entry.
You're already generating `"{pattern} if true => Some({index}),"` (at lalrpop/src/lr1/codegen/parse_table.rs:750), so it would be as "easy" as throwing the condition in there. :)
Then I would simply write:
  enum PreToken {
  ...
    Defined => Ident(<String>) => {<>.as_str() == "defined"},
    HasInclude => Ident(<String>) => {<>.as_str() == "__has_include"}
    Ident => Ident(<String>),
...
  }

osa1:
> I'm currently working with a version of lalrpop that does not sort the terminals, all tests pass, and I haven't encountered any bugs, but I'd be nice to know if it has some purpose.

I don't think sorting is necessary, it's probably just there to help with debugging (maybe generated code looks nicer). In my LR(1) parser generator I don't re-order token declarations and [your example works][1].

I just tried removing that `sort` line in lalrpop and all the tests pass. It should be OK to remove it.

[...]

PR #918, opened 2024-07-05
Rebase and fix of #670 Fix token_to_integer to not require Copyable

dburgener:


dburgener:
I rebased and fixed up #670, but I can't push a non-ff to destroyerrocket's remote, so I've uploaded my change here.

This PR essentially consists of two changes, but I've left them together in order to keep @Destroyerrrocket getting credit on the commits.

One change is to drop an unneeded dereference in the generated code.  This seems to me like an obviously good change to make, although I am unclear on the specific problem @Destroyerrrocket was trying to solve by making it.

The second is to drop a sort of nonterminals.  As mentioned in #671 and #860, this sort causes problems by not preserving the precedence order in `Token` match rules.  The problem here is that it's not really clear why the `sort()` call was added in the first place, or whether anything elsewhere in the code assumes that the list of non-terminals is sorted.  I've done some code inspection and haven't found anywhere that relies on that list being sorted, but it's far from exhaustive, and I don't yet feel confident in this change.  It's worth noting that a few users in the linked bugs have ran on local sort-removing branches for some time without issues, and removing the sort passes our test suite.

So, I'm putting this up as a draft for now, mostly to have an up to date #670 recorded.  I'm still hoping to gain more confidence that removing the sort isn't going to cause any unforeseen problems.

Destroyerrrocket:
Hi! Thank you for taking care of this, I've been a bit occupied since then! I do not recall why I added a dereference there, and it most probably is fine to remove

Pat-Lafon:
Thanks for keeping this alive. Looking at the changes in the generated grammar, it is hard to see how removing the sort would cause issues. Maybe it helps in deterministically generating the same file each time?

dburgener:
> Maybe it helps in deterministically generating the same file each time?

Interesting theory.  I hadn't thought about that angle.  I don't think it should matter for determinism though, since without the sort, we just reproduce the order in the input file.

I'm more concerned about something subtly relying on sorted behavior, like a search algorithm that assumes a sorted input for example.  The list of states in the parser for example does need to be sorted because `iter_tools::chunk_by()` only groups consecutive elements (https://github.com/lalrpop/lalrpop/blob/786e2fe2650185182702d2caa40c69bcff5fd97c/lalrpop/src/lr1/codegen/parse_table.rs#L604).  I don't think the ordering of the nonterminals can influence the order of the states at all, but I'm not 100% sure of that.

The problem as I see it is that while some failure cases where dropping this `sort()` breaks things would be obvious, there's plenty of opportunities for dropping a sort to reduce performance, or fail only in narrow corner cases that we may not test for (for example removing a sort could theoretically break duplicate detection somehow, but maybe our duplicate detection tests happen to have sorted input anyways, or maybe it only breaks duplicate detection in some narrower corner case that we don't happen to test).

Here's what I've done so far:

* I'm pretty sure we can limit the scope of concern to the `lr1` folder.  The sort happens very near the end of normalization, and I've checked the rest of normalization pretty thoroughly.  So if something breaks, it's in the parser generation.
* I've mostly tracked the nonterminals Vec around inside parser generation through manual code inspection, and haven't found anything that looks like it relies on sorting directly.

However, the nonterminals Vec gets copied into other things like `state.items`, hence the reordering of the state item lists in the generated code in this PR.  I haven't traced everything down on `state.items` yet.  I also think there was one other place the vec gets copied off into another binding that I didn't follow, but I didn't take notes, so I don't recall the details, and will probably have to redo the work in the second bullet above to make sure :).

Where I am today is that if I had to guess, I'd guess that removing the `sort()` is fine, but I don't have the sort of confidence yet that I need to feel okay signing off on this change.  I'm not sure whether more code inspection by itself will get me there.  The next steps I have in mind are:

1. Looking more closely at the `state.items` angle mentioned above
2. Going through all the docs on functions we use in the code in `Vec`, `Iterator` and `itertools`, and verifying that anything documented to rely on a sorted list doesn't touch nonterminals.
3. Looking into if there's test augmentation that should be done (eg what's our coverage on duplicate detection today?  Would we expect it to catch the sort of issues this PR might cause if we're missing something?).  I kind of want to add some test code that compares various permutations of nonterminals across some invariants, but I'm not sure what invariants to check.

So I hope to poke at this some more soon, but I'm not 100% settled on what the path to success here looks like.  Open to any suggestions or feedback.

dburgener:
Here's my argument for why this is safe to merge:

* As I noted above, the all_terminals sort is very near the end of normalization.  I've inspected the rest of normalization and nothing there seems to touch the list of terminals
* That leaves the main building a state machine part of the code as the scope of where this may matter
* Thinking about the lane table algorithm at a high level, we build LR(0) states, which depends on the lookaheads, not the overall terminal list.  Then we attempt lalr, and finally build a lane table to resolve conflicts, both of which depend on the lookaheads for each state, rather than the overall list of terminals.  In other words, there's no occasion to reference the full list of terminals between normalization and output.  I verified this by checking the main entry points for each stage of the algorithm.  In other words, I didn't dig into the nitty gritty of lane table construction to verify that the list of terminals isn't used there, but I did look at the high level entry points (most of lalrpop/src/lr1/lane_table/construct/mod.rs) and verify that the code matches the algorithm in terms of constructing states based on previous states and lookaheads
* I also did a fair amount of grepping around on the string "terminal" and found nothing suspicious in terms of relying on ordering.
* Finally our test suite does pass without sort, and In #671, @Destroyerrrocket reports having used a version without sort() on some of his own workloads, so we have some decent testing confidence.

In my opinion this is good to go.  I can't mark it approved, because I technically created this PR, but you can consider this to have my "approved" on it.  If no one objects, I'll go ahead and bypass merge this (since it's morally approved, but not technically marked as approved) at some point (but I'll leave it open for a while longer in case other maintainers want to chime in).

Pat-Lafon (approved):

PR #920, opened 2024-07-09
Don't call exit() in a library, instead return an error to the user

dburgener:
This lets users who want to proceed even on lalrpop failure do so.  The common case of calling .unwrap() on lalrpop's return will continue to exit at (roughly the same place), seeing a panic rather than an exit, which is probably more in line with expected behavior.

The approach in this commit is the least invasive way to achieve this goal, since the existing returns are io::Result, this commit just returns an io::Error (mostly ErrorKind::InvalidData) on various lalrpop failures.  An alternative approach would be to define a lalrpop error type and return that.  The existing lalrpop_util::ParseError covers some but not all of the relevant cases, since it only covers errors at the parse stage, not later stages of compilation.  I'm not sure the effort to fully propagate all the errors up is worth the benefit, since the error messages are displayed anyways, and I don't know what the use case for actually parsing specific lalrpop compilation errors programmatically from build.rs is.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -23,7 +23,6 @@ use std::ffi::OsStr;
 use std::fs;
 use std::io::{self, BufRead, IsTerminal, Read, Write};
 use std::path::{Path, PathBuf};
-use std::process::exit;
 use std::rc::Rc;
 
 mod action;
@@ -277,6 +276,7 @@ fn parse_and_normalize_grammar(session: &Session, file_text: &FileText) -> io::R
                 pt::Span(location, location),
                 &format!("invalid character `{}`", ch),
             );
+            return Err(io::Error::from(io::ErrorKind::InvalidData));
         }
 
         Err(ParseError::UnrecognizedEof { location, .. }) => {
@@ -285,6 +285,7 @@ fn parse_and_normalize_grammar(session: &Session, file_text: &FileText) -> io::R
                 pt::Span(location, location),
                 "unexpected end of file",
             );
+            return Err(io::Error::from(io::ErrorKind::UnexpectedEof));
         }
 
         Err(ParseError::UnrecognizedToken {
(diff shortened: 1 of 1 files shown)


Pat-Lafon:
This seems reasonable to me. Usually the benefit of coming up with a custom error enum is to help communicate to the user what went wrong(besides what you stated). Our writing to standard error seems to report the equivalent information.

I think it would be worth using a custom error enum only if we were interested in better capturing/describing the error modes of lalrpop.

The alternative I think is to use some `dyn Error` if we want to be both opaque enough for changes but still provide something the user can handle.

A random idea, should we just forbid `exit` in the repository? I think the main other usage is https://github.com/lalrpop/lalrpop/blob/786e2fe2650185182702d2caa40c69bcff5fd97c/lalrpop/src/main.rs#L154 There seems to be a rare clippy lint that can automate this detection for us https://rust-lang.github.io/rust-clippy/master/index.html#/exit

dburgener:
> A random idea, should we just forbid exit in the repository?

The `exit()` call you link does seem a little unnecessary.  I do think `exit()` does occasionally have its place, so if we want to use that lint, I'd say that we should be very open to annotating exceptions.  I find `exit()` in libraries to be much more problematic than in binaries, which is where the one I left alone is.  In a library, `exit()` prevents the caller from deciding whether to handle the error and continue or not.  In a binary, we're exiting anyways.

That said, yes, in that situation you linked, we could just return an error instead, there's no real benefit to the `exit()` call.  I don't feel strongly one way or the other on that one.

I think on the clippy lint, my tentative opinion is that it's not necessary - we can catch such things easily enough in code review.

yannham (approved):
Sounds reasonable to me as well. We can always provide finer errors later if needed, this PR is a good step forward in any case.

As per the lint, I think it's better to enable it if we think avoiding exits is the right thing to do. Relying on code review for catching errors is fragile (and I think that's the whole point of safe languages like Rust), and here enabling the lint seems to really cost nothing (no risk of false positives or anything making the maintainers' life harder for example).

We can still annotate exceptions, but at least this forces committers to use `exit` willingly, and not by accident.

Pat-Lafon (approved):
Lets merge this and we can move the lint discussion to a different pr.

PR #922, opened 2024-07-10
Bump bit-set to 0.6

dburgener:
The main change in bit-set 0.6 is bumping their bit-vec depenendency to 0.7.  bit-vec has a number of performance improvements and bug fixes in
  the new release.

While we're looking at versions, update Cargo.lock to grab the latest versions of everything.


Pat-Lafon (approved):
LGTM

PR #923, opened 2024-07-18
Lint on calls to process::exit()

dburgener:
Add an exception for main.rs.  exit() calls are unexpected behavior in a library, where the caller should have control on whether errors are fatal or not.  In binaries, there is no caller and the arguments against exit() are weaker.  In many cases, exit() may still be a control-flow anti-pattern, but exceptions exist, so allow this for everything that's definitely only in a binary context.

Regardless of the allow() in main.rs, the clippy lint contains an exception for exit calls in entrypoint functions, so the existing exit() call, which is in fn main() does not trigger this lint, even without the allow() attribute.

### DIFF
--- a/lalrpop/src/lib.rs
+++ b/lalrpop/src/lib.rs
@@ -8,6 +8,7 @@
 // ε shows up in lalrpop/src/lr1/example/test.rs
 #![cfg_attr(test, allow(dead_code, mixed_script_confusables))]
 #![warn(rust_2018_idioms)]
+#![deny(clippy::exit)]
 
 // hoist the modules that define macros up earlier
 #[macro_use]


dburgener:
To recapture my opinion from #920 here, I think that exit() calls in libraries are always wrong (in other languages, there are some exceptions, such as allocation failure, but as far as I'm aware in rust, those exceptions should all be panic()s, not exit()s).  In a binary, the situation is more complex, since it doesn't create unexpected user behavior, but may be undesirable from a control flow standpoint.

This PR reflects that opinion, by excepting main.rs from the lint.  If a future contributor wants to add an exit() in the specifically binary context, that seems to me like a PR/style discussion in code review to determine which code flow is more maintainable in that specific context.  Linting on that situation, in my opinion, decides the question prematurely.

Pat-Lafon (approved):
This seems like a fair tradeoff to me. I'm a little surprised this is added to the workflow instead of something like `#[deny(clippy::exit)]` in the lib file. Otherwise one has to remember to add this flag when checking locally(Or just get reminded by the failing workflow I guess).

dburgener:
> I'm a little surprised this is added to the workflow instead of something like #[deny(clippy::exit)] in the lib file

Oh, you're right.  I always get confused about the "custom inner attributes are unstable" message.  I guess this isn't a *custom* inner attribute, so it's allowed here (and clearly I did the equivalent in main.rs, so I don't really know what I was thinking).  I updated as you suggest.

dburgener:
Okay, I dropped the main.rs allow, leaving just the lib.rs deny, which should apply to everything built into the library, but not things that are just built into the binary.  I tried adding an exit() in both situations and the lint reported only the one in the library, but not the one in the binary.

Issue #728, opened 2023-03-25
potentially incorrect conflict; suggest inline?


nikomatsakis:
[This grammar](https://gist.github.com/nikomatsakis/d93463b2430abe65264a7578ab86d810) that I was playing with gets a [*ton* of errors](https://gist.github.com/nikomatsakis/28368cc95aa55e071ffc01242d96ed38). They all go away if you uncomment [the `#[inline]` on `Generics`](https://gist.github.com/nikomatsakis/d93463b2430abe65264a7578ab86d810#file-grammar-lalrpop-L29); but I am not convinced that should be required for the grammar to be LR(1). We should investigate if this is a bug in the lane-table algorithm or legit; if legit, there really ought to be a way to suggest `#[inline]` as a way to resolve the problem.

Also, the error messages here are a mix of "good and bad", it'd be nice to investigate why the nice errors aren't triggering.

dburgener:
I think this conflict is correct, with inlining being a good fix.  Digging through the error messages, I found good ones like:

```
in.txt:31:5: 31:16: Local ambiguity detected

  The problem arises after having observed the following symbols in the input:
    Flags
  At that point, if the next token is a `"short"`, then the parser can proceed
  in two different ways.

  First, the parser could execute the production at in.txt:31:5: 31:16, which
  would consume the top 0 token(s) from the stack and produce a `Generics`. This
  might then yield a parse tree like
    Flags ╷          ╷ Type Id "(" Comma<Type> ")" Throws ";" Descriptor
    │     └─Generics─┘                                                 │
    └─Method───────────────────────────────────────────────────────────┘

  Alternatively, the parser could shift the `"short"` token and later use it to
  construct a `ScalarType`. This might then yield a parse tree like
    Flags "short"      ╷ Id ";"
    │     ├─ScalarType─┤      │
    │     └─Type───────┘      │
    └─Field───────────────────┘

  See the LALRPOP manual for advice on making your grammar LR(1).
```

That points at the problem being that Method and Field are ambiguous.  After a Flags, if we see a Type, we don't know whether or not to insert a `()` Generics in the stack, since we won't resolve everything else until several tokens later.  Inlining addresses this by allowing the parser to put off the decision and track each inlined rule as separate options.

This is essentially exactly the situation described in [this test](https://github.com/lalrpop/lalrpop/blob/2678f897a6860c0be48cb465ac2f4f9ffd1f1e4c/lalrpop/src/normalize/inline/test.rs#L17).

[...]

Issue #23, opened 2015-09-22
Better error messages for shift-reduce and reduce-reduce failures


nikomatsakis:
Currently S/R and R/R failures give you some pretty opaque error messages that basically just dump the state of the LR-table construction process. We can do better. [Menhir], for example, tries to give more "grammar-centric" errors, and I'd like to do the same. Another option is also to identify common scenarios and try to target those in particular, as well.

Some examples of scenarios worth targeting:

- [ ] Cases where inlining could help
- [ ] Associativity conflicts like #22

[Menhir]: http://gallium.inria.fr/~fpottier/menhir/

pierreganty:
You also might want to have a look at this:
http://conf.researchr.org/event/pldi2015/pldi2015-papers-finding-counterexamples-from-parsing-conflicts

nikomatsakis:
@pierreganty ah, definitely! At least based on the abstract, this looks like precisely what I had in mind.

nikomatsakis:
At least, the way I wanted to present the error was something like [the two parse trees I wrote here](https://github.com/nikomatsakis/lalrpop/issues/22#issuecomment-142235297), but automatically generated.

nikomatsakis:
Here is another example. The following grammar fails to compile with a shift-reduce error. It works if `#[inline]` is added to `Path`, though:

```
grammar;

pub ImportDecl: u32 = {
    "import" <Path> ";" => 1,
    "import" <Path> "." "*" ";" => 2,
};

Path: Vec<String> = {
    <head: Ident> <tail: ("." <Ident>)*> => {
        let mut v = tail;
        v.push(head);
        v
    }
};

Ident = r"a-z";
```

from an e-mail I wrote: the reason that this works is that the conflict is centered around when
to officially declare something a *path*. Without the inline, when you
have seen "import java", and then you see "." -- you have to decide
right then if this is a path (so that you can parse "import java.*;") or
not (so that you can parse "import java.lang.*;"). 

But when you inline, you effectively get:

ImportDecl =
    "import" Ident ("." Ident)* ";",
    "import" Ident ("." Ident)* "." "*" ";"

Now we don't need to decide yet whether to call this Ident a Path -- we
can just shift the `"."` and wait to see what the next token is.

ahmedcharles:
Is this still relevant, in the context of being a specific suggestion? I know the error messages have been improved a lot since the last comment on this.

There's always room for improvement but issues would ideally be specific enough to know when to close.

[...]

PR #926, opened 2024-07-27
If there are both naive and better errors available for a given

dburgener:
conflict, supress the naive ones.

When reporting an error, conflict resolution creates error states for the all the terminals in the grammar, which can be a lot of errors. Some of these will hopefully classify as the more helpful error messages, but for terminals that can't shift to a valid production, lalrpop won't be able to produce a good error message.  This can result in error messages that have many naive errors burying the more helpful ones.

Check what sorts of errors we will be displaying before displaying them. If all we got is naive errors, just display those.  If we got some good errors, use those instead and discard the naive errors.

### DIFF
--- a/lalrpop/src/lr1/error/test.rs
+++ b/lalrpop/src/lr1/error/test.rs
@@ -28,7 +28,7 @@ pub Ty: () = {
     let err = build_states(&grammar, nt("Ty")).unwrap_err();
     let mut cx = ErrorReportingCx::new(&grammar, &err.states, &err.conflicts);
     let conflicts = super::token_conflicts(&err.conflicts);
-    let conflict = &conflicts[0];
+    let conflict = &conflicts[0][0];
 
     println!("conflict={:?}", conflict);
 
@@ -70,7 +70,7 @@ pub Expr: () = {
     let err = build_states(&grammar, nt("Expr")).unwrap_err();
     let mut cx = ErrorReportingCx::new(&grammar, &err.states, &err.conflicts);
     let conflicts = super::token_conflicts(&err.conflicts);
-    let conflict = &conflicts[0];
+    let conflict = &conflicts[0][0];
 
     println!("conflict={:?}", conflict);
 
@@ -102,7 +102,7 @@ fn suggest_question_conflict() {
     let err = build_states(&grammar, nt("E")).unwrap_err();
     let mut cx = ErrorReportingCx::new(&grammar, &err.states, &err.conflicts);
     let conflicts = super::token_conflicts(&err.conflicts);
-    let conflict = &conflicts[0];
+    let conflict = &conflicts[0][0];
 
     println!("conflict={:?}", conflict);
 
@@ -146,7 +146,7 @@ Ident = r#"[a-zA-Z][a-zA-Z0-9]*"#;
     let err = build_states(&grammar, nt("ImportDecl")).unwrap_err();
     let mut cx = ErrorReportingCx::new(&grammar, &err.states, &err.conflicts);
     let conflicts = super::token_conflicts(&err.conflicts);
-    let conflict = &conflicts[0];
+    let conflict = &conflicts[0][0];
 
     println!("conflict={:?}", conflict);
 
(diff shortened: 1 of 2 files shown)


dburgener:
This doesn't touch all of the situations where we are displaying naive errors.  The comments repeatedly say that naive errors *should* never happen, but in practice they happen a lot.

Another situation is if the nonterminal with the conflict is the entry point for the parser.  The errors are looking for context around the nonterminal with the conflict, but they can't find anything that can shift into the conflicting nonterminal if it's actually the entry point for the parser.  I'm hoping to get that fixed as well, but I'm not yet sure what the best way to address it is.  I might write up the problem and file a bug to discuss there and see if others have thoughts.

Neither the problem in this PR, nor the problem in the above paragraph seem to address #728 (I tested this PR against it, and it doesn't change the output.  Suggesting that the above paragraph is different is just speculation on my part.)  I suspect there are more issues here.

dburgener:
This can likely be taken a step further.  Looking a little more at #728 that does seem to be a variation on the essential problem here.

The core problem is that when lalrpop detects a conflict, it then looks at all possible lookaheads and tries to find good examples for an error message.  In some cases, such as the test added here, one or a small number of lookaheads produce good error messages, and the rest don't.  So this PR says that if we have some good message and some bad discard the bad, which helps a lot in the case where we have a relatively small number of good messages compared to bad ones.  I think that may be the case in many grammars (in many states, a minority of all possible tokens as a lookahead can produce a valid production).  However, if the grammar either fails to find any good messages, or if there are valid productions on many lookaheads, this PR doesn't go far enough and we still end up producing many errors.

#728 is an example of such a grammar.  It looks to me as though we have some conflicts where many lookaheads make valid productions and produce good errors, as well as some conflicts where no lookaheads make valid productions and so we get 1 naive error per token in the grammar.

In general, I think we probably want to do something like what this PR does - consider all possible lookaheads, but only display the best errors.  What we really need to do though is go farther in reducing the error count.  The question is how.  With good errors, we have some things we can work with - for example we may have multiple errors for the exact same example production.  We could easily enough detect that and just pick one (for example, the first one).  I don't think that in the case of good errors, the specific example matters too much.  We could also rank the good errors more than just good/bad.  For example, is it possible that we have an ambiguity and inlining error for the same conflict?  We might just discard the ambiguity and only report the inlining which may be more useful.

If all we have is naive errors, I think things get a little more tricky.  The errors are less helpful - both to the user, as well as to lalrpop internals trying to determine what to use.  They are naive precisely because lalrpop couldn't find valid productions.  So one thing to say is that lalrpop should have found valid productions and just fix that.  For example, the situation mentioned above where the conflict is directly in a symbol that's an entry point.  If we teach lalrpop to find valid examples in that case, I'm not sure we'll see any more naive errors at all.  Even with that, I think that doing something better in the naive case than spewing a number of errors equal to the number of tokens seems worthwhile.  The question is which error to display.  Just taking the first is easy, but is there any way to say "this token would be more helpful".  Probably not - or else we would have made a good error on that token?

I know, that was a lot of text.  TL;DR - I think this PR is a good first step, but once it merges, I'd like to do follow-on work to reduce the number of errors displayed even more.

lalrpop/src/lr1/error/test.rs
```diff
@@ -180,8 +180,69 @@ VarDecl = "let";
...
+        message.emit(&mut canvas);
+        assert!(canvas
+            .to_strings()
+            .iter()
+            .map(|r| r.to_string())
+            .collect::<Vec<String>>()
+            .join("\n")
+            .contains(text));
+        assert!(calls < unique_conflicts);
+        calls += 1;
```
Pat-Lafon, line 218: Nit: I think the slightly clearer intention is to switch the ordering of these two and using a `<=`

lalrpop/src/lr1/error/test.rs
```diff
@@ -180,8 +180,69 @@ VarDecl = "let";
...
+) {
+    use crate::message::{Content, Message};
+    use ascii_canvas::AsciiCanvas;
+    let _tls = Tls::test();
+    let _lr1_tls = Lr1Tls::install(grammar.terminals.clone());
+    let err = build_states(&grammar, nt(pub_state)).unwrap_err();
+    let mut cx = ErrorReportingCx::new(&grammar, &err.states, &err.conflicts);
+    let conflicts = super::token_conflicts(&err.conflicts);
+    assert_eq!(conflicts.len(), unique_conflicts); // One group of conflicts
+    assert_eq!(conflicts[0].len(), terminal_count); // terminal count
```
Pat-Lafon, line 204: Nit: Unsure how much more extensively you plan to use this in later pr's but converting this to check terminal count with the sum of all conflicts is probably more robust.

Pat-Lafon (approved):
Thank you for the detailed write up. I had a nit or two but given what you've mentioned, I find this pr to be a rather clear/understandable first step.

Interested to see how much error reporting can be improved.

Pat-Lafon:
> Another situation is if the nonterminal with the conflict is the entry point for the parser. The errors are looking for context around the nonterminal with the conflict, but they can't find anything that can shift into the conflicting nonterminal if it's actually the entry point for the parser. I'm hoping to get that fixed as well, but I'm not yet sure what the best way to address it is. I might write up the problem and file a bug to discuss there and see if others have thoughts.

👀 I think this is worth exploring.

Pat-Lafon:
> In general, I think we probably want to do something like what this PR does - consider all possible lookaheads, but only display the best errors. What we really need to do though is go farther in reducing the error count. The question is how. With good errors, we have some things we can work with - for example we may have multiple errors for the exact same example production. We could easily enough detect that and just pick one (for example, the first one). I don't think that in the case of good errors, the specific example matters too much. We could also rank the good errors more than just good/bad. For example, is it possible that we have an ambiguity and inlining error for the same conflict? We might just discard the ambiguity and only report the inlining which may be more useful.

I imagine length of the conflict is also relevant here. If I understand correctly, shorter variations are probably going to be more helpful to users.

> If all we have is naive errors, I think things get a little more tricky. The errors are less helpful - both to the user, as well as to lalrpop internals trying to determine what to use. They are naive precisely because lalrpop couldn't find valid productions. So one thing to say is that lalrpop should have found valid productions and just fix that. For example, the situation mentioned above where the conflict is directly in a symbol that's an entry point. If we teach lalrpop to find valid examples in that case, I'm not sure we'll see any more naive errors at all. Even with that, I think that doing something better in the naive case than spewing a number of errors equal to the number of tokens seems worthwhile. The question is which error to display. Just taking the first is easy, but is there any way to say "this token would be more helpful". Probably not - or else we would have made a good error on that token?

It sounds like 'Just have LALRPOP produce only valid examples' is ideal if it is possible. I was trying to search around and see if we could take any inspiration from other parser generators. I don't think I found much that was relevant(https://www.cs.cornell.edu/andru/papers/cupex/cupex.pdf for example which didn't seem as related).

Pat-Lafon:
Is #23 related?

Issue #685, opened 2022-09-17
reduce/reduce conflicts report shows same reduction twice


osa1:
Grammar:

```rust
grammar;

extern {
    enum Token {
        "(" => Token::LParen, // 0
        ")" => Token::RParen,
        "bool" => Token::Bool(<bool>),
        "not" => Token::Not,
        "if" => Token::If,
        "then" => Token::Then, // 5
        "else" => Token::Else,
        "let" => Token::Let,
        "rec" => Token::Rec,
        "in" => Token::In,
        "-" => Token::Minus, // 10
        "-." => Token::MinusDot,
        "+" => Token::Plus,
        "+." => Token::PlusDot,
        "*." => Token::AstDot,
        "/." => Token::SlashDot, // 15
        "=" => Token::Equal,
        "<>" => Token::LessGreater,
        "<=" => Token::LessEqual,
        "<-" => Token::LessMinus,
        "<" => Token::Less, // 20
        ">=" => Token::GreaterEqual,
        ">" => Token::Greater,
        "." => Token::Dot,
        "," => Token::Comma,
        ";" => Token::Semicolon, // 25
        "_" => Token::Underscore,
        "Array.create" => Token::ArrayCreate,
        "id" => Token::Id(<String>), // 28
        "int" => Token::Int(<i64>),
        "float" => Token::Float(<f64>), // 30
    }
}

pub Expr: ParsedExpr = {
    #[precedence(level = "1")]
    "(" ")" =>
        ParsedExpr::Unit,

    "(" <expr:Expr> ")" =>
        expr,

    <b:"bool"> =>
        ParsedExpr::Bool(b),

    <i:"int"> =>
        ParsedExpr::Int(i),

    <f:"float"> =>
        ParsedExpr::Float(f),

    <var:"id"> =>
        ParsedExpr::Var(var),

    "let" <bndr:Binder> "=" <rhs:Expr> "in" <body:Expr> =>
        ParsedExpr::Let { bndr, rhs: Box::new(rhs), body: Box::new(body) },

    "let" "rec" <bndr:Binder> <args:Binders1> "=" <rhs:Expr> "in" <body:Expr> =>
        ParsedExpr::LetRec { bndr, args, rhs: Box::new(rhs), body: Box::new(body) },

    "let" <bndr:Binder> "=" <rhs:Expr> "in" <body:Expr> =>
        ParsedExpr::Let { bndr, rhs: Box::new(rhs), body: Box::new(body) },
};

[...]

PR #927, opened 2024-07-27
Add a new error message for ambiguous reductions to the same nonterminal

dburgener:
The existing error messages assume that the source of a reduce/reduce conflict is which nonterminal is reduced to.  This is a confusing result if there are two possible reductions to the same nonterminal ("I can reduce to an A, but I can also reduce to an A").  Identical reductions to the same nonterminal are easy to do and can be hard to spot in the presence of macros.

This error message detects when the conflict is for the same nonterminal and helpfully points to the specific rules in that nonterminal that conflict.

### DIFF
--- a/lalrpop/src/file_text.rs
+++ b/lalrpop/src/file_text.rs
@@ -87,14 +87,7 @@ impl FileText {
     }
 
     pub fn span_text(&self, span: pt::Span) -> &str {
-        let (start_line, start_col) = self.line_col(span.0);
-        let (end_line, end_col) = self.line_col(span.1);
-
-        if start_line == end_line {
-            &self.line_text(start_line)[start_col..end_col]
-        } else {
-            todo!()
-        }
+        &self.input_str[span.0..span.1]
     }
 
     pub fn highlight(&self, span: pt::Span, out: &mut dyn Write) -> io::Result<()> {
--- a/lalrpop/src/file_text.rs
+++ b/lalrpop/src/file_text.rs
@@ -58,7 +58,7 @@ impl FileText {
         )
     }
 
-    fn line_col(&self, pos: usize) -> (usize, usize) {
+    pub fn line_col(&self, pos: usize) -> (usize, usize) {
         let num_lines = self.newlines.len();
         let line = (0..num_lines)
             .filter(|&i| self.newlines[i] > pos)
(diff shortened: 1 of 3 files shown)


dburgener:
I set this as a draft for now, because I want to add a test case for it, but I added some test infrastructure for testing errors in #926 that I'd like to use for that, so I'll update this once #926 is merged (or pull that in here if it's not).

It's also worth noticing that this situation is very parallel to #926.  Looking at the following grammar:

```
grammar;

A: () = {
        "a" "c",
        "a" "b"? "c"
}

pub B: () = {
        "x" A "z"
}
```

In the current master branch as of the time of this writing, this outputs 4 naive errors.  This PR independently turns 1 of those naive errors into the better new error message here.  This plus #926 will silence the other three errors.

lalrpop/src/lr1/error/mod.rs
```diff
@@ -371,6 +386,36 @@ impl<'cx, 'grammar> ErrorReportingCx<'cx, 'grammar> {
...
+            .end()
+            .body()
+            .begin_lines()
+            .wrap_text(format!(
+                "The following symbols can be reduced into a {} in two ways",
+                reduce.reductions.first().unwrap().nonterminal
+            ))
+            .push(reduce.to_symbol_list(reduce.symbols.len(), styles))
+            .wrap_text("They could be reduced using the following production:")
+            .wrap_text(span1_str)
```
dburgener, line 411: TODO: I'd like to have this point at line numbers too.  Right now it just includes the text at the location.  Line numbers will be particularly helpful if the user has written the exact same thing twice.

dburgener:
closes #685

lalrpop/src/file_text.rs
```diff
@@ -86,6 +86,17 @@ impl FileText {
...
     }
 
+    pub fn span_text(&self, span: pt::Span) -> &str {
+        let (start_line, start_col) = self.line_col(span.0);
+        let (end_line, end_col) = self.line_col(span.1);
+
+        if start_line == end_line {
+            &self.line_text(start_line)[start_col..end_col]
+        } else {
+            todo!()
```
dburgener, line 96: I still need to implement this `todo!()`

PR #932, opened 2024-08-21
#[cfg] attributes overhaul

k2d222:
This PR reworked attributes in lalrpop grammar. I needed this feature so I implemented it. If you don't want to merge this, feel free to close :)

## Changes

This PR is slightly monolithic:
 * renamed `annotations` to `attributes`, to match rust naming and lalrpop's own documentation
 * changed the grammar of attributes slightly to be more flexible. it now accepts e.g. `#[ident]`, `#[a(b(c))]`, `#[a, b, c]`, `#[a = "foo", b(c = "bar"), d]`...
 * implemented `any`, `all` and `not` for `#[cfg]` attributes. So testing for *not* a feature is now possible: `#[cfg(not(feature = ""))`.
 * moved handling of `#[cfg]` to a preprocessing phase, that symply eliminates nonterminals decorated by a cfg attribute that evaluates to false. This way, one can declare two nonterminals with the same name, but guarded by different features. Also, this removes the constraint that cfg attributes must be placed on pub nonterminals.

## Example use-cases

implementing optional syntax extensions for a language:
```
#[cfg(feature = "allow-imports")]
pub TopLevel: TranslationUnit = <ImportStatement*> <Declaration*> => ...;

#[cfg(not(feature = "allow-imports"))]
pub TopLevel: TranslationUnit = <Declaration*> => ...;
```

add optional side-effects / change the action clause
```
#[cfg(feature = "debug-parser")]
Ident: String = <l: @L> <id: TokIdent> <r: @R> => { println!("parsed ident {id} at {}", l..r); id }

#[cfg(not(feature = "debug-parser"))]
Ident: String = TokIdent;
```

## Docs and Tests

I have not touched the docs nor do I have tested the code too much. I can do it if you are interested in merging.


Pat-Lafon:
This is very cool! I'm interested in getting this merged.

I've ran the pipeline so that you are aware of a the minor lints clippy reports.

Having some tests would be very helpful.

>  moved handling of #[cfg] to a preprocessing phase, that symply eliminates nonterminals decorated by a cfg attribute that evaluates to false. This way, one can declare two nonterminals with the same name, but guarded by different features. Also, this removes the constraint that cfg attributes must be placed on pub nonterminals.

This makes sense to me but I'm trying to wrap my head around the implications. Some thoughts that may be unsubstantiated:

- Does this mean users need to enable features at lalrpop build time and not when trying to compile/run with their generated grammar? Do we rebuild the grammar if the set of features changes?(See the Configuration for rebuilding on file change)
- Was there a path forward where we parsed these attributes and then generated them into the file such that Rust handles their evaluation instead of us? (Not that this is bad, having a clear subset that we support is a good place to be in)

k2d222:
Glad you're interested!

I'll make sure to clean this up then.

> Does this mean users need to enable features at lalrpop build time and not when trying to compile/run with their generated grammar? Do we rebuild the grammar if the set of features changes?(See the Configuration for rebuilding on file change)

It still uses the features field of the Configutation struct. [there](https://github.com/lalrpop/lalrpop/pull/932/files#diff-d90c67c3a3cfecc7e344fdc1652a822d2f6db3d5608ac041aecb4837da513207R40-R47). So if I get your question correctly it doesn't change from the previous implementation: a user needs to set the features at grammar build time through the Configuaration script. If they want to support different feature sets, then they have to recompile or ship different preconpiled versions.

> Was there a path forward where we parsed these attributes and then generated them into the file such that Rust handles their evaluation instead of us? (Not that this is bad, having a clear subset that we support is a good place to be in)

huh! that sounds very tempting and a 'clean' solution, but my guess is that would be too much work for little reward. Also, lalrpop has custom attributes such as precedence that would need to be handled separattel.y.

k2d222:
Cleaned up the code and commits and added a few tests.

lalrpop/src/normalize/mod.rs
```diff
@@ -51,6 +51,11 @@ fn lower_helper(session: &Session, grammar: pt::Grammar, validate: bool) -> Norm
             prevalidate::validate(&grammar)?;
         }
     );
+    let grammar = profile!(
+        session,
+        "Grammar preprocessing",
+        preprocess::preprocess(session, grammar)?
```
dburgener, line 57: I'm not sure "preprocess" is the best name for this.  Really, the whole normalize step is preprocessing.  How does "expand_cfg" sound as alternative?  I'm not tied to that name at all, just brainstorming for a more specific name.

lalrpop-test/src/lib.rs
```diff
@@ -192,6 +192,9 @@ pub fn use_cfg_created_parser() {
     #[cfg(feature = "test-set")]
     cfg::CreatedParser::new();
     cfg::AlwaysCreatedParser::new();
+    cfg::CreatedWithNotParser::new();
+    cfg::CreatedWithAnyParser::new();
+    cfg::CreatedWithAllParser::new();
 }
```
Pat-Lafon, line 198: I like this approach to testing, if Rust resolves the cfg feature annotation then we should too.

This test does look incomplete however. Note what happens when you run `cargo test --no-default-features` since this test currently only works because of the default setting.

lalrpop-test/src/cfg.lalrpop
```diff
@@ -12,3 +12,33 @@ pub Created: () = {
...
+#[cfg(all(feature = "test-set"))]
+pub CreatedWithAll: () = {
+    "x" => ()
+};
+
+#[cfg(all(feature = "test-set", feature = "test-not-set"))]
+pub NotCreatedWithAll: () = {
+    "x" => ()
+};
+
```
dburgener, line 44: It would be nice to also have some tests around the intersection of `cfg` with macro expansion.  Eg, using attributes inside a macro definition, or a conditionally defined macro.

Also for precedence interaction with `cfg`.

I'm just looking at the ordering change in normalization and wanting to make sure we have test coverage for moving `cfg` processing earlier.  Macros and precedence are the things that jump out to me as possible problem sources (although I can't think of anything specific I'd expect to go wrong there).

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -710,6 +718,55 @@ impl NonterminalData {
...
+    /// get the (key, value) of an attribute of the form #[key = "value"]
+    pub fn get_arg_equal(&self) -> Option<(&Atom, &String)> {
+        match &self.arg {
+            AttributeArg::Paren(arg) => arg
+                .first()
+                .and_then(|attr| attr.arg.get_equal().map(|eq| (&attr.id, eq))),
+            _ => None,
+        }
+    }
+}
```
Pat-Lafon, line 746: Thanks for the documentation. These two functions seem a little redundant? Do we need both or can we get away with the `get` version and just unwrap as needed?

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -710,6 +718,55 @@ impl NonterminalData {
...
+    }
+
+    /// get the content of the `Equal` enum variant
+    pub fn get_equal(&self) -> Option<&String> {
+        match self {
+            AttributeArg::Equal(eq) => Some(eq),
+            _ => None,
+        }
+    }
+}
```
Pat-Lafon, line 768: Same comment as above

doc/src/conditional-compilation.md
```diff
@@ -1,13 +1,21 @@
 # Conditional compilation
 
-LALRPOP support conditional compilation of public non-terminal declarations via
+LALRPOP support conditional compilation of non-terminal declarations via
 `#[cfg(feature = "FEATURE")]` attributes. If run in a build script LALRPOP will
 automatically pickup the features from `cargo` and use those. Alternatively an
 explicit set of features can be set using the `Configuration` type.
 
+Like rust's `cfg` attribute, the syntax accepts `not()`, `any()` and `all()`
```
Pat-Lafon, line 8: From the grammar it looks like this pr might also support nestings like `not(any(feature = "x"))`. Is this true?

lalrpop/src/normalize/preprocess/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+            match &attr.arg {
+                AttributeArg::Paren(attrs) => {
+                    attrs.iter().any(|attr| test_feat_attr(attr, session))
+                }
+                _ => false,
+            }
+        } else if attr.id == *"feature" {
+            match &attr.arg {
+                AttributeArg::Equal(feature) => session
+                    .features
```
Pat-Lafon, line 46: Not sure if this would help you but this could instead be like `match (attr.id, attr.arg)` which would maybe reduce the number of cases you have.

dburgener:
> This is very cool! I'm interested in getting this merged.

+1

> Does this mean users need to enable features at lalrpop build time and not when trying to compile/run with their generated grammar? Do we rebuild the grammar if the set of features changes?(See the Configuration for rebuilding on file change)

This is an interesting question (although as @k2d222 pointed out, this PR doesn't change that behavior).  I'm trying to think about the implications and I think I'm landing on "why does it matter?"  If we do the "pass through and let rust handle it" idea, I'm not clear on what user use cases will actually benefit?  It would only matter if the user is using a different feature set at build time vs compile time, right?  So if you're doing something like running lalrpop to generate the parser offline and then committing the `.rs` file to source control, then you'd want to handle the `cfg`s at compile time, but is that a legitimate use case?

Put another way, let's say that I'm building tool `foo`, which depends on library `bar`, which uses lalrpop.  When I build `foo` and it compiles `bar` for me, it will run `bar`'s build script to recompile the `.rs` file based on `bar`'s `.lalrpop` file, right?  Then the feature set I configured for `bar` in my `foo` `Cargo.toml` will work as expected.  Or am I missing something here?

Lastly, I dropped a couple of initial comments in the PR, but I haven't taken a thorough look at all the details.

Pat-Lafon (changes requested):
Thanks for pushing forward on testing! Given that we probably want to test all feature combinations and thus will probably want to leverage `cargo hack`, what do you think about moving cfg testing out of `larlpop` and into it's own doc example like `doc/cfg`?

I'm a little concerned that doing feature combination testing will have a lot of redundant work(and thus be slow) if run over the entire of `lalrpop-test`.

doc/src/conditional-compilation.md
Pat-Lafon: If so, lets so this off!

k2d222:
Thanks for your reviews, this week will be very busy for me so I'll address them on friday the soonest.

> If we do the "pass through and let rust handle it" idea, I'm not clear on what user use cases will actually benefit?

Need to think more about this compile vs. build time too. But my initial thought is this:

* more cfg options could be automatically supported by rust; cfg attributes support [many options](https://doc.rust-lang.org/reference/conditional-compilation.html#set-configuration-options), not just features.
* perhaps, others attributes than cfg could be supported too. think inline, clippy, derive?
* But I am afraid that could leak parser impl details and not work as expected. But maybe  some attributes could be forwarded as-is to action code?
* That means we would want to handle different attributes differently, so implementation is not straightforward.
* if the attrs are simply forwarded to the parser states, then the error reporting moves to the parser too. such errors are quite unreadable (like errors in action code btw). A parser-build-time sanitizing of attributes seems necessary.

lalrpop/src/grammar/parse_tree.rs
k2d222: yeah looking at it, they definetly look redundant and somehat hacky.
There's two of them because the unwrap one owns self. but refactoring is definitely possible.

lalrpop-test/src/cfg.lalrpop
k2d222: good catch, also should test how it interferes with inline possibly.

dburgener:
> I'm mostly wondering about the rerun directives 

Interesting question.  I've been digging in cargo docs for a while, and I'm still unclear on this behavior.  [This bug report](https://github.com/rust-lang/cargo/issues/9135) sounds suspiciously similar to our use case though (in the situation where we generate the parser in the source directory).  That makes me worry we could have problems here.

That said, I did a little poking at things and it seemed to work fine.  I flipped the lalrpop unicode feature on and off a bit in my Cargo.toml from another repo, with rerun directives in place, and changing features seemed to trigger build.rs rebuilds/reruns even in the presence of rerun directives.  All I did was pretty light smoke testing though, so I wouldn't say I have a ton of confidence.

Pat-Lafon:
@k2d222 

I've done some of the straightforward work to help this pr along.
- It's now rebased on master after the version bump
- Addressed clippy lints
- Fleshed out the cfg test to create a 1-to-1 correspondence between the feature flag on the rust side and the flag on the lalrpop side
- Move the cfg to a doc example so it can be tested separately. Otherwise our cargo hack workflow was re-testing all of lalrpop test with each possible feature combination. Saves about a minute of workflow time

The current blocker as I understand it is testing the intersection of cfg with macro expansion/precedence.

k2d222:
thanks a lot for your help @Pat-Lafon :)
I'll do a pass this weekend. Hopefully it gets merged soon!

Pat-Lafon:
I'm using this pr in my own project to get more of a hands on feel.

One thing I've noticed is that it's probably pretty free to add attributes to `Use`'s as well. Here we can probably just pass any attributes on a `use ...` statement straight through to the rust code? Not a blocker

@dburgener I sorta ran into the issue I mentioned, but with `generate_in_source_tree` instead of the rerun directives. With the following build.rs file and the corresponding changes for setting up the module in lib.rs, building will fail as you toggle between `cargo build` and `cargo build --features test-not-set` or `cargo build --no-default-features`. It's something to investigate but maybe for a separate pr since `generate_in_source_tree` is kinda weird. It could also be an issue with trying to get the build script as a whole to rerun each time.

```rust
use std::error::Error;

fn main() -> Result<(), Box<dyn Error>> {
    let mut config = lalrpop::Configuration::new();
    /*     config.emit_rerun_directives(true); */
    config.generate_in_source_tree();
    config.process_current_dir()
}
```

dburgener:
Interesting, you're sure you get that behavior with just `generate_in_source_tree()` without `emit_rerun_directives()`?  My impression from the docs was that the build script was rerun unconditionally without rerun directives.

This sounds like an issue that needs addressing, but I like the separate PR approach, since this PR doesn't introduce the behavior.

Is `generate_in_source_tree()` something you use because you want that behavior, or just to test this all out?  I ask because if that by itself breaks feature cfgs, my inclination is to wonder if it's more trouble than it's worth and deprecate it.  I've never really used that option, since I think the default of generating in OUT_DIR is just better.  But if you're using it because you want that behavior, I'm curious why, and if I'm missing something.

Pat-Lafon:
An extension for a future pr might be to allow cfg attributes in more parts of the grammar. For example, I wonder if the following would be eventually possible(since it seems much more ergonomic in some cases).

```rust
pub Created: () = {
    "x" => (),
    #[cfg(feature = "feature_y")]
    "y" => (),
};
```

Pat-Lafon:
I've tried this pr out with my own use-case and besides the ergonomic suggestions I have made, it seems to work as expected. https://github.com/Pat-Lafon/bril/commit/834dbd0aa8e120f9b41c1fc52d401149ad443d8d

k2d222:
Added a light test with macros and inline.

It would be great to support attributes on nonterminal productions!

I also needed them on the explicit terminals, which I already implemented already [here](https://github.com/k2d222/lalrpop/tree/more-cfg-attributes) and used in my project [here](https://github.com/k2d222/wgsl-tools/blob/e8eec61967e6b79bb175843739e5549b6dd27143/crates/wgsl-parse/src/wgsl.lalrpop#L106).

But I suggest we merge this first and follow-up with another PR.

dburgener (approved):
This looks good to me.  Thanks for the PR!

I'm not going to merge, since @Pat-Lafon has been more involved, so I'll let him put his approval on as well and do the final merge assuming he's okay with it.

(I did push up a quick fix for the CI failure, since the cfg parser fails to build (by design) when both test-set and test-not-set are enabled).

doc/cfg/src/cfg.lalrpop
```diff
@@ -0,0 +1,85 @@
...
+
+#[cfg(all(feature = "test-set"))]
+pub CreatedWithAll: () = {
+    "x" => ()
+};
+
+#[cfg(all(feature = "test-set", feature = "test-not-set"))]
+pub NotCreatedWithAll: () = {
+    "x" => ()
+};
```
Pat-Lafon, line 43: Given that we've added the restriction to only have test-set or test-not-set, is the following no longer a meaningful case since it will never get created? It's not clear to me that we can test the absence of this case in general.

.github/workflows/test.yaml
```diff
@@ -74,4 +74,4 @@ jobs:
       - run: cargo build --bin lalrpop --features pico-args
       - name: Run feature powerset check
         # test with minimal amount of features plus a few extra on regex/regex-syntax
-        run: cargo hack test --workspace --feature-powerset --exclude-features pico-args,default --optional-deps
+        run: cargo hack test --workspace --feature-powerset --exclude-features pico-args,default --optional-deps --mutually-exclusive-features test-set,test-not-set
```
dburgener, line 77: Cool flag that I didn't know about.  I think this means we can revert most of 8c65638baba6517897d31ecb7af3d8a6a165833b?

All the `cfg` changes in there are redundant now, but as far as I can tell `use_cfg_created_parser()` wasn't being called before I added a test case for it.

Pat-Lafon: Yeah, it's a little hidden on the README page.

I guess it depends whether we want to explicitly test for the error as opposed to just avoiding it(which is what I thought you were going for)?

dburgener: But with this change we don't explicitly test for the error anymore, right?  This causes feature-powerset to just not use these two features together, so the error check never comes up.

Pat-Lafon: Sure, lets keep that behavior.

Pat-Lafon (approved):
Thanks for bringing this pr! It is a very useful addition to LALRPOP.

doc/cfg/src/cfg.lalrpop
dburgener: One of [these](https://github.com/rust-lang/cargo/issues/1194#issuecomment-1482596022) could be an option...

Kind of hacky, but we could also just declare a type called `NotCreatedWithAllParser` at a point where it would name-collide with the generated one if it existed?

The whole thing is a little weird, since lalrpop will just fail to build the grammar with both features on, due to the `TwiceMacro` below.  Maybe what we want is a third feature that is compatible with `test-set` like `test-set2`, since otherwise we're not testing `all()` with multiple features I think.

Pat-Lafon:
> It would be great to support attributes on nonterminal productions!
> 
> I also needed them on the explicit terminals, which I already implemented already [here](https://github.com/k2d222/lalrpop/tree/more-cfg-attributes) and used in my project [here](https://github.com/k2d222/wgsl-tools/blob/e8eec61967e6b79bb175843739e5549b6dd27143/crates/wgsl-parse/src/wgsl.lalrpop#L106).
> 
> But I suggest we merge this first and follow-up with another PR.

@k2d222 as this pr has gotten merged, are you interested in opening that follow-up?

k2d222:
@Pat-Lafon I would be interested but that would have to wait a month or two unfortunately, as I'm quite busy atm.

The code for attribute on terminals is [here](https://github.com/k2d222/lalrpop/tree/more-cfg-attributes), so if you're willing to PR that yourself then go ahead. Idk how good my quick impl is.

PR #937, opened 2024-08-28
Switch to_triple to use Result::map_err()

dburgener:
It was open coding the same functionality anyways, so this is just better.

As a side effect, it dodges a new warning in rust 1.82 for unreachable code triggered by the whitespace example

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -630,14 +630,11 @@ fn emit_to_triple_trait<W: Write>(
             T = T,
             parse_error = parse_error,
         );
-        rust!(rust, "match value {{");
-        rust!(rust, "Ok(v) => Ok(v),");
         rust!(
             rust,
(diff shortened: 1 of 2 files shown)


lalrpop/src/build/mod.rs
```diff
@@ -630,14 +630,11 @@ fn emit_to_triple_trait<W: Write>(
             T = T,
             parse_error = parse_error,
         );
-        rust!(rust, "match value {{");
-        rust!(rust, "Ok(v) => Ok(v),");
         rust!(
             rust,
-            "Err(error) => Err({p}lalrpop_util::ParseError::User {{ error }}),",
+            "value.map_err(|error| {p}lalrpop_util::ParseError::User {{ error }})",
```
Pat-Lafon, line 635: While we are here, could we just call this `self` instead of `value`?

Pat-Lafon (approved):

PR #938, opened 2024-09-04
When encountering naive errors on reduce/reduce conflicts, drop all but

dburgener:
the first.

On conflicts, lalrpop tries every possible lookahead token to find good conflict examples.  On shift/reduce conflicts that makes a lot of sense, different lookahead tokens are different conflicts, because the shift is different.  On reduce/reduce conflicts on the other hand, the lookahead is sort of irrelevant.  Hopefully, at least one lookahead produces a good example, so we won't need naive errors at all.  The fact that we only found naive errors means that none of the lookaheads were helpful. So rather than reporting one error message per token, just drop all but the first, since we can't tell what lookahead is more helpful (if we'd found a helpful lookahead, we'd be using that for a better error).

### DIFF
--- a/lalrpop/src/lr1/error/mod.rs
+++ b/lalrpop/src/lr1/error/mod.rs
@@ -95,11 +95,21 @@ impl<'cx, 'grammar> ErrorReportingCx<'cx, 'grammar> {
         mut reporter: impl FnMut(Message) -> Result<(), E>,
     ) -> Result<(), E> {
         for conflict_group in &token_conflicts(self.conflicts) {
-            let (naive_conflicts, better_conflicts): (Vec<_>, Vec<_>) = conflict_group
+            let (mut naive_conflicts, better_conflicts): (Vec<_>, Vec<_>) = conflict_group
                 .iter()
                 .map(|c| (c, self.classify(c)))
                 .partition(|c| matches!(c.1, ConflictClassification::Naive));
             let conflicts = if better_conflicts.is_empty() {
+                // If we have a reduce/reduce conflict, we end up with one conflict per token, but
+                // they're all the same reduce/reduce. We don't have a meaningful way to determine
+                // if some lookahead token will be more valuable to the user, so just take the
+                // first one and drop the rest as redundant
+                if matches!(
+                    naive_conflicts.first().map(|c| &c.0.action),
+                    Some(&Action::Reduce(_))
+                ) {
+                    naive_conflicts.truncate(1);
+                }
                 naive_conflicts
             } else {
                 better_conflicts


dburgener:
On #728, this drops the number of lines of error output from 5181 to 4925, which is honestly not as much as I was hoping for.  It turns out that many of the naive conflicts produced by that grammar are shift/reduce conflicts, which this doesn't touch.  Still, this seems like a straightforward improvement without a lot of downside.  The long term goal remains to eliminate all naive conflicts, but if we have them, this is just better because the extra naive reduce/reduce conflict messages are just noise.

Pat-Lafon (approved):
The one part I struggle with when reviewing these better error message improvements is understanding how well they generalize. Something like, "Is there a built-in assumption to this that could be removed to have this apply to more cases". On the other hand, that's not a very actionable approach to improving pr's.

Let's merge this. If it or the other pr we merged don't generalize, I would hope that users will be surprised enough when they hit a not so great error message that they raise it as a bug for us to improve(as oppose to assuming it is par for the course when using parser generators).

dburgener:
> The one part I struggle with when reviewing these better error message improvements is understanding how well they generalize. Something like, "Is there a built-in assumption to this that could be removed to have this apply to more cases". On the other hand, that's not a very actionable approach to improving pr's.

I hear you, and I think that's going to be a case by case basis on specific errors.

> (as oppose to assuming it is par for the course when using parser generators).

Our docs state a goal of being "the most usable parser generator ever", so I imagine users want to hold us to that :).  We've definitely had a few issues opened in the past few months around "errors could be better", so it's happening at least some of the time.

> I was wondering if there was an ever so slightly better naive conflict to choose but it's not that obvious.
> Is there a reason to return 2 instead of 1 to be clear of what the conflict is? Or as you said, there are enough other issues in this naive case to make that obvious.

Yeah, I gave this some thought while working on the PR.  I think the answer really is "definitionally, no".  The idea behind naive errors, according to the comments in the code is that we really should have found *something* worthwhile to say already, but if we didn't, the naive error is better than nothing.  The reality is that there are plenty of cases currently where we do return naive errors.  But I think if we knew some reason to prefer one naive error to another, we should make the better naive error into a more helpful non-naive error, rather than just returning a "better" naive error.  To rephrase - if we can differentiate naive errors, we can say something useful about them, which means they shouldn't all be naive.

In the long term, I think we want to try to make good on the claim in the code that naive errors should never happen, and this code path truly becomes a fallback for lalrpop bugs rather than a semi-common case.  I'm continuing to make slow progress offline on the case where the conflict is in a start nonterminal, which is the case that I'm aware of where we generate naive errors.  But it's slow going, since it's kind of a messy problem, and I tend to get 15-30 minutes chunks of time to work on lalrpop.  I think I've understood the problem and have a way forward, but I need to finish implementing that in order to make sure no gotchas pop out.  So, my point is, there is plenty more to do here, and I'm working on it, with the ultimate goal of eliminating naive errors entirely.  But it will be a while I think, so this is good in the interim.

PR #939, opened 2024-09-04
Bump bit-set dependency to 0.8

dburgener:
There don't seem to be major user facing changes here.  The big change is that bit-set 0.8 bumps the transitive bit-vec dependency to 0.8 as well, and bit-vec seems to have some minor cleanups and performance improvements


Pat-Lafon (approved):
Should we update the lockfile too then?

dburgener:
> Should we update the lockfile too then?

Good catch.  I'm used to running `cargo update` locally and then not committing the lockfile when I'm not changing dependencies, so I missed it.

PR #952, opened 2024-09-21
Add `no_std` support for `lalrpop-util` crate with `lexer` feature

JayWhite2357:
This unlinks the `lexer` feature from the `std` feature. While expanded `no_std` support is possible in `lalrpop`, this is the key component that prevents using a the lexer within a `no_std` environment.

In order to do this, the `regex-automata` crate needs to have the `perf` feature removed, because it implies the `perf-literal-multisubstring` feature, which implies the `std` feature. To minimize the changes caused by this, the `perf` feature is re-enabled for the `std` feature of the `lalrpop-util` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,8 @@ diff = { version = "0.1.12", default-features = false }
 regex = { version = "1.3", default-features = false, features = ["std"] }
 regex-syntax = { version = "0.8.2", default-features = false }
 regex-automata = { version = "0.4", default-features = false, features = [
-        "perf",
+        "perf-inline",
+        "perf-literal-substring",
         "syntax",
         "hybrid",
 ] }
--- a/lalrpop-util/Cargo.toml
+++ b/lalrpop-util/Cargo.toml
@@ -14,9 +14,9 @@ rust-version.workspace = true
 regex-automata = { workspace = true, optional = true }
 
 [features]
-lexer = ["regex-automata/std", "std"]
+lexer = ["regex-automata"]
 unicode = ["regex-automata?/unicode"]
-std = []
+std = ["regex-automata/perf"]
 default = ["std", "unicode"]
 
 [package.metadata.docs.rs]
--- a/lalrpop-util/src/lexer.rs
+++ b/lalrpop-util/src/lexer.rs
@@ -1,4 +1,5 @@
-use std::{fmt, marker::PhantomData};
+use alloc::{fmt, vec::Vec};
+use core::marker::PhantomData;
 
 use crate::ParseError;
 
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -46,7 +46,7 @@ rand = "0.8"
 [features]
 default=["lexer", "unicode", "pico-args"]
 unicode = ["regex/unicode", "regex-syntax/unicode", "lalrpop-util/unicode"]
-lexer = ["lalrpop-util/lexer"]
+lexer = ["lalrpop-util/lexer", "lalrpop-util/std"]
 
 [package.metadata.docs.rs]
 features = ["lexer"]


dburgener (approved):

dburgener:
This seems reasonable.  Thanks for the PR!

I think in terms of the performance tradeoff losing perf-literal-multisubstring, that's just a sacrifice someone who need no-std will have to make, so there's a sense in which it doesn't matter too much how bad the performance hit is, but out of curiosity, do you have a sense of what the performance tradeoff here will be?  If not, I might do some brief perf testing on this early next week.

JayWhite2357:
> This seems reasonable. Thanks for the PR!
> 
> I think in terms of the performance tradeoff losing perf-literal-multisubstring, that's just a sacrifice someone who need no-std will have to make, so there's a sense in which it doesn't matter too much how bad the performance hit is, but out of curiosity, do you have a sense of what the performance tradeoff here will be? If not, I might do some brief perf testing on this early next week.

I have no idea. I didn't look into the internals of anything. I'm guessing it's not that much different. But, from a `no_std` perspective, slow is better than impossible.

It's also plausible that `perf-literal-multisubstring` could be made `no_std`. I didn't try, but did notice the only `std` dependencies are `memchr` and `regex-syntax`, which I believe only have `std` as an performance optimization.

In general, with 1.81 stabilizing `core::errror::Error`, `no_std` will be able expand it's reach a bit. (And this crate could benefit from `core::errror::Error`, but I didn't want to put too many changes at once.)

yannham (approved):
Double-checked the regex-automata features, this sounds about right :+1:

Pat-Lafon (approved):

dburgener:
parse-rosetta-rs using this PR with std feature on lalrpop-util:

```
Name | Overhead (release) | Build (debug) | Parse (release) | Downloads | Version
-----|--------------------|---------------|-----------------|-----------|--------
null | 0 KiB | 315ms | 2ms | - | -
chumsky | 639 KiB | 8210ms | 367ms | ![Download count](https://img.shields.io/crates/dr/ariadne) | v0.4.1
logos | 186 KiB | 5400ms | 25ms | ![Download count](https://img.shields.io/crates/dr/ariadne) | v0.4.1
grmtools | 2,483 KiB | 14411ms | 184ms | ![Download count](https://img.shields.io/crates/dr/cfgrammar) | v0.13.7
combine | 174 KiB | 4270ms | 54ms | ![Download count](https://img.shields.io/crates/dr/combine) | v3.8.1
nom | 89 KiB | 2260ms | 76ms | ![Download count](https://img.shields.io/crates/dr/nom) | v7.1.3
peg | 80 KiB | 2365ms | 29ms | ![Download count](https://img.shields.io/crates/dr/peg) | v0.8.4
pest | 121 KiB | 4571ms | 73ms | ![Download count](https://img.shields.io/crates/dr/pest) | v2.7.11
serde_json | 51 KiB | 3249ms | 17ms | ![Download count](https://img.shields.io/crates/dr/serde_json) | v1.0.121
lalrpop | 1,468 KiB | 14204ms | 41ms | ![Download count](https://img.shields.io/crates/dr/v0.21.0) | (https://github.com/JayWhite2357/lalrpop.git?rev=9e0ccd64aab70e166614b34be932cc58c407945c#9e0ccd64)
winnow | 75 KiB | 1972ms | 31ms | ![Download count](https://img.shields.io/crates/dr/winnow) | v0.6.18
yap | 55 KiB | 659ms | 37ms | ![Download count](https://img.shields.io/crates/dr/yap) | v0.12.0

*System: Linux 6.3.12-200.fc38.x86_64 (x86_64), rustc 1.81.0 (eeb90cda1 2024-09-04) w/ `-j 8`*
```

parse-rosetta-rs using this PR without std feature on lalrpop-util:

```
Name | Overhead (release) | Build (debug) | Parse (release) | Downloads | Version
-----|--------------------|---------------|-----------------|-----------|--------
null | 0 KiB | 312ms | 2ms | - | -
chumsky | 639 KiB | 8310ms | 364ms | ![Download count](https://img.shields.io/crates/dr/ariadne) | v0.4.1
logos | 186 KiB | 5277ms | 26ms | ![Download count](https://img.shields.io/crates/dr/ariadne) | v0.4.1
grmtools | 2,483 KiB | 14683ms | 194ms | ![Download count](https://img.shields.io/crates/dr/cfgrammar) | v0.13.7
combine | 174 KiB | 4164ms | 54ms | ![Download count](https://img.shields.io/crates/dr/combine) | v3.8.1
nom | 89 KiB | 2262ms | 76ms | ![Download count](https://img.shields.io/crates/dr/nom) | v7.1.3
peg | 80 KiB | 2341ms | 29ms | ![Download count](https://img.shields.io/crates/dr/peg) | v0.8.4
pest | 121 KiB | 4579ms | 76ms | ![Download count](https://img.shields.io/crates/dr/pest) | v2.7.11
serde_json | 51 KiB | 3263ms | 16ms | ![Download count](https://img.shields.io/crates/dr/serde_json) | v1.0.121
lalrpop | 1,470 KiB | 14010ms | 40ms | ![Download count](https://img.shields.io/crates/dr/v0.21.0) | (https://github.com/JayWhite2357/lalrpop.git?rev=9e0ccd64aab70e166614b34be932cc58c407945c#9e0ccd64)
winnow | 75 KiB | 1973ms | 31ms | ![Download count](https://img.shields.io/crates/dr/winnow) | v0.6.18
yap | 55 KiB | 685ms | 39ms | ![Download count](https://img.shields.io/crates/dr/yap) | v0.12.0

*System: Linux 6.3.12-200.fc38.x86_64 (x86_64), rustc 1.81.0 (eeb90cda1 2024-09-04) w/ `-j 8`*
```

In the margin of error, so seems like the performance cost to drop that optimization is negligible.

Another note, I believe that this is a breaking change requiring a semver bump: https://doc.rust-lang.org/cargo/reference/semver.html#cargo-feature-remove-another

JayWhite2357:
> Another note, I believe that this is a breaking change requiring a semver bump: https://doc.rust-lang.org/cargo/reference/semver.html#cargo-feature-remove-another

I'm not quite sure if this is the case since no functionality is removed here. The only way that this would be breaking is if someone enabled the `lexer` feature, had `default-features = false` enabled, and at the same time was implicitly relying on the additive nature of cargo features (`regex-automata/perf-literal-multisubstring` or one of its dependencies.) It feels a bit closer to this: https://doc.rust-lang.org/cargo/reference/semver.html#cargo-change-dep-feature.

I don't personally have a preference, and I know SemVer with features is hairy, so perhaps bumping is safer, anyway.

dburgener:
I think that's almost true, but it's this one bit that gives me pause:

```
- lexer = ["regex-automata/std", "std"]
+ lexer = ["regex-automata"]
```

I agree that changing which features of regex-automata we use would not be semver breaking, but this PR changes the fact that the lalrpop-util `lexer` feature used to imply the lalrpop-util `std` feature, and no longer does.  So someone relying on the implementation of `std::Error` for `lalrpop-util::ParseError` based on the `lexer` feature (which previously implied `std`) would lose that functionality.

Pat-Lafon:
Lalrpop is pre-1.0. I think our rational so far is that the cost of a breaking version bump(that is still very unlikely to break the majority of users) is acceptable. See the macro_recursion_limit change for example.

yannham:
According to SemVer, any version update (even patch) for versions with major version `0` is considered a potential breaking change: https://semver.org/#spec-item-4. So as long as we're pre-1.0, I think we can do pretty much whatever we want on paper. Of course that doesn't mean we _should_ do it, but in this case I think we're good - and I agree that @dburgener this should be a breaking change if we are to take cargo's documentation literally.

dburgener:
Sounds like we're all agreed.  I'm going to go ahead and merge this in a couple of hours unless someone tells me to wait.  We can always change our minds about breaking or not at release time too, so that shouldn't necessarily hold up the merge.

> In general, with 1.81 stabilizing core::errror::Error, no_std will be able expand it's reach a bit. (And this crate could benefit from core::errror::Error, but I didn't want to put too many changes at once.)

I agree that this change is desirable.  Are you planning on submitting a follow-up PR for it?

BTW, I spent a while poking at this in terms of breaking-ness, and it looks like if we switch to `core::error::Error` in lalrpop-util, users importing `std::error::Error` still have access to our trait impl, so I *think* that change would *not* be breaking.  I was wondering if we wanted to make sure to package them together to be nicer to our users in terms of minimizing number of manual updates.  (Uptake of 0.21 has been slow)

dburgener:
Merged.  Thanks for the PR!

JayWhite2357:
@dburgener I see that there is a manual release process. Any chance a release could get cut soon? It's the only thing currently blocking the `proof-of-sql` crate from supporting `no_std`.

dburgener:
> @dburgener I see that there is a manual release process. Any chance a release could get cut soon? It's the only thing currently blocking the proof-of-sql crate from supporting no_std.

I'll make a PR to get the conversation started.  There are a few other things I'd like to get in first in my ideal world, and we'll see what the other maintainers think.  Relevant here is the switch from `std::error::Error` to `core::error::Error`.  That seems to complete the `no_std` story better by giving `no_std` users access to error handling.  Of course the downside to that is that `core::error` need 1.81 rust, which requires bumping our msrv to the latest stable, which is more recent than we prefer.  I'm not wholly opposed to releasing without that given the msrv issue, but I'd at least like to understand our plan for that piece of `no_std` first.  So repeating my question from above, are you planning on submitting a PR for the `core::error` change?  I can put one together myself, but if you were going to do it anyways, I'll let you.

PR #953, opened 2024-09-26
Release version 0.22.0

dburgener:
For discussion.  In my opinion, I'd like to at a minimum get the following tasks done first:

- [x] Merge #927
- [x] Merge #938 
- [x] Release ascii-canvas to get a release with term 1.0 in it (https://github.com/lalrpop/ascii-canvas/pull/13)
- [x] Update lalrpop to ascii-canvas 4.0 and term 1.0
- [x] Maybe switch `std::error::Error` to `core::error::Error`?  (Issue: `core::error::Error` is rust 1.81, which is the latest stable, so this would require either waiting ~9 weeks, or violating our msrv guideline
- [x] Update our Cargo.lock to the latest
- [x] Merge #959 

Happy to add any other release blockers you all want.  I'm willing to take on all of the above, except the two PRs I mentioned need code reviews from someone else.

### DIFF
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -11,11 +11,11 @@ the following lines to your `Cargo.toml`:
 ```toml
 # The generated code depends on lalrpop-util.
 [dependencies]
-lalrpop-util = "0.21.0"
+lalrpop-util = "0.22.0"
 
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
-lalrpop = "0.21.0"
+lalrpop = "0.22.0"
 # If you are supplying your own external lexer you can disable default features so that the
 # built-in lexer feature is not included
 # lalrpop = { version = "0.21.0", default-features = false }
--- a/doc/src/quick_start_guide.md
+++ b/doc/src/quick_start_guide.md
@@ -18,7 +18,7 @@ lalrpop-util = "0.22.0"
 lalrpop = "0.22.0"
 # If you are supplying your own external lexer you can disable default features so that the
 # built-in lexer feature is not included
-# lalrpop = { version = "0.21.0", default-features = false }
+# lalrpop = { version = "0.22.0", default-features = false }
 ```
 
 Next create a [`build.rs`](https://doc.rust-lang.org/cargo/reference/build-scripts.html)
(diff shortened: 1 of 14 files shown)


JayWhite2357:
#957 only violates the msrv guideline when the `std` feature is disabled

doc/src/quick_start_guide.md
```diff
@@ -11,11 +11,11 @@ the following lines to your `Cargo.toml`:
...
-lalrpop-util = "0.21.0"
+lalrpop-util = "0.22.0"
 
 # Add a build-time dependency on the lalrpop library:
 [build-dependencies]
-lalrpop = "0.21.0"
+lalrpop = "0.22.0"
 # If you are supplying your own external lexer you can disable default features so that the
 # built-in lexer feature is not included
 # lalrpop = { version = "0.21.0", default-features = false }
```
Pat-Lafon, line 21: Maybe an easy change in the perl script but this comment always seems to be left out and its version needs to be updated manually.

dburgener: The version.sh script is a bit of a pain.  It has weird special handling for all the different variations on how the version line can be written, and tries to enumerate which files use this syntax.  That syntax is supported in all Cargo.toml files in docs, but not in the doc md files.  I'll fix the one liner here and rather than adding another exception case to the version.sh script, I'll see about just writing a single regex that handles all the formats and feed all the files to it.  There's not so much variation that that doesn't sound feasible.

dburgener: See #958.  If someone else wants to take a stab at a better approach, feel free.

dburgener:
I included the Cargo.lock update in this PR.

The only thing left on my list is #938, which still needs a review.  I don't have an issue with going ahead with the release without that.  @Pat-Lafon if you want to look over this and the release notes we can go ahead and make the release today or tomorrow.

Pat-Lafon (approved):
LGTM, we can merge that other pr on the checklist and then merge this.

dburgener:
Great, we're good to go.  I'd like to get an uninterrupted block of time to give everything a final spin, merge this and then do the release, and I've got some interruptions planned for the next few hours.  So I'll plan on merging this and generating the release either later today or tomorrow morning, based on how my day goes.

dburgener:
Published https://crates.io/crates/lalrpop/0.22.0 https://crates.io/crates/lalrpop-util/0.22.0

Thanks everyone for your contributions here!

PR #954, opened 2024-09-26
Mention the clog-cli prereq in the CONTRIBUTING release process

dburgener:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,6 +56,9 @@ So this approach has been abandoned a while ago.
 
 ### Releasing LALRPOP
 
+#### Prerequisites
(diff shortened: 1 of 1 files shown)


Pat-Lafon (approved):
I might want to remove clog-cli one day and just use a pure cargo-release workflow but until then...

dburgener:
> I might want to remove clog-cli one day and just use a pure cargo-release workflow but until then...

Honestly, even now this could really be dropped.  Since we aren't doing the commit prefixes like "chore:", "feat:" etc, it doesn't actually do much changelog writing.  All this did on my run is put an html header in RELEASES.md and I filled in the rest by hand.

PR #955, opened 2024-09-26
Fix version.sh to update lalrpop's lalrpop-util dependency

dburgener:
This could probably be lumped in the doc case, but that case is a little weird in that it assumes that any dependency with the old version number is a lalrpop dependency.  So if we happened to have some third party dependency on the same version number as the old lalrpop version in the doc examples, the script would update that too.  So rather than propagating that bug into the larger and more impactful lalrpop Cargo.toml handling, just special case it.


Pat-Lafon (approved):
I only vaguely understand perl but this seems believable to me.

PR #957, opened 2024-09-27
Support `Error` with `std` feature disabled

JayWhite2357:
This uses `core::error::Error` instead of `std::error::Error` when the `std` feature is not enabled.

As a result, the MSRV is 1.81 when the `std` feature is not enabled, but unchanged when the `std` feature is enabled.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,13 +10,16 @@ on:
 permissions:
   contents: read
 
+env:
+  MSRV : "1.70"
+
 jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         rust:
-          - "1.70"
+          - $MSRV
           - stable
           - beta
           - nightly
@@ -66,7 +69,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
         with:
-            toolchain: stable
+            toolchain: $MSRV
       - name: Install cargo-hack
         uses: taiki-e/install-action@cargo-hack
       - name: Update
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -270,6 +270,7 @@ name = "lalrpop-util"
 version = "0.21.0"
 dependencies = [
  "regex-automata",
+ "rustversion",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,3 +35,4 @@ regex-automata = { version = "0.4", default-features = false, features = [
         "syntax",
         "hybrid",
 ] }
+rustversion = { version = "1.0" }
--- a/lalrpop-util/Cargo.toml
+++ b/lalrpop-util/Cargo.toml
@@ -12,6 +12,7 @@ rust-version.workspace = true
 
 [dependencies]
 regex-automata = { workspace = true, optional = true }
+rustversion = { workspace = true }
 
 [features]
 lexer = ["regex-automata"]
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -4,6 +4,8 @@
 extern crate alloc;
 
 use alloc::{string::String, vec::Vec};
+#[cfg(not(feature = "std"))]
+use core::error::Error;
 use core::fmt;
 #[cfg(feature = "std")]
 use std::error::Error;
@@ -148,7 +150,6 @@ impl<L, T, E> From<E> for ParseError<L, T, E> {
     }
 }
 
-#[cfg(feature = "std")]
 impl<L, T, E> Error for ParseError<L, T, E>
 where
     L: fmt::Debug + fmt::Display,
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -4,7 +4,8 @@
 extern crate alloc;
 
 use alloc::{string::String, vec::Vec};
-#[rustversion::attr(since(1.81), cfg(not(feature = "std")))]
+#[rustversion::since(1.81)]
+#[cfg(not(feature = "std"))]
 use core::error::Error;
 use core::fmt;
 #[cfg(feature = "std")]
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -4,7 +4,7 @@
 extern crate alloc;
 
 use alloc::{string::String, vec::Vec};
-#[cfg(not(feature = "std"))]
+#[rustversion::attr(since(1.81), cfg(not(feature = "std")))]
 use core::error::Error;
 use core::fmt;
 #[cfg(feature = "std")]
@@ -150,6 +150,7 @@ impl<L, T, E> From<E> for ParseError<L, T, E> {
     }
 }
 
+#[cfg_attr(not(feature = "std"), rustversion::since(1.81))]
 impl<L, T, E> Error for ParseError<L, T, E>
 where
     L: fmt::Debug + fmt::Display,


JayWhite2357:
Alternatively, I could add a `core_error` feature, but I'm not sure I like that.

dburgener:
Thanks for the PR!

I like the idea of only using the recent feature when needed, but a problem with this approach is that our `rust-version` field in Cargo.toml still says 1.70, even though in `no_std` environments that is insufficient.  I'm not aware of a way to have the msrv change based on features, or what best practices in that area are.

As I understand it, the situation right now is that `lexer` works without `std`, but without `std`, `ParseError` doesn't implement the `std::error::Error` trait, which is not a showstopper, but is less than ideal, and is a surprising breakage for users.

Our options that I am aware of all have downsides:

1. Delay releasing 0.22 for about 9 weeks until 1.81 is less fresh and then use `core::error::Error` unconditionally, while bumping `rust-version` to 1.81.
2. Release 0.22 without this change, and `ParseError` doesn't implement the `Error` trait in `no_std` environments.  Bumping `rust-version` is apparently [recommended to not be treated as a breaking change](https://doc.rust-lang.org/cargo/reference/semver.html#env-new-rust) (see also https://github.com/rust-lang/api-guidelines/discussions/231), so we could later release this in 0.22.1.  That said, the second link recommends ", if possible, batch MSRV increases with semver-breaking changes.".  It certainly seems possible in this case, and user friendly to do so.
3. Migrate to core::error::Error and bump rust-version to 1.81.  1.81 is the latest stable right now.
4. Take this and leave rust-version set at 1.70 as it currently is.  Everything is "correct" and works well with the `std` feature enabled.  With `no_std`, users will receive errors trying to build with a rust version between 1.71 and 1.80, despite our claimed rust version supporting them.

I'm not wild about any of these options.  My tentative thought is that option 3 is the least bad of the set.  The rust community moves fairly quickly and generally updating toolchain is not a problem for many people.  I'm not set on that though.  I'll give it some more thought over the weekend, and I'm curious to hear others thoughts.

Pat-Lafon:
> I'm not aware of a way to have the msrv change based on features, or what best practices in that area are.

I've been trying to look around and I think these kinds of decisions are rather ad-hoc. One seems to be able to find each of these variations in other crates.

> Our options that I am aware of all have downsides:
> 
>     1. Delay releasing 0.22 for about 9 weeks until 1.81 is less fresh and then use `core::error::Error` unconditionally, while bumping `rust-version` to 1.81.

petgraph seems to be taking this route, though they also have other breaking changes they are trying to get merged all at once https://github.com/petgraph/petgraph/pull/370.

I am not sure delaying 0.22 is that worth it(unless it just happens to take that long to resolve the other deliverables).

>     2. Release 0.22 without this change, and `ParseError` doesn't implement the `Error` trait in `no_std` environments.  Bumping `rust-version` is apparently [recommended to not be treated as a breaking change](https://doc.rust-lang.org/cargo/reference/semver.html#env-new-rust) (see also [Minimum Supported Rust Version (MSRV) Policies rust-lang/api-guidelines#231](https://github.com/rust-lang/api-guidelines/discussions/231)), so we could later release this in 0.22.1.  That said, the second link recommends ", if possible, batch MSRV increases with semver-breaking changes.".  It certainly seems possible in this case, and user friendly to do so.

I have read these guidelines. I can see how it makes sense for some of the largest crates in the ecosystem but I think the pain/churn is a little lower for us compared to the chance of unexpected msrv breakage.

You have pointed out recently that not many crates have migrated to our more recent versions. It might be worth it for us to survey why. Is it a lack of awareness, the current msrv, or something else.

>     3. Migrate to core::error::Error and bump rust-version to 1.81.  1.81 is the latest stable right now.

I think I'm a little less enthusiastic about doing this with a release unless we also want to change our msrv policy. Not to chain us to it... but I think the point of having a policy is to make these kinds of decisions more consistently. 

>     4. Take this and leave rust-version set at 1.70 as it currently is.  Everything is "correct" and works well with the `std` feature enabled.  With `no_std`, users will receive errors trying to build with a rust version between 1.71 and 1.80, despite our claimed rust version supporting them.

It's not fully clear to me how dtolnay has managed this with serde/anyhow given their low msrv. There seems to be some build.rs trick mentioned here https://github.com/dtolnay/thiserror/pull/304#issuecomment-2371846932

I think it is similar to https://crates.io/crates/rustversion/ ?

Thus my additional ideas are:

5. 2 except that this gets merged in tree after 0.22 so that those who are interested can point to lalrpop's git in cargo.toml. We can optionally produce an 0.23-alpha release if there is demand. Wait to release this with other changes in an eventual 0.23.

6. Investigate whether 4 can be done with https://crates.io/crates/rustversion/ to try and get the best of both worlds. If I understand, no_std with a lower version will not get the Error impl on ParseError(but they won't expect it anyways).

I might even be fine with 4 as it is, but adding a rust version gate seems worth exploring in a limited fashion.

dburgener:
> You have pointed out recently that not many crates have migrated to our more recent versions. It might be worth it for us to survey why. Is it a lack of awareness, the current msrv, or something else.

I know you said "survey", not "speculate", but FWIW, I think part of it is just that they have to actually do a release to have our version reported on crates.io.  Even if a crate does a version update PR the day after we release and grabs our latest version, they'll still be using the old version until their next release.  If they don't ship a Cargo.lock and we release a semver compatible version, they don't need to do anything and their users get our new version instantly.  That's on top of most crates needing a human to check for new semver incompatible versions and update manually.  I'd be surprised if people are consciously rejecting our new version for some tangible reason and not also opening an issue to inform us of the blocker.

> Investigate whether 4 can be done with https://crates.io/crates/rustversion/ to try and get the best of both worlds. If I understand, no_std with a lower version will not get the Error impl on ParseError(but they won't expect it anyways).

This sounds really promising.  Great find on rustversion.  One concern that jumps to mind is that I'm not sure if rustversion can be combined with normal cfg feature flags cleanly?  I think we would want to compile the trait impl if `cfg(feature = "std")` OR `rustversion(since(1.81))`, and similarly include `core::error::Error` if `cfg(not(feature = "std"))` AND `rustversion(since(1.81))`.  The latter logic, I imagine we can just chain both attributes on top of each other, but I don't know off the top of my head if there's a way to get the OR logic when combining these attributes?

Pat-Lafon:
I modified our feature_powerset workflow to use MSRV which exposed the breakage we were concerned about. I have then given a pass at using `rustversion`. This seems to work as intended after I thought about the right way to compose features.

dburgener (approved):

dburgener:
I think this looks fine to me, and IMO we should take it for 0.23.0.  I'll wait to hit merge for a bit, in case there are other opinions to share.

I do think that at some point down the road, we should bump on up to 1.81 (or later) as our msrv anyways and then we can drop the rustversion stuff (and probably just use `core::error::Error` even with the std feature at that point.  I don't think we get any benefit from the std version except farther backwards rust compatibility).  But no rush on that.  I wouldn't be terribly surprised if it gets forced on us by a dependency bumping their msrv.  I assume this `core::error::Error` change is one that will get picked up a fair amount and ripple msrv changes through the ecosystem.  But I'd rather be late to that party personally, rather than us being the ones to force people forward.

Pat-Lafon:
> I think this looks fine to me, and IMO we should take it for 0.23.0. I'll wait to hit merge for a bit, in case there are other opinions to share.

Did you mean 0.22.0? I thought this would now be safe to add now if there is no msrv change.

> I do think that at some point down the road, we should bump on up to 1.81 (or later) as our msrv anyways and then we can drop the rustversion stuff (and probably just use `core::error::Error` even with the std feature at that point. I don't think we get any benefit from the std version except farther backwards rust compatibility). But no rush on that. I wouldn't be terribly surprised if it gets forced on us by a dependency bumping their msrv. I assume this `core::error::Error` change is one that will get picked up a fair amount and ripple msrv changes through the ecosystem. But I'd rather be late to that party personally, rather than us being the ones to force people forward.

Sounds reasonable to me... I would not want to rely on rustversion too much as it would increase the matrix of things to check for.

dburgener:
> Did you mean 0.22.0? I thought this would now be safe to add now if there is no msrv change.

Yes.  It's Monday morning and I thought about double checking that I was remembering the right number, but didn't actually do that.

> I would not want to rely on rustversion too much as it would increase the matrix of things to check for.

+1

Pat-Lafon (approved):

dburgener:
Merged, thanks!

PR #958, opened 2024-09-27
Kind of unify version.sh regex search

dburgener:
This really only removes one special casing situation, but it seems somewhat more comprehensive and covers the comment case that was missed before

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -27,16 +27,21 @@ echo "Found consistent version $VERSION"
 
 clog --setversion $1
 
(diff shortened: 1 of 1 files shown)


dburgener:
This I think is somewhat better, and it fixes the bug @Pat-Lafon pointed about about in the comment in one of the markdown files.  This whole approach is still sort of wonky though, since there are a lot of different syntaxes to specify the version.

Pat-Lafon (approved):

Issue #681, opened 2022-09-06
Feature request: Allow conditional compilation of (at least) lexer mappings via #[cfg()]


kleinesfilmroellchen:
In [spcasm](https://github.com/kleinesfilmroellchen/spcasm), which I'm currently in the process of moving from a hand-rolled parser to LALRPOP (a super pleasant process by the way!), the hand-rolled recursive descent parser has a collection of conditionally compiled code that only parses test comments in a test build. These comments check the output of the assembler against the known good machine code and thereby serve as the test suite for the assembler. All of the code pertaining to this test output, including the TestComment token, relevant AST struct and enum fields, lexer, parser and assembler support is gated behind `#[cfg(test)]` in order to speed up production code that doesn't need it.

This poses a problem for LALRPOP, as some of the data structures change depending on compilation configuration; so it's not enough to always parse test comments and discard them if necessary. I'd have to do something like this in my lexer adapter:

```rust
extern {
	enum Token {
		"identifier" => Token::Identifier(<String>, <SourceSpan>),
                /* snip */
		"=" => Token::Equals(<SourceOffset>),
		#[cfg(test)]
		";=" => Token::TestComment(<Vec<u8>>, <SourceSpan>),
	}
}
```

For the generated code, it's probably enough to include the cfg above the respective statement in `__token_to_integer`.

For me, this would be all that is necessary. I'd be fine with having the rules for test comments always exist, just never being actioned on a regular build as the respective lexer tokens don't exist and are never emitted. After all, LALRPOP internally doesn't store my custom tokens. However, I'd love to be able to #[cfg()] other parts of the grammar, like individual rules or entire non-terminals. I recognize that that's probably a lot more complicated or even impossible, as the state machine needs to depend on the configuration in order to truly "remove" the #[cfg()]'d code as normal Rust does it. The only workaround right now is to build this logic yourself via build.rs, e.g. preprocessing the .lalrpop file or choosing between two different versions.

So TL;DR I'm asking for two things:
1. (easy) Conditionally compiled lexer mappings via #[cfg()]
2. (hard) Conditionally compiled anything (rules, nonterminals, macros?) via #[cfg()]

Pat-Lafon:
It was relatively unknown to me but lalrpop has rudimentary cfg support in https://github.com/lalrpop/lalrpop/blob/master/doc/src/conditional-compilation.md . Not that this helps entirely here, but it does provide a road map for future improvements/help anyone who stumbles upon this issue.

[...]

PR #959, opened 2024-09-27
Cfg attibutes on Alternatives

Pat-Lafon:
This is an initial follow up to https://github.com/lalrpop/lalrpop/pull/932#issuecomment-2376762661 and #932.

I've pulled in the 2 un-merged commits from @k2d222 to get me started. I then added a simple test to see if this extension worked.

I found I needed to add `cfg` as a valid attribute(for this part of the grammar) and I needed to implement the filtering based on whether `cfg_active` was true. Not sure if this was because this work was unfinished or if because there were some unmerged changes. This cleaned up the build time errors though I should probably try this out in my project to see if it's actually working as expected.

I currently also need to update the documentation


Pat-Lafon:
Hmm, maybe I miss-understood what was previously implemented now that I look at https://github.com/k2d222/wgsl-tools/blob/e8eec61967e6b79bb175843739e5549b6dd27143/crates/wgsl-parse/src/wgsl.lalrpop#L11-L119 which uses Alternatives in a different part of the grammar(in the token enum). Seems like this more complete support wasn't to hard to add?

dburgener (approved):

dburgener:
This looks fine to me.  I didn't really follow your comment about mis-understanding what was previously implemented though.  Is that casting doubt on the whole approach here?

Minor notes:
* I agree that documentation will be helpful
* I'm somewhat concerned with all of this that users will get confusing error messages if they mess up the cfg overlap (they meant to specify a disjoint set of attributes, and with complicated logic didn't).  I think that raises the importance of #927.  And I (or someone) should probably look more closely at how nice the error message for declaring the same non-terminal twice.  It also shouldn't be too hard to add follow-on work to #927 to detect the use of cfg attributes and point to that as a possible problem.

Pat-Lafon:
More that the original change was to support something like #681 (atleast my interpretation of that issue). I had then extended that to support what I included in the test.

Your second point is a good point, though I'm not sure it's just limited to cfg's(though that may make the problem more likely to happen). For example, in adding this new test case, I purposely started by leaving the cfg flag off of 

```rust
 //  #[cfg(feature = "bit")] <- remove this
  #[precedence(level="4")] #[assoc(side="right")]
    <lhs:Expression> "<<" <rhs:Expression> => {
    Box::new(ast::Expression::BinaryOperation {
      lhs,
      operator: ast::Operator::Shl,
      rhs
    })
  },
```

which gives me the following error when I run with out the feature enabled.

```
Caused by:
  process didn't exit successfully: `.../lalrpop/target/debug/build/lexer-00a5c8f12de02016/build-script-build` (exit status: 101)
  --- stdout
  processing file `./src/grammar.lalrpop`

  --- stderr
  thread 'main' panicked at lalrpop/src/lr1/lookahead.rs:194:53:
  no entry found for key
```

I would contend that even a user not using cfg's could hit this

dburgener:
I agree that a lot of the problems here are not unique to cfgs, just that cfgs make them more likely to happen.  When I first saw #685 I thought it seemed pretty low priority, since in that case, the user had literally written the same alternative twice.  It was when I accidentally did it myself using macros that I realized how easy it was to do, and made #927 to try to make it clearer.  The cfg is something that makes it worse.  I think that #927 plus existing errors for duplicate nonterminals is sufficient for the duplication case (I just tested declaring the same non-terminal twice, and got a very nice error message telling me I had duplicate nonterminal names).

That said, I'm a little confused by the example you posted.  Is it really possible without cfg?  Is that just what happens if a custom lexer doesn't declare a terminal that occurs in the grammar?  Or is something deeper going on?

dburgener (approved):

Pat-Lafon:
> That said, I'm a little confused by the example you posted. Is it really possible without cfg? Is that just what happens if a custom lexer doesn't declare a terminal that occurs in the grammar? Or is something deeper going on?

I dug a little deeper and figured out that there was an additional part of the grammar that was not being checked for an active cfg. So your intuition was spot on. The grammar after the conditional compilation normalization still had one part referencing what should have been an inactive terminal. After a fix, the error message is now the same as if you had forgotten to add the "<<" case to the token enum.

```
Caused by:
  process didn't exit successfully: `.../lalrpop/target/debug/build/lexer-00a5c8f12de02016/build-script-build` (exit status: 101)
  --- stdout
  processing file `./src/grammar.lalrpop`
  ./src/grammar.lalrpop:83:20: 83:23 error: terminal `"<<"` does not have a pattern defined for it

  --- stderr
      <lhs:Expression> "<<" <rhs:Expression> => {
                       ~~~~
  thread 'main' panicked at doc/lexer/build.rs:2:28:
  called `Result::unwrap()` on an `Err` value: Kind(InvalidData)
```

Basically my mental model is that there should be some equivalence between grammars using cfg's with some active set of features, and an equivalent grammar that does not use cfg's for which lalrpop gives the same error/result modulo span information. Or in another way, if you have an error with a grammar using cfgs, deleting all of the parts of the grammar that have an inactive feature should not change anything.

dburgener (approved):

dburgener:
Sounds good!  Thanks for the extra investigation!

I do agree with your mental model in the last paragraph 100%.

Issue #960, opened 2024-09-28
Missing extern token causes weird runtime behavior


arnaudgolfouse:
Here is a simplified version of a project a friend of mine is working on:

<details> 
  <summary>Project files</summary>

`main.rs`
```rust
pub mod ast;
pub mod lexer;

use lalrpop_util::lalrpop_mod;
lalrpop_mod!(grammar);

use lexer::Lexer;

fn main() {
    let input = "1";
    let res = grammar::LiteralParser::new()
        .parse(Lexer::new(input))
        .unwrap();
    println!("{:?}", res);
}
```

`lexer.rs`
```rust
use logos::Logos;

#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Logos)]
#[logos(error = LexerError)]
#[logos(skip r"[ \t\n\f]+")] // Ignore this regex pattern between tokens
pub enum Token<'input> {
    #[token("=")]
    Assign,
    #[token("let")]
    KeywordLet,
    #[regex("[0-9]+", |lex| lex.slice().parse().ok())]
    Integer(i64),
    #[regex("[a-zA-Z]+")]
    Identifier(&'input str),
}

#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
pub struct LexerError;

pub struct Lexer<'input> {
    inner: logos::SpannedIter<'input, Token<'input>>,
}
impl<'input> Lexer<'input> {
    pub fn new(input: &'input str) -> Self {
        Self {
            inner: Token::lexer(input).spanned(),
        }
    }
}
impl<'input> Iterator for Lexer<'input> {
    type Item = Result<(usize, Token<'input>, usize), LexerError>;
    fn next(&mut self) -> Option<Self::Item> {
        let (next_token, range) = self.inner.next()?;
        Some(next_token.map(|t| (range.start, t, range.end)))
    }
}
```

`ast.rs`
```rust
use crate::lexer::Token;

#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
pub enum Ast<'input> {
    LetBinding(&'input str, Box<Self>),
    Literal(Token<'input>),
}
```

`grammar.lalrpop`
```rust
use crate::lexer::{Token, LexerError};
use crate::ast::Ast;
use Token::*;

grammar<'input>;

extern {
    type Location = usize;
    type Error = LexerError;

    enum Token<'input> {
        "=" => Assign,
        "let" => KeywordLet,
[...]

PR #961, opened 2024-09-30
Turn a few warnings back on when we match token values

dburgener:
Since these are user-supplied, this can help them diagnose bugs

### DIFF
--- a/lalrpop/src/lr1/codegen/parse_table.rs
+++ b/lalrpop/src/lr1/codegen/parse_table.rs
@@ -756,6 +756,9 @@ impl<'ascent, 'grammar, W: Write> CodeGenerator<'ascent, 'grammar, W, TableDrive
             .emit()?;
         rust!(self.out, "{{");
 
+        // This match contains user-supplied token names.  Reenable some warnings to help them
+        // catch errors if they've got a bug in their custom lexer implementation
+        rust!(self.out, "#[warn(unused_variables)]");
         rust!(self.out, "match {p}token {{", p = self.prefix);
 
         for (terminal, index) in self.grammar.terminals.all.iter().zip(0..) {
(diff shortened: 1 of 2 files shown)


Pat-Lafon:
I'm fine with this, but it feels a little clunky. We are now forcing users into never using NonSnakeCase for their tokens right? Is the bigger problem that we set our allow lints too broadly?

dburgener:
> I'm fine with this, but it feels a little clunky

I do agree.  Although the existing code that blanket allows a lot of warnings is arguably more clunky.

> We are now forcing users into never using NonSnakeCase for their tokens right?

Yes.  Although #960 hits both the non_snake_case and the unused_variables warning if they aren't suppressed, so unused_variables would be sufficient for at least that case.  Maybe just `#[warn(unused_variables)]` and keep non snake case suppressed would be a better solution.

> Is the bigger problem that we set our allow lints too broadly?

Yes.  Although at least the non-snake-case one has a lot of other hits for variables that we insert in the generated code.  The broad unused_variables isn't as bad.  But by nature of doing code generation, we do end up with unused variables.  It's possible that we could target our allow lints more specifically on that one.

I think based on your point on non-snake-case, I support at a minimum scaling this back to just the unused variables lint.  That really should be targeted pretty specifically at the situation in #960, and I can't think of a situation in which users would want unused variables there (and if they do, they can always prefix with an underscore).  If that works for you, I'll update the PR.  If you think even that is too much, then I'd be happy to table this in the hope of making the allow lints more granular (or unnecessary).  That isn't something that I personally am planning on getting to soon though.

Pat-Lafon:
> I think based on your point on non-snake-case, I support at a minimum scaling this back to just the unused variables lint. That really should be targeted pretty specifically at the situation in #960, and I can't think of a situation in which users would want unused variables there (and if they do, they can always prefix with an underscore). If that works for you, I'll update the PR.

Yup, that sounds reasonable to me. Lints in general are a constant wack-a-mole. We can always tweak this someday.

Pat-Lafon (approved):
Just linting on unused variables seems like a safe choice here.

Pat-Lafon (approved):

PR #963, opened 2024-09-30
Update to ascii-canvas 4.0 and term 1.0

dburgener:
The main change in ascii-canvas is also the term 1.0 fix.  Bumping to term 1.0 gets us several bugfixes, and a nice dependency tree reduction.


Pat-Lafon (approved):
Very nice(fun to note that we already had rustversion in our dependency tree)

PR #967, opened 2024-10-04
Elide needless lifetimes in impls

dburgener:
Tracking lifetime support in impls was added in recent nightly clippy.

https://github.com/rust-lang/rust-clippy/pull/13286


Pat-Lafon (approved):
Looks cleaner to me!

PR #968, opened 2024-10-07
Update and fix README links

dburgener:
* Add docs.rs links for our autogenerated API docs
* Fix the example project links to 1. Link to the most recent version instead of a specific commit from years ago 2. Point at the new RustPython location, since they have migrated their parser into a separate repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ to! Here are some tips:
 [advanced setup]: https://lalrpop.github.io/lalrpop/advanced_setup.html
 [cheat sheet]: https://lalrpop.github.io/lalrpop/cheatsheet.html
 [tutorial]: https://lalrpop.github.io/lalrpop/tutorial/index.html
-[LALRPOP]: https://github.com/lalrpop/lalrpop/blob/8034f3dacc4b20581bd10c5cb0b4f9faae778bb5/lalrpop/src/parser/lrgrammar.lalrpop
-[Gluon]: https://github.com/gluon-lang/gluon/blob/d7ce3e81c1fcfdf25cdd6d1abde2b6e376b4bf50/parser/src/grammar.lalrpop
-[RustPython]: https://github.com/RustPython/RustPython/blob/fb5ac9e79bfd5029397652a12883a8cedfa01620/compiler/parser/python.lalrpop
-[Solang]: https://github.com/hyperledger/solang/blob/b867c8a6c7a1ee89d405993abef458fc59e9b0fb/solang-parser/src/solidity.lalrpop
+[LALRPOP]: https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/parser/lrgrammar.lalrpop
+[Gluon]: https://github.com/gluon-lang/gluon/blob/master/parser/src/grammar.lalrpop
+[RustPython]: https://github.com/RustPython/Parser/blob/main/parser/src/python.lalrpop
+[Solang]: https://github.com/hyperledger/solang/blob/main/solang-parser/src/solidity.lalrpop
 [gitter lobby]: https://gitter.im/lalrpop/Lobby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,7 @@ to! Here are some tips:
 - Returning users of LALRPOP may benefit from the [cheat sheet].
 - The [advanced setup] chapter shows how to configure other aspects of LALRPOP's
   preprocessing.
+- docs.rs API documentation for [lalrpop](https://docs.rs/lalrpop/latest/lalrpop/) and [lalrpop-util]
```
Pat-Lafon, line 40: Is there a corresponding link for lalrpop-util as well?

Pat-Lafon: Oh never mind, it is linked all the way at the end

Pat-Lafon (approved):

README.md
dburgener: Yeah, to clarify the confusion on that, I tried linking them both at the bottom, there's already a "LALRPOP" link and I guess markdown links are case insensitive, so it was linking to the lalrpop grammar instead.

I suppose I could put the lalrpop-util link in line as well for consistency

README.md
```diff
@@ -51,11 +52,12 @@ to! Here are some tips:
 [advanced setup]: https://lalrpop.github.io/lalrpop/advanced_setup.html
 [cheat sheet]: https://lalrpop.github.io/lalrpop/cheatsheet.html
 [tutorial]: https://lalrpop.github.io/lalrpop/tutorial/index.html
-[LALRPOP]: https://github.com/lalrpop/lalrpop/blob/8034f3dacc4b20581bd10c5cb0b4f9faae778bb5/lalrpop/src/parser/lrgrammar.lalrpop
-[Gluon]: https://github.com/gluon-lang/gluon/blob/d7ce3e81c1fcfdf25cdd6d1abde2b6e376b4bf50/parser/src/grammar.lalrpop
-[RustPython]: https://github.com/RustPython/RustPython/blob/fb5ac9e79bfd5029397652a12883a8cedfa01620/compiler/parser/python.lalrpop
-[Solang]: https://github.com/hyperledger/solang/blob/b867c8a6c7a1ee89d405993abef458fc59e9b0fb/solang-parser/src/solidity.lalrpop
+[LALRPOP]: https://github.com/lalrpop/lalrpop/blob/master/lalrpop/src/parser/lrgrammar.lalrpop
```
yannham, line 55: I have mixed feeling about this change, because nothing guarantees that the link will stay valid forever (the file may move on master, or the repo could even just stop using LALRPOP altogether). So permalink sounds better. On the other hand, the precise pinned revision is a bit arbitrary and we surely don't want to bother updating them regularly.

dburgener: I do hear you on that.  That said, the Lalrpop and Gluon links were six years old at this point, which in my view is not a great spot to be in either.  (RustPython and Solang were switched from tracking master to a specific commit two years ago).

If we had been just tracking master since the last updates, RustPython is the only link that would have broken.  Both gluon and solang switched from naming their primary branch "master" to "main", but github handles that redirect automatically for us.  The only thing that broke here is RustPython moving the file.  My inclination is to say that breakage of that sort should be expected to be rare enough that it's worth keeping the links current and dealing with occasional breakage.  I can definitely see the other side though, so if there are votes for referencing specific commits, I can switch this to just referencing modern commits instead, with a hope of doing that every few years.

In the lalrpop case specifically, we control the location of lrgrammar.lalrpop, so in my opinion, even if we want to do specific commits on the others, we should point at lalrpop master for the lalrpop grammar to keep that one up to date.  In the unlikely event that we move the file in our repo, it shouldn't be too hard to remember to update the README link at the same time.

yannham: Right, I meant for the other links - the one in LALRPOP is fine as we control it. I think I agree that there's no good solution here, and it's most likely to remain valid, so let's proceed as it is. Having out of date examples isn't great as well.

PR #971, opened 2024-10-13
Support for cfg test

Pat-Lafon:
I'm interested to try and close the loop on #681 plus flesh out the cfg support. Hypothetically, there shouldn't be too much to do since `#[cfg(test)]` is already accepted by the lalrpop grammar.

Unfortunately, because of https://github.com/rust-lang/cargo/issues/4789, users probably need to go through the extra step of manually setting the `CARGO_CFG_TEST` flag since this is otherwise not currently exposed to the build script by `cargo test`. One can control all of this flag-setting in the build script, but then you need to signal to the build script when to `std::env::set_var("CARGO_CFG_TEST", "true");`.

I've included a test via the lexer cfg example which depending on whether `cfg(test)` is set, it will produce a `print` or `eprint` in the AST.

@kleinesfilmroellchen might have guidance on what I've missed.

(whoops, I meant to make a personal branch. My bad)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -49,7 +49,9 @@ jobs:
       - name: Run rustfmt check
         run: cargo fmt --all -- --check
       - name: Run clippy
-        run: cargo clippy --all --all-targets -- -D warnings
+        run: cargo clippy --all --all-targets --exclude lexer -- -D warnings
+      - name: Run clippy on lexer separately
+        run: export CARGO_CFG_TEST=true && cargo clippy --all-targets --manifest-path doc/lexer/Cargo.toml -- -D warnings
   min_version:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -77,4 +77,4 @@ jobs:
       - run: cargo build --bin lalrpop --features pico-args
       - name: Run feature powerset check
         # test with minimal amount of features plus a few extra on regex/regex-syntax
-        run: cargo hack test --workspace --feature-powerset --exclude-features pico-args,default --optional-deps
+        run: export CARGO_CFG_TEST=true && cargo hack test --workspace --feature-powerset --exclude-features pico-args,default --optional-deps
--- a/doc/lexer/build.rs
+++ b/doc/lexer/build.rs
@@ -1,3 +1,6 @@
 fn main() {
+    // https://github.com/rust-lang/cargo/issues/4789
+    // println!("cargo::rustc-cfg=test");
+    // std::env::set_var("CARGO_CFG_TEST", "true");
     lalrpop::process_src().unwrap();
 }
--- a/doc/lexer/src/ast.rs
+++ b/doc/lexer/src/ast.rs
@@ -4,9 +4,7 @@ pub enum Statement {
         name: String,
         value: Box<Expression>,
     },
-    #[cfg(not(test))]
     Print { value: Box<Expression> },
-    #[cfg(test)]
     EPrint { value: Box<Expression> },
 }
 
--- a/doc/lexer/src/ast.rs
+++ b/doc/lexer/src/ast.rs
@@ -4,8 +4,12 @@ pub enum Statement {
         name: String,
         value: Box<Expression>,
     },
-    Print { value: Box<Expression> },
-    EPrint { value: Box<Expression> },
+    Print {
+        value: Box<Expression>,
+    },
+    EPrint {
+        value: Box<Expression>,
+    },
 }
 
 #[derive(Clone, Debug, PartialEq)]
--- a/doc/lexer/src/ast.rs
+++ b/doc/lexer/src/ast.rs
@@ -4,9 +4,10 @@ pub enum Statement {
         name: String,
         value: Box<Expression>,
     },
-    Print {
-        value: Box<Expression>,
-    },
+    #[cfg(not(test))]
+    Print { value: Box<Expression> },
+    #[cfg(test)]
+    EPrint { value: Box<Expression> },
 }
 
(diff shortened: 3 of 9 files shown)


dburgener:
Just making sure I understand this correctly.

There are essentially two issues blocking #681.  One is that cargo doesn't set CFG_FEATURE_TEST for build scripts.  We can't really do anything directly about that except advise our users to set it manually, and all this PR does for that is to set it manually in our CI.  The second issue is that we were look for the `feature = "feature_name"` syntax.  We parsed `cfg(foo)` successfully, but didn't map it into our logic for including or excluding grammar items.  This PR adds support for that in the case of `cfg(test)`.

Do I have the high level idea correct?

One initial concern I have about this is that while exporting CFG_FEATURE_TEST in the CI is fine, it seems like a pretty major usability flaw that this PR breaks just typing `cargo test` from the command line.

Pat-Lafon:
> The second issue is that we were look for the `feature = "feature_name"` syntax. We parsed `cfg(foo)` successfully, but didn't map it into our logic for including or excluding grammar items. This PR adds support for that in the case of `cfg(test)`.

This is correct. If we look at https://doc.rust-lang.org/reference/conditional-compilation.html which Lalrpop supports a subset of, cfg's ast ends with either `feature = X` or `X`(or some other `Y = Z` variations which I don't think are as relevant to Lalrpop). `cfg(test)` is defined explicitly and is probably a somewhat common pattern.

>  One is that cargo doesn't set CFG_FEATURE_TEST for build scripts. We can't really do anything directly about that except advise our users to set it manually, and all this PR does for that is to set it manually in our CI. 

(Nit: it is `CARGO_CFG_TEST` which maybe denotes that test handling is special compared to other features.) This is basically true. I've tried to investigate this further and have found there to be a much larger effort/discussion around cargo "modes" and build scripts. For example https://github.com/rust-lang/cargo/issues/4001. These efforts have been around for multiple years and have largely stalled afaik. Some people have found hacks related to build outdir paths, a special cargo-doc env var, attempts at profile specific environment variables, etc. None of these work for the "test" mode.

TLDR: There is no method for determining in a build script whether you are in test or build mode(so that cargo can reuse the same artifact). There is interest, but probably little effort in changing this in the near future.

> One initial concern I have about this is that while exporting CFG_FEATURE_TEST in the CI is fine, it seems like a pretty major usability flaw that this PR breaks just typing `cargo test` from the command line.

I had hoped to find a nicer solution. If the issue is just with CI, then what we can do is set the build script to always set the `CARGO_CFG_TEST` variable and export it to the rest of the build process(see the commented out lines in `build.rs`). This would force this one crate to always be built in test mode even if you run `cargo build`(though the tests won't run just off cargo build to be clear).

There is maybe a more fundamental question of if we should merge this given that this feels a lot more like a "nightly" kind of feature. Cargo could eventually come up with a solution that doesn't involve the `CARGO_CFG_TEST` env variable being set for build scripts. Alternatively the status quo could just remain as is/cargo ends up setting the `CARGO_CFG_TEST` env var.

I think this pr has value and `cfg(test)` seems like a useful and well known/expected pattern. 

On the other hand, there are probably workarounds that are of equivalent level of effort for users, if less intuitive. For example, creating a `feature = test` feature(#681 seemed to be explicitly against this option). We could alternatively close/stall this pr as "waiting for upstream".

dburgener:
> TLDR: There is no method for determining in a build script whether you are in test or build mode(so that cargo can reuse the same artifact). There is interest, but probably little effort in changing this in the near future.

Yeah, this is a major pain point, but one that doesn't look likely to change.

> If the issue is just with CI, then what we can do is set the build script to always set the CARGO_CFG_TEST variable and export it to the rest of the build process(see the commented out lines in build.rs). This would force this one crate to always be built in test mode even if you run cargo build(though the tests won't run just off cargo build to be clear).

Would that just affect the lexer doc crate?  That might not be the end of the world.  The whole concept of testing `cfg(test)` is weird though, since you can't actually run *tests* with `cfg(test)` disabled.

> I think this pr has value and cfg(test) seems like a useful and well known/expected pattern.
> On the other hand, there are probably workarounds that are of equivalent level of effort for users, if less intuitive. 

Yeah, I'm torn.  The part where we pass through `cfg(test)` appropriately makes a lot of sense and seems like somewhat of a value add.  The thing is that `cfg(test)` isn't actually supported at build time in general, and that's when we need it.  The change to our CI to support it is rather unpleasant, and I imagine our users will find it the same.  This PR doesn't solve @kleinesfilmroellchen's problem out of the box by itself; they'll still need to get CARGO_CFG_TEST set at build time in their repo.

If there was a clean workaround that we could use here and recommend to our users, I think this would make sense.  The workaround in this PR right now is not something I particularly like.

> For example, creating a feature = test feature(https://github.com/lalrpop/lalrpop/issues/681 seemed to be explicitly against this option).

I wasn't super sure on that reading.  I'm guessing you're referring to their comment of "cfg(feature = blah) is nice, but I am only interested in cfg(test).".  I read that as more "The functionality I'm interested in is running in test mode, not feature configuration" rather than "I reject this workaround".  They may reject the workaround, but if it's explicitly mentioned in #681 as a workaround, I don't see it.  The "set a feature" workaround seems like possibly the least-bad option, but I can certainly sympathize with someone who wants to avoid it.

> We could alternatively close/stall this pr as "waiting for upstream".

That's my leaning at this point.  I'd love to support `cfg(test)`, and I don't have an issue with the parts of this that expose knowledge of it to lalrpop.  But they're useless by themselves.  We need CARGO_CFG_TEST set at build time, which cargo doesn't do.  And we need clean guidance for our users about how they can get CARGO_CFG_TEST set at build time.  I don't have good answers to either of those.

Pat-Lafon:
Converting this to a draft since it's unclear to me how to have a nice solution in the face of no build/test distinction at build time.

Issue #973, opened 2024-10-17
Document guards in macro


stephanemagnenat:
According to #921 and [this test](https://github.com/lalrpop/lalrpop/blob/d700700a5c32d31871067111b51ae86ed2a479b7/lalrpop/src/normalize/macro_expand/test.rs#L62-L73), lalrpop supports guards in macro. For certain use cases, such as when implementing a Rust-like language with optional trailing ";" in some control-flow expressions within blocks (as used in [this old partial Rust grammar implementation in lalrpop](https://github.com/nikomatsakis/rustypop/blob/master/src/rusty.lalrpop)), as the alternative is a lot of manual duplication.

I did not find any documentation of that feature, and as a user I think that it is worth documenting, hence this issue.

dburgener:
Thanks for the rustypop link!  I looked into documenting this feature a few months ago, but I couldn't figure out what the motivating use-case was.  The rustypop example helps with that.  It's too involved to use directly in docs though, so I'll need to think on how to simplify it down.

Can you elaborate on the "optional trailing ";"" use case you mentioned?  This is basically the same contextual thing rustypop is doing where in some contexts you want to mandate a trailing ";" and some you don't, so you pass that context through a macro argument?  If you have any other lalrpop code to point me to, that would be helpful.  Like I said, rustypop is helpful, but in trying to distill the feature down to a good example for the documents, more inspiration around straightforward motivating use cases would be nice.

stephanemagnenat:
> Can you elaborate on the "optional trailing ";"" use case you mentioned?

Yes. Rust allows to omit the trailing ";" when joining expressions in a block for certain expressions (called `ExpressionWithBlock` [here](https://doc.rust-lang.org/reference/expressions.html)). These expressions are essentially the ones that contain a mandatory block and therefore the trailing ";" is not strictly necessary to have a non-ambiguous grammar. However, while the resulting grammar is non-ambiguous, a naive implementation of it (e.g. e recursive tree of expressions and blocks) is, and some expressions should be disallowed in some contexts to avoid the ambiguity. The naive solution to this is to duplicate a large part of the grammar. The guard macro feature allows to avoid this, by keeping a single expression grammar "generic" over the macro parameters.

yannham:
[...]

PR #976, opened 2024-10-21
Added a new calculator example with blocks to demonstrate guards in macro feature.

stephanemagnenat:
This PR adds an example of macro guards. It motivates this by adding Rust-style blocks to the calculator, in which expressions are chained by ";". As Rust, it allows to omit the ";" when an expression is a block itself. Doing that DRY can be done using guards in macros. For a more detailed discussion, see #973.  

The new example extends the calculator 3 with the following features:
- block expressions in which expressions can be chained and the last one is returned.
- a print expression that prints a value (to motivate the chaining in the first place).

I am not sure that the number 3b is the right place to add this example, in particular because I am using "*" and "?" macros already.

### DIFF
--- a/doc/calculator/src/calculator3b.lalrpop
+++ b/doc/calculator/src/calculator3b.lalrpop
@@ -1,0 +1,64 @@
+use std::str::FromStr;
+
+grammar;
+
+match {
+    "+",
+    "-",
+    "*",
+    "/",
+    "(",
+    ")",
+    "{",
+    "}",
+    ";",
+    r"[0-9]+",
+    "print",
+
+    // Skip whitespace and comments
+    r"\s*" => { },
+    r"//[^\n\r]*[\n\r]*" => { }, // `// comment`
+    r"/\*([^\*]*\*+[^\*/])*([^\*]*\*+|[^\*])*\*/" => { }, // `/* comment */`
+}
+
+pub Expr = ExprRestricted<"B">;
+ExprNoBlock = ExprRestricted<"">;
+
+// - B: if non-empty, include block expressions
+ExprRestricted<B>: i32 = {
+    <l:ExprRestricted<B>> "+" <r:Factor<"B">> => l + r,
+    <l:ExprRestricted<B>> "-" <r:Factor<"B">> => l - r,
+    Factor<B>,
+};
+
+Factor<B>: i32 = {
+    <l:Factor<B>> "*" <r:Unary<"B">> => l * r,
+    <l:Factor<B>> "/" <r:Unary<"B">> => l / r,
+    Unary<B>,
+};
+
+Unary<B>: i32 = {
+    Term<B>,
+    "-" <e:Term<"B">> => -e,
+}
+
+Term<B>: i32 = {
+    Num,
+    "(" <e:Expr> ")" => e,
+    "print" "(" <e:Expr> ")" => { println!("{}", e); 0 },
+    Block if B != "",
+};
+
+Block: i32 = {
+    "{" <s:Stmt*> <e:ExprNoBlock?> "}" => e.unwrap_or_else(|| s.last().copied().unwrap_or(0)),
+}
+
+Stmt: i32 = {
+    <e:ExprNoBlock> ";" => e,
+    <b:Block> ";"? => b,
+};
+
+Num: i32 = {
+    r"[0-9]+" => i32::from_str(<>).unwrap(),
+};
+
--- a/doc/calculator/src/main.rs
+++ b/doc/calculator/src/main.rs
@@ -68,6 +68,23 @@ fn calculator3() {
     test3!(22 * (44 + 66) / 3);
 }
 
+lalrpop_mod_doc!(pub calculator3b);
+
+#[cfg_attr(not(test), allow(dead_code))]
+fn test3b(src: &str, result: i32) {
+    println!("parsing {}", src);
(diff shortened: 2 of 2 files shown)


dburgener:
Thanks for the PR!

> I am not sure that the number 3b is the right place to add this example, in particular because I am using * and ? macros already.

I'm not strongly wedded to any particular numbering, but calculator5.lalrpop seems to me like the more obvious place to extend - this is an extension after introducing macros.  Is there a particular reason to prefer extending calculator3 vs calculator5?

Regardless of which calculator is extended, this should get some extra verbage in the book.  I think it makes sense to group it in the "Advanced Setup" section, as 7.4 after "Location Tracking".  We can just refer back to the macro section there and augment a previous calculator.  Do you want to write up the doc section for the book as well?  If not, I can write up a description of this and push onto the PR.

stephanemagnenat:
> Is there a particular reason to prefer extending calculator3 vs calculator5?

`calculator5` is using AST and I did not want to complexify `ast.rs` with things like `Block` just for this example. But of course we could.

Section 7.4 seems good. In that case, we could even use calculator with a fully different number, like calculator10.

I can try writing a section for the book. I've seen that contrary to the other sections, "Advanced setup" does not have its own directory. Is that on purpose? I'm happy to follow this structure of course and could add a `guards_in_macro.md` file directly in `doc/src/`.

dburgener:
> calculator5 is using AST and I did not want to complexify ast.rs with things like Block just for this example. But of course we could.

Good point, that seems fair.

> Section 7.4 seems good. In that case, we could even use calculator with a fully different number, like calculator10.

Yeah, I like the calculator10 idea.

> I can try writing a section for the book. I've seen that contrary to the other sections, "Advanced setup" does not have its own directory. Is that on purpose? I'm happy to follow this structure of course and could add a guards_in_macro.md file directly in doc/src/.

That's before my time, so I can't say for sure.  I imagine it's due to the history of this growing from various one-off docs.  I'm not personally tied to the current layout, and it would surprise me a bit if anyone else is.  I'm good either way if you want to honor the existing layout and put your new doc in `doc/src` like you said, or if you'd prefer to clean up "Advanced setup" while you're in there.

Thanks!

dburgener:
While we're updating references to this functionality elsewhere in the documentation, [this line](https://github.com/lalrpop/lalrpop/blob/2641493eb3ec3dd1509e0743ab6611a79b9abd12/doc/src/tutorial/index.md?plain=1#L28) references this functionality as something that needs to be documented in the future, so once the text about the feature is added, that line should be deleted too.

dburgener:
@stephanemagnenat is this something you're still intending to finish up?  I'm happy to take it over if needed, but it sounded like you were going to take a stab at the section for the book, and the other mentioned doc cleanups should be pretty quick and easy.

stephanemagnenat:
I hope to still finish it, but I'm overloaded with work this end of year, so if you have bandwidth, please feel free to take over.

dburgener:
Great!  As long as it isn't dropped, I'm fine with waiting.  We can revisit in the new year.

stephanemagnenat:
@dburgener I have added the documentation as requested. Can you or someone else please review?

doc/src/cheatsheet.md
```diff
@@ -18,5 +18,5 @@ functionality, use this table to jump to the right section.
 | extern_location | `type Location = MyLoc;` | sets the type to for locations instead of `usize` | [Writing a custom lexer](lexer_tutorial/003_writing_custom_lexer.md) |
 | extern_tok | `enum MyToken { }` | declares the type of lexer tokens to be consumed by the generated parser  | [Using tokens with references](lexer_tutorial/004_token_references.md) |
 | auto_parameters | `<>` | refers to all the parameters of the non-terminal as a tuple | [Type inference](tutorial/003_type_inference.md) |
-|conditional actions | `Expr<I> = { ... , <T> if I == "a" => (), ...}` | Conditional definition of a macro's alternative | [index pointer](tutorial/index.md) |
+|conditional actions | `Expr<I> = { ... , <T> if I == "a" => (), ...}` | Conditional definition of a macro's alternative | [index pointer](guards_in_macro.md) |
```
Pat-Lafon, line 21: Rename `index pointer`

doc/src/guards_in_macro.md
```diff
@@ -0,0 +1,70 @@
+# Guards in macro
+
+Sometimes, you may want two very similar set of declarations, but with a few changes. A typical example is when implementing a Rust-like language with optional trailing ";" in some control-flow expressions within blocks. In that case, you want to allow omitting the trailing ";" only after a block, because in that case the grammar is unambiguous.
```
Pat-Lafon, line 3: I feel like the start of this is a bit vague/couched, I was having a hard time understanding what the setup + use case is.

Can we use an example here? I think we want to say something like "you want to do this but Lalrpop tells you your grammar is ambigious. We know some additional contextual information that we can pass through the rules which will be expanded into two slightly different cases. Lalrpop now accepts this grammar and the input as expected while rejecting this other thing we didn't want to accept `{1 2}`.

If you have a different way to show this off with examples go for it!

Pat-Lafon:
@stephanemagnenat Thanks for pushing this forward! I have two high level suggestions to try to make this documentation a little bit more clear and feature-focussed. Let me know what you thing. It does seem like the ci wants you to break up a few lines that go over 80 characters.

PR #977, opened 2024-10-22
Use our README as a top level docstring

dburgener:
This shows on the homepage of lib.rs, and is currently blank.  The README contains lots of good information and points users at the various documentation sources, so make it more visible to people checking out the crate on docs.rs.

While we're at it, warn about missing docs (which would have caught this omission).

### DIFF
--- a/lalrpop/src/lib.rs
+++ b/lalrpop/src/lib.rs
@@ -1,3 +1,5 @@
+#![doc = include_str!("../../README.md")]
+#![warn(missing_docs)]
 // Need this for rusty_peg
 #![recursion_limit = "256"]
 // I hate this lint.


yannham (approved):
I didn't even know about this Rust attribute, that's cool! LGTM

dburgener:
> https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html

Yeah, it was new to me yesterday.  I noticed that we were missing a top level docstring, so I was digging around documentation about writing documentation and came across the [rustdoc book](https://docs.rs/lalrpop/latest/lalrpop/) which mentioned that.  I haven't read a particularly high percentage of the rustdoc book yet, but I hope to.  I imagine it has other suggestions for how to improve our autogenerated documentation.

Pat-Lafon (approved):

PR #978, opened 2024-10-22
Specify optional regex_automata as "dep:".

dburgener:
https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

I recently learned that optional dependencies will be exposed as features unless the are explicitly marked as "dep:" for a another feature.  This means that lalrpop-util has a feature called "regex-automata" currently (see https://docs.rs/crate/lalrpop-util/0.22.0/features), which isn't what we want.  We want to have the dependency optionally when certain other features are enabled, but not expose it directly.

This is technically a breaking change
(https://doc.rust-lang.org/cargo/reference/semver.html#cargo-feature-remove). However, it would surprise me if any users use this feature.  In the interest of caution, it might make sense to hold this until we plan a breaking release anyways, since it's low priority.

### DIFF
--- a/lalrpop-util/Cargo.toml
+++ b/lalrpop-util/Cargo.toml
@@ -17,7 +17,7 @@ categories.workspace = true
 regex-automata = { workspace = true, optional = true }
 
 [features]
-lexer = ["regex-automata"]
+lexer = ["dep:regex-automata"]
 unicode = ["regex-automata?/unicode"]
 std = ["regex-automata/perf"]
 default = ["std", "unicode"]


dburgener:
This worked on my local box, but failed here.  It turns out that there was a bug around this that was fixed in rust 1.71:

https://github.com/rust-lang/cargo/issues/10788
https://github.com/rust-lang/cargo/pull/12130

So we could either MSRV bump to 1.71 or greater for this, or use the workaround in the first link.

dburgener:
I converted this to a draft.  I think there are two obstacles to taking it in the short term:

* The change itself is breaking, by removing a feature
* To do the change properly without a workaround requires rust 1.71, which is an MSRV bump, with is pseudo-breaking as well.

Since this is a low priority change, my opinion is that we should hold off on it until we are doing a breaking change anyways for some other reason, and take this along with that.  We can bump the MSRV at that point as well - either to 1.71 as needed here, or potentially to 1.81 for core::error support as discussed in https://github.com/lalrpop/lalrpop/pull/957.

Pat-Lafon:
> I converted this to a draft. I think there are two obstacles to taking it in the short term:
> 
>     * The change itself is breaking, by removing a feature

As you have mentioned, I rather doubt that anyone is using this feature. Does enabling it even really do anything on it's own? My understanding of what you have said, given that we don't feature gate on this "feature", is that it just says to also compile regex_automata.

>     * To do the change properly without a workaround requires rust 1.71, which is an MSRV bump, with is pseudo-breaking as well.

> Since this is a low priority change, my opinion is that we should hold off on it until we are doing a breaking change anyways for some other reason, and take this along with that. We can bump the MSRV at that point as well - either to 1.71 as needed here, or potentially to 1.81 for core::error support as discussed in #957.

Yup, though I would be perfectly fine setting the msrv to 1.81 to both remove our core::error hack and in keeping with our msrv policy.

dburgener:
> As you have mentioned, I rather doubt that anyone is using this feature. Does enabling it even really do anything on it's own? My understanding of what you have said, given that we don't feature gate on this "feature", is that it just says to also compile regex_automata.

That's a really good point.  I think you may be right that it isn't technically breaking for that reason.

> Yup, though I would be perfectly fine setting the msrv to 1.81 to both remove our core::error hack and in keeping with our msrv policy.

We should discuss what (if anything) to bump our msrv to.  This PR needs 1.71 or later.  #981 needs 1.74 or later.  #979 doesn't need a bump, but could take advantage of `fs::exists` if we go to 1.81 or later.  And there's the hack in lalrpop-util that can be dropped with 1.81 or later.  I'm of two minds.

1. We're at sort of a "middle" place in the ecosystem.  We're not one of those fundamental libraries that's trying to support an ancient version because they're used everywhere.  But we are a library, and at least some of our users are libraries themselves, so we do ripple forcing users to a more recent msrv with a bump.
2. The community is divided about whether msrv bumps are strictly breaking changes, and the recommendation I understand is to try to package msrv bumps with breaking changes if possible.  If we're going to bump msrv, there is benefit to doing the sort of break once and getting it over with for a while.  Particularly with 1.81 stabilizing `core::error`, that may end up being the sort of version that many crate list as an msrv over the long term.  My speculation is that it's only a matter of time before one of our dependencies bumps to 1.81 and we have to move forward.  Going to 1.81 now (by which I mean after 1.83 is stable) would somewhat future-proof us for hopefully at least the next year or so.

I would love to stop publishing so many breaking changes, since point releases are near-invisible to our users and our improvements get out to the community faster.  However, #979 seems to me to be both 1. breaking and 2. a bug of some importance to get fixed.  So it might make sense to take the opportunity to go all the way to 1.81 on our msrv when we go to 0.23.

dburgener:
Now that #1001 has merged, the MSRV check no longer fails here, and as @Pat-Lafon pointed out above, this isn't truly a breaking change.  So I've gone ahead and published this, as I think it's ready to go.

Pat-Lafon (approved):
Seems like this foot-gun was going to be fixed in the next edition but it's been delayed due to implementation issues: https://github.com/rust-lang/cargo/issues/12826. That doesn't effect us here, so I'm glad to clean this up.

PR #979, opened 2024-10-25
Stop using in_dir once we get to process_* family functions

dburgener:
The prior code has weird corner cases around the difference between in_dir and the path passed to process_dir().  In many cases, they are the same, but sometimes they are not.  When they are not, the only behavior of in_dir is to effect where *output* files are placed, which is weird.  Depending on the various function calls, sometimes in_dir sets the path to search, but sometimes not, without rhyme or reason.

This commit changes in_dir to set the path to search if process() is called specifically.  Other process_*() family functions have a built in path to search, in which case in_dir is illegal.

Changing the output path based on the setting of in_dir is confusing and unexpected for users.  Drop that.  That seems to have originally been maintained for backwards compatibility.  The result of this commit is that sometimes generated .rs files that were previously written in the source repository will end up in the OUT_DIR instead, which is a breaking change.  That said, OUT_DIR really should be the correct and preferred place for generated code.  Having users explicitly opt out from that makes more sense than opting them out surprisingly based on the in_dir setting.

For historical discussion on this, see:
https://github.com/lalrpop/lalrpop/issues/908#issuecomment-2176200995
https://github.com/lalrpop/lalrpop/issues/280

### DIFF
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -5,6 +5,7 @@ use std::default::Default;
 use std::env;
 use std::env::current_dir;
 use std::error::Error;
+use std::io;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 
@@ -208,6 +209,15 @@ impl Configuration {
     /// Process all files in the current directory, which -- unless you
     /// have changed it -- is typically the root of the crate being compiled.
     pub fn process_current_dir(&self) -> Result<(), Box<dyn Error>> {
+        // If we can get a current dir, check to make sure session.in_dir either *wasn't* set, or
+        // wasn't set to that dir.  If we can't get a current dir, we'll error out in a moment
+        // anyways, and that's the bigger problem.
+        if let Ok(current_dir) = current_dir() {
+            if self.session.in_dir.is_some() && self.session.in_dir != Some(current_dir) {
+                eprintln!("Error: \"process_current_dir()\" contradicts previously set in_dir");
+                return Err(Box::new(io::Error::from(io::ErrorKind::InvalidInput)));
+            }
+        }
         self.process_dir(current_dir()?)
     }
 
@@ -215,14 +225,8 @@ impl Configuration {
     pub fn process_dir<P: AsRef<Path>>(&self, path: P) -> Result<(), Box<dyn Error>> {
         let mut session = self.session.clone();
 
-        // If in/out dir are empty, use cargo conventions by default.
+        // If out dir is empty, use cargo conventions by default.
         // See https://github.com/lalrpop/lalrpop/issues/280
-        if session.in_dir.is_none() {
-            let mut in_dir = env::current_dir()?;
-            in_dir.push("src");
-            session.in_dir = Some(in_dir);
-        }
-
         if session.out_dir.is_none() {
             let out_dir = env::var_os("OUT_DIR").ok_or("missing OUT_DIR variable")?;
             session.out_dir = Some(PathBuf::from(out_dir));
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -9,6 +9,9 @@ use std::io;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 
+#[cfg(test)]
+mod test;
+
 /// Configure various aspects of how LALRPOP works.
 /// Intended for use within a `build.rs` script.
 /// To get the default configuration, use `Configuration::new`.
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -218,7 +218,7 @@ impl Configuration {
         if let Ok(current_dir) = current_dir() {
             if self.session.in_dir.is_some() && self.session.in_dir != Some(current_dir) {
                 eprintln!("Error: \"process_current_dir()\" contradicts previously set in_dir");
-                return Err(Box::new(io::Error::from(io::ErrorKind::InvalidInput)));
+                return Err(Box::new(io::Error::new(io::ErrorKind::InvalidInput, "\"process_current_dir()\" contradicts previously set in_dir.  Either use `process()` instead, or omit `set_in_dir()`.  (Note: in previous versions of lalrpop, this combination could affect the parser output dir.  If you were relying on this behavior to output the parser in your source directory, you may want to use `set_out_dir()` to retain that behavior.")));
             }
         }
         self.process_dir(current_dir()?)
--- a/lalrpop/src/api/test.rs
+++ b/lalrpop/src/api/test.rs
@@ -1,0 +1,168 @@
+use std::env::{current_dir, set_current_dir, set_var, temp_dir};
+use std::fs;
+use std::path;
+use std::sync::{LockResult, Mutex, MutexGuard};
+
+use super::*;
+
+// tests may be run in parallel and these tests affect global state, so lock
+static API_TEST_MUTEX: Mutex<i32> = Mutex::new(0);
+
+const TEST_DIR: &str = "lalrpop-test";
+const CUSTOM_TEST_DIR: &str = "lalrpop-test2";
+
+#[derive(Debug, PartialEq)]
+enum GenFileLoc {
+    Src,
+    Other,
+    Root,
+    OutDir,
+    CustomOut,
+    DoesntExist,
+}
+
+// Set up for API tests.  The directory structure in test_files
+// is:
+//
+// outer.lalrpop
+// other
+//   - other.lalrpop
+// src
+//   - src.lalrpop
+//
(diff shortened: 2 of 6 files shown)


dburgener:
An additional note is that our test coverage for the api is lacking.  I did some local testing on this and it all seems good.  I wanted to add automated tests, but it's somewhat tricky since:

* OUT_DIR isn't set at runtime (when tests run), just at build time
* A lot of the api uses "src", which we already use for our other test files
* Clearing state between tests when we generated files in weird locations is strange
* We need to do all the building, then all the testing, so we can't reuse the same files with different build commands

I tried poking a little bit at setting up a directory structure in lalrpop-test for api testing files and adding additional build commands to build.rs to build them under various configurations, but it was hard to test all the configurations simultaneously in a meaningful way, since so many of our helper functions want to browse an entire directory tree, expressed as a relative path to CWD.  If we could put the build commands directly in test code, we could build, check, cleanup and repeat on the same input files.  But in the test code OUT_DIR isn't defined, and a lot of the config commands assume it is.

So, the result is that we still don't have automated testing for this.

dburgener:
I just realized the obvious answer to the test problems. If we just use setenv to set OUT_DIR manually we can do everything at runtime, and output to a directory that we control and can clean up. I'll work on support for that approach.

dburgener:
The original code is actually more broken than I thought.  If you do `set_in_dir(Path::new("."))`, it tries to strip the prefix "." from the output path and when that fails, puts files in your source dir regardless of your out_dir configuration.  So `in_dir` really needs to be an absolute path for `out_dir` to work.

dburgener:
Okay, I wrote some unit tests for the api module with some framework to manipulate environmental variables appropriately and various helper functions.

Not everything has coverage yet, but I hit the common cases, including some of the cases that this PR fixes (eg without the first commit `test_process_root()` here would fail, since other.rs and outer.rs would get written in the source directory).

Pat-Lafon (changes requested):
I apologize for putting off reviewing this pr for so long.

I've read through your discussion/linked comments and I agree that the current behavior is rather unintuitive. Overall the non-test-related diff looks cleaner and I agree with your reasoning.

It might be inaccurate but my recollection is that this will be one of the larger breaking changes we have made? Given that, I want to understand exactly what is the migration path for users/what we would write in the changelog/release message.

lalrpop/src/api/mod.rs
```diff
@@ -208,21 +212,24 @@ impl Configuration {
     /// Process all files in the current directory, which -- unless you
     /// have changed it -- is typically the root of the crate being compiled.
     pub fn process_current_dir(&self) -> Result<(), Box<dyn Error>> {
+        // If we can get a current dir, check to make sure session.in_dir either *wasn't* set, or
+        // wasn't set to that dir.  If we can't get a current dir, we'll error out in a moment
+        // anyways, and that's the bigger problem.
+        if let Ok(current_dir) = current_dir() {
+            if self.session.in_dir.is_some() && self.session.in_dir != Some(current_dir) {
+                eprintln!("Error: \"process_current_dir()\" contradicts previously set in_dir");
```
Pat-Lafon, line 220: Is it accurate that this can be reached from `process_root()` as well? If so, lets not used the function name explicitly(or I guess add checking to process_root as well but that would duplicate the logic)

Pat-Lafon: Hmm, actually I guess it's reachable but will never occur because `process_root` creates its own Configuration object and an `in_dir` isn't set

Pat-Lafon, line 220: Is there a typical fix that we expect users to make? Like drop `set_in_dir(P)` and maybe update `set_out_dir`? I would like to give a more actionable error message if possible

lalrpop/src/api/mod.rs
```diff
@@ -208,21 +212,24 @@ impl Configuration {
...
+                return Err(Box::new(io::Error::from(io::ErrorKind::InvalidInput)));
```
Pat-Lafon, line 221: What would be the difference between using `Error::from(..)` versus `Error::new(io::ErrorKind::InvalidInput, "The above error message here")`. I think the latter means that the our error message will be more apparent? My intuition could be mistaken

lalrpop/src/api/test.rs
```diff
@@ -0,0 +1,168 @@
+use std::env::{current_dir, set_current_dir, set_var, temp_dir};
+use std::fs;
+use std::path;
+use std::sync::{LockResult, Mutex, MutexGuard};
+
+use super::*;
+
+// tests may be run in parallel and these tests affect global state, so lock
+static API_TEST_MUTEX: Mutex<i32> = Mutex::new(0);
```
Pat-Lafon, line 9: I'm on board with using a Mutex here to make sure the tests are somewhat isolated. I was a little concerned with all of the panic'ing while holding a mutex(since I don't often reach for Mutex myself), but I imagine if there were issues with mutex poisoning, the test suite is failing anyways so it doesn't matter too much.

I will note that the comments I found also recommended https://docs.rs/serial_test/latest/serial_test/ as a dev dependency but also say that a mutex is completely reasonable.

Pat-Lafon:
The tests took me a little bit of time to grok, but they make sense to me. I appreciate that figuring this out was non-trivial.

dburgener:
> It might be inaccurate but my recollection is that this will be one of the larger breaking changes we have made?

Probably depends on how you define "large".  The change to rename public types and enums to camel case in 0.20.0 probably affected most users, but the fix to update was extremely trivial.  This should hopefully not affect too many people, but for those that it does impact, they will likely have to do a bit of work to resolve it.

> Given that, I want to understand exactly what is the migration path for users/what we would write in the changelog/release message.

Good question.  Let me take a stab at some text here that we can use in the changelog at release time:

```
#### Breaking Changes

* `Configuration.set_in_dir()` can now no longer affect the output dir.  In certain configurations, setting in_dir using `set_in_dir()` could actually cause the generated parser to be put in the input directory rather than in `OUT_DIR`.  In most cases `OUT_DIR` may be what you want anyways and should work seamlessly.  If you want to preserve the existing behavior of writing to the source directory, you should do that explicitly with `set_out_dir()` (or `use_cargo_dir_conventions()`)
* `Configuration.set_in_dir()` is now incompatible with `process_current_dir()`, and `process_dir()` , because all of these functions set an input directory themselves (either the current directory, or the argument).  Previously, these functions would all silently ignore the `set_in_dir()` setting, except possibly to change the *output* directory as mentioned above.  Now these functions are incompatible.  If you get an error about this, remove `set_in_dir()` from your `Configuration`, or use it with `process()`.  You may also need to change your output directory as mentioned in the previous bullet.
* ```

File: lalrpop/src/api/mod.rs (hunk 1 above)
Comment (dburgener) [reply lalrpop/src/api/mod.rs:220]: Looking over everything, I do think this is the only place the message is reachable, as you said above.

File: lalrpop/src/api/test.rs (hunk 3 above)
Comment (dburgener) [reply lalrpop/src/api/test.rs:9]: The `serial_test` crate looks cool to me. I didn't come across that when looking into it.  That would probably simplify the code a bit.  I vote that we not block this PR over that, but keep it in mind for a future enhancement.

File: lalrpop/src/api/mod.rs (hunk 2 above)
Comment (dburgener) [reply lalrpop/src/api/mod.rs:221]: I would think that if the user is using `.unwrap()` on the result, which I think is the recommended and normal approach, then the code as written would display the message and then panic, whereas if we change to your suggestion, it would panic with that message.  If the user is doing something else, then not displaying the message would allow them to suppress it.  Which I guess is desirable.  I'll update to that approach.

File: lalrpop/src/api/mod.rs (hunk 1 above)
Comment (dburgener) [reply lalrpop/src/api/mod.rs:220]: Updated test.  Will push in a moment.

Comment (Pat-Lafon) [reply lalrpop/src/api/mod.rs:220]: Thanks!

File: lalrpop/src/api/test.rs [hunk 4]
```diff
@@ -0,0 +1,168 @@
...
+// other
+//   - other.lalrpop
+// src
+//   - src.lalrpop
+//
+// So we want to set CWD to directly above that, and OUT_DIR to a temp directory
+fn setup() -> (path::PathBuf, LockResult<MutexGuard<'static, i32>>) {
+    let lock = API_TEST_MUTEX.lock();
+    let orig_dir = current_dir().unwrap();
+    set_current_dir(path::Path::new("./src/api/test_files")).unwrap();
```
Pat-Lafon, line 37: I did a quick double check and Rust does do the right thing here for Windows https://doc.rust-lang.org/std/path/struct.Path.html

Pat-Lafon (approved):
Thanks. I'm now more confident about this. It shouldn't matter but like in #1055 can we rebase this against the windows action pr? All of the path/file things should #JustWork on windows

PR #981, opened 2024-10-25
Benchmarking

dburgener:
Add benchmarking support.

### DIFF
--- a/lalrpop-test/benches/lalrpop_benchmark.rs
+++ b/lalrpop-test/benches/lalrpop_benchmark.rs
@@ -1,0 +1,12 @@
+use criterion::criterion_main;
+
+mod src {
+    pub mod compile_benches;
+    pub mod parser_benches;
+    pub mod parsers;
+}
+
+use crate::src::compile_benches::compile;
+use crate::src::parser_benches::parser;
+
+criterion_main!(compile, parser);
--- a/lalrpop-test/benches/src/parsers/json.lalrpop
+++ b/lalrpop-test/benches/src/parsers/json.lalrpop
@@ -1,0 +1,43 @@
+use std::collections::HashMap;
+use std::str::FromStr;
+use crate::src::parsers::json_val::Value;
+
+grammar;
+
+// https://datatracker.ietf.org/doc/html/rfc7159
+
+pub Value: Value = {
+	Object => Value::Object(<>),
+	Array => Value::Array(<>),
+	Number => Value::Num(<>),
+	String => Value::Str(<>),
+	"false" => Value::Boolean(false),
+	"null" => Value::Null,
+	"true" => Value::Boolean(true),
+};
+
+Object: HashMap<String, Value> = {
+	"{" <Comma<Member>> "}" => HashMap::from_iter(<>)
+};
+
+Member: (String, Value) = {
+	<s: String> ":" <v: Value> => (s,v), 
+};
+
+Array: Vec<Value> = {
+	"[" <Comma<Value>> "]",
+};
+
+Number: f64 = {
+	r"-?(0|([1-9][0-9]*))(\.[0-9]*)?([eE][-+]?[0-9]+)?" => f64::from_str(<>).unwrap()
+};
+
+String: String = {
+	r#""[^"]*""# => <>.into(),
+};
+
+Comma<V>: Vec<V> = {
+	<v: (<V> ",")*> <e: V?> => {
+		v.into_iter().chain(e).collect()
+	}
+};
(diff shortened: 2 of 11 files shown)


dburgener:
I'm leaving this as a draft for the time being, but wanted to get it up in the context of #979.  This mostly works, but because of the in_dir setting out_dir bug in #979, it does two undesirable things:

1. It overwrites the local checked in lrgrammar.rs with a newly generated grammar (that doesn't match.  I guess that's about `--no-whitespace` in update_lrgrammar.sh?)
2. It puts the generated json.rs in the source dir rather than OUT_DIR.

I tried to work around these by modifying build.rs, but I couldn't get it working right, since `src` is the package root for lalrpop-test proper, while `.` is the package root for lalrpop/benches.  I guess maybe two separate build commands could get it working if we want to take this before #979, but I haven't gone down that route yet, because #979 should (in theory) just fix this to work correctly as is.

lalrpop-test/benches/src/parsers/json.lalrpop
```diff
@@ -0,0 +1,43 @@
...
+	"[" <Comma<Value>> "]",
+};
+
+Number: f64 = {
+	r"-?(0|([1-9][0-9]*))(\.[0-9]*)?([eE][-+]?[0-9]+)?" => f64::from_str(<>).unwrap()
+};
+
+String: String = {
+	r#""[^"]*""# => <>.into(),
+};
```
Pat-Lafon, line 37: Just curious, do other implementations we might benchmark against try to avoid allocating a bunch of new strings here? I could imagine just having a `&'data str` or an arena allocator here. On the other hand, if other parsers typically do this, then its more important to have an apples-to-apples comparison.

Basically, should we be worried about the cost of allocations in this benchmark?

dburgener: "benchmark against" is sort of an interesting term here.  I guess a lot of this depends on our goal.

This file is directly taken from the parser in parse-rosetta-rs, where we are directly compared head to head against other parsers.  In parse-rosetta-rs, glancing through the parsers, it looks like they mostly return allocated `String`s in their data.  Pest seems to use an internal type that takes a lifetime, so may not be allocated, and serde does whatever serde does, but it's a specialized json parser, not a parser generator.  So while it's not universal, we're generally playing by the rules there and having a fair-ish comparison.

Locally, I view these numbers as much more about benchmarking against ourselves, than against any external comparison.  So whatever we do here it will be apples-to-apples unless we're changing the benchmark.  This will allow us to compare PRs that we have performance questions about (eg #952 or #814) without spending a bunch of effort hooking the PR up to an external tool like parse-rosetta-rs.  To that end, the allocation will be apples-to-apples either way.  We might care in terms of allocation dominating performance though.  If allocation does turn out to be very expensive here, then performance changes could appear less impactful than they actually are, causing us to make bad decisions.  I can try to spin up a version of this that skips the allocation and see how expensive it is.  Not sure if I'll have time before the holidays or not, so it may be in 2025.

dburgener: Thanks!

How would you feel about just having both versions as different benchmarks?  I agree that based on those performance numbers the allocation-less version is better, so we definitely want that one.  Having the allocation one as well has two benefits:

1. It lets us know how we'll impact the parse-rosetta-rs metric
2. It keeps a live comparison of the cost of allocations to help answer questions around the cost of allocations going forward.

Pat-Lafon: Sure, I've re-added back the initial variation in that pr. You should be able to run the parser benchmarks and see a comparable speed-up.

dburgener: I cherry-picked the first two commits from that PR into here.  Thanks!

dburgener:
It took a few tweaks, but with #979 merged, the issues mentioned above are all resolved.  This seems to functionally work as expected as far as I can tell in my local testing.

teofr:
Hey, I was just looking at these benchmarks and tinkering with them locally, I noticed because of the [`build.rs`](https://github.com/dburgener/lalrpop/blob/65e294f1f535d48de0c6b79cf87ed97323831f13/lalrpop-test/build.rs#L5) in `lalrpop-test` both json grammar are compiled as test parsers, which mean they'll parse using both ascent and LR(1) parsers, and then compare the results.

If this is intended apologies for the noise here, but I think if the goal is to benchmark both types of parsers splitting the benchmarks could be better. Measuring the time of the comparison seems strange.

lalrpop-test/benches/lalrpop_benchmark.rs
```diff
@@ -0,0 +1,14 @@
+use criterion::criterion_main;
+
+mod src {
+    pub mod compile_benches;
+    pub mod parser_benches;
+    // This can just use lalrpop_mod!() once #979 merges
+    #[rustfmt::skip]
+    pub mod parsers;
```
Pat-Lafon: Just noting, is this good now? https://github.com/lalrpop/lalrpop/pull/979

PR #984, opened 2024-10-30
Add some doc comments to lalrpop-util

dburgener:
This isn't everything triggered by warn(missing-docs).  I tried to hit the low hanging fruit and important stuff, but there are still quite a few gaps, so we can't turn on the lint for it yet.


dburgener:
We might consider doing `allow(missing-docs)` in state_machine.rs and lexer.rs.  As far as I can tell, those are only public in order to be linked by the generated parsers, and aren't intended for user consumption (which is what I wrote in the module level docs strings).  There is certainly some value in better comments (I don't understand a lot of it, which is a big part of why I haven't commented it), but calling this a "public interface" is technically true but practically not really.

Pat-Lafon:
> We might consider doing `allow(missing-docs)` in state_machine.rs and lexer.rs. As far as I can tell, those are only public in order to be linked by the generated parsers, and aren't intended for user consumption (which is what I wrote in the module level docs strings). There is certainly some value in better comments (I don't understand a lot of it, which is a big part of why I haven't commented it), but calling this a "public interface" is technically true but practically not really.

Would https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#hidden help here?

Pat-Lafon (approved):

dburgener:
> Would https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#hidden help here?

Yeah, that does seem like a good idea, thanks for suggesting it!  I think leaving the top level doc comments that I wrote in addition to that makes sense even though they won't appear in the docs for anyone that stumbles across them in the codebase.

I just pushed that version.  Let me know what you think.  It does remove all the missing doc warnings in lexer.rs and state_machine.rs, but there are still a few unhandled cases in lib.rs that I do think we want to document.  The remaining set of warnings is:

```
warning: missing documentation for a struct field
  --> lalrpop-util/src/lib.rs:52:20
   |
52 |     InvalidToken { location: L },
   |                    ^^^^^^^^^^^
   |
note: the lint level is defined here
  --> lalrpop-util/src/lib.rs:13:9
   |
13 | #![warn(missing_docs)]
   |         ^^^^^^^^^^^^

warning: missing documentation for a struct field
  --> lalrpop-util/src/lib.rs:77:18
   |
77 |     ExtraToken { token: (L, T, L) },
   |                  ^^^^^^^^^^^^^^^^

warning: missing documentation for a struct field
  --> lalrpop-util/src/lib.rs:80:12
   |
80 |     User { error: E },
   |            ^^^^^^^^

warning: missing documentation for a method
   --> lalrpop-util/src/lib.rs:112:5
    |
112 |     pub fn map_location<LL>(self, op: impl FnMut(L) -> LL) -> ParseError<LL, T, E> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: missing documentation for a method
   --> lalrpop-util/src/lib.rs:116:5
    |
116 |     pub fn map_token<TT>(self, op: impl FnOnce(T) -> TT) -> ParseError<L, TT, E> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: missing documentation for a method
   --> lalrpop-util/src/lib.rs:120:5
    |
120 |     pub fn map_error<EE>(self, op: impl FnOnce(E) -> EE) -> ParseError<L, T, EE> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: `lalrpop-util` (lib) generated 6 warnings
```

The first few about the struct fields are straightforward enough.  The only reason I didn't do them already was that they didn't strike me as super high value and I wasn't going all the way to support `warn(missing_docs)` anyways.  The second set with the `map_*()` functions I've never used those functions and would just need to dig in more to understand them before writing documentation.

Pat-Lafon:
I haven't used the map_* functions either but there are some example usages in https://github.com/lalrpop/lalrpop/pull/254.

I think this is good step forward, we can always have later prs to handle some of the remaining set.

Issue #759, opened 2023-04-22
remove recursive-ascent, LALR support for 1.0, and other non-recommended modes


nikomatsakis:
LALRPOP has a lot of modes. I propose we remove some of them for 1.0 and simplify our lives. For example:

* Remove recursive ascent. Generates too much code. Probably more performant, but I'm not aware of anyone who cares. Maybe we can search github or crates.io to see.
* Remove LALR support. The lane table algorithm is fine.
* Remove the option to generate code into the src directory and only support the `lalrpop_mod` approach.

Thoughts @lalrpop/maintainers ?

dburgener:
Searching github a little for `#[recursive_ascent]` and `#[LALR]` reveals a number of users.  Projects that seem active include:

Recursive ascent:

* https://github.com/gwenn/sqlpop

(Only two users on github, the other was last updated in 2017).

LALR:

* https://github.com/Storyyeller/cubiml-demo
* https://github.com/polarity-lang/polarity/tree/main
* https://github.com/miguel-nascimento/algorithm-j-rs

As well as quite a number of other projects that haven't been updated recently (or have no users).

On the generate_in_source topic see also #946 where that came up again.

dburgener:
It would be nice if we could mark `#[LALR]` as deprecated in a way that our users will see.  We [try to do that](https://github.com/lalrpop/lalrpop/blob/bda35a4c77eea4aa537c95be6b80f90ff766e217/lalrpop/src/lr1/build_lalr/mod.rs#L32-L33) now by printing a message, but cargo suppresses output in build scripts, which is how lalrpop is designed to be used.

https://github.com/rust-lang/cargo/issues/3777

The [cargo warning](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargo-warning) functionality only displays if one of the following is true:

* The build script fails
* The warning comes from a local crate (as opposed to crates.io)
* The user enables `-vv` when running cargo

The common case is that none of the above are true for most of our users.  It would be nice to display a "#[LALR] is deprecated and will be removed in version x.x.x" message, but I don't think it's possible. The loudest we are able to be about this is documentation.

Pat-Lafon:
Hmm, what if we put a deprecation attribute with a message on it on the generated parser/parse function? https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute 

Calls to `.parse()` on such a generator would then raise a deprecation warning at compile time.

PR #990, opened 2024-11-15
Drop lalr support

dburgener:
As suggested in https://github.com/lalrpop/lalrpop/issues/759.

This makes Lane tables the only available algorithm to generate LR(1)
states.  In theory, lane tables are strictly better, and they have been
the default in lalrpop for quite some time.

Searching github reveals no public users of LALR that aren't directly lalrpop
test code.

This is of course a breaking change, since it drops support for an
existing feature.


dburgener:
I'm converting to draft.  I didn't realize there were a number of users (see list in #759).  I'm not particularly convinced that any of them want to be using LALR.  I might look into seeing seeing about migrating active projects away from LALR mode first.

PR #995, opened 2024-11-18
replace thread rng with deterministic and portable seeded generator

### LINKED ISSUE #994: Our CI has a non-deterministic test
I'm having a little trouble reproducing it. Apparently that test case does some random testing? https://github.com/lalrpop/lalrpop/blob/ab154926da43fbbb9107bd374cef4159aff1fa4c/lalrpop/src/lr1/build/test.rs#L167

I'm interested in random testing in general, but it might not be great to have it in CI.

There might even be a comment about this failure mode? https://github.com/lalrpop/lalrpop/blob/ab154926da43fbbb9107bd374cef4159aff1fa4c/lalrpop/src/generate.rs#L13-L27

Since the call stack was exceeded, we didn't get a printout.

_Originally posted by @Pat-Lafon in https://github.com/lalrpop/lalrpop/issues/993#issuecomment-2484091509_

Pat-Lafon:
Fixes #994. 

I was looking at Rand's documentation and in https://rust-random.github.io/rand/rand/rngs/struct.StdRng.html they note that while some of their generators are deterministic, they may not be portable(and thus might change over various system updates).

In following their documentation (https://rust-random.github.io/book/guide-seeding.html#a-simple-number), I've pulled in Rand_chacha and I manually seed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,6 +235,7 @@ dependencies = [
  "petgraph",
  "pico-args",
  "rand",
+ "rand_chacha",
  "regex",
  "regex-syntax",
  "sha3",
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -23,7 +23,9 @@ doctest = false
 ascii-canvas = { version = "4.0", default-features = false }
 bit-set = { version = "0.8", default-features = false }
 ena = { version = "0.14", default-features = false }
-itertools = { version = "0.13", default-features = false, features = ["use_std"] }
+itertools = { version = "0.13", default-features = false, features = [
+    "use_std",
+] }
 petgraph = { version = "0.6", default-features = false }
 regex = { workspace = true }
 regex-syntax = { workspace = true }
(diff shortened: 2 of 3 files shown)


dburgener (approved):

dburgener:
Oh, I meant to ask what you thought about bumping up the ITERATIONS variable in lr1/build/test.rs in concert with this.  Since making this deterministic means that we test the same thing every time, it might be nice to test a few more things.

I'm not sure at all what the time cost for this test is though.  I wouldn't want to overdo it, but if it's efficient enough, we could do like 48 or 96 or something instead of the 22 it has now (no idea how that number was chosen, last commit on the line bumps it up from 10 to 22 without explanation)

Pat-Lafon:
> I'm not sure at all what the time cost for this test is though. I wouldn't want to overdo it, but if it's efficient enough, we could do like 48 or 96 or something instead of the 22 it has now (no idea how that number was chosen, last commit on the line bumps it up from 10 to 22 without explanation)

We can safely up this number to 96 without much of a noticeable timing regression.

This turned out to be a useful question because I realized that when I made the seed deterministic, I needed to pull it outside of the iteration loop to avoid just generating the same random tree over and over. This was noticeable because I was setting rather large iteration values with little timing change. Now with this fix, large iteration values like `1000` add noticeably to the run time(about an extra 2 seconds).

dburgener (approved):

Issue #992, opened 2024-11-18
Error Recovery not working when encountering a lexical error with Logos


chanbengz:
Trying to build a compiler with lalrpop and Logos and play with error recovery following [error recovery](https://lalrpop.github.io/lalrpop/tutorial/008_error_recovery.html) and [external lib](https://lalrpop.github.io/lalrpop/lexer_tutorial/005_external_lib.html)

```rust
use logos::{Logos, FilterResult};
// ...

#[derive(Default, Debug, Clone, PartialEq)]
pub enum LexicalError {
    InvalidInteger(String),
    InvalidCharacter(String),
    InvalidString(String),
    UnexpectedEndOfProgram,
    NonAsciiCharacter,
    #[default]
    UnknownToken
}

#[derive(Debug, Logos, PartialEq, Clone)]
#[logos(error = LexicalError)]
#[logos(skip r"[ \t\r\n\f]+")]
pub enum Token {
    #[end]
    EndOfProgram,

    // Operators
    #[token(">")]
    OpGreaterThan,
    // ...
}
```

```rust
use lexer::tokens::{Token, LexicalError};
use ast::tree;
use lalrpop_util::ErrorRecovery;

grammar<'err>(errors: &'err mut Vec<ErrorRecovery<usize, Token, LexicalError>>);

extern {
  type Location = usize;
  type Error = LexicalError;

  enum Token {
    "identifier" => Token::Identifier(<String>),
  // ...
  }
}

pub CompExpr: Box<tree::CompExpr> = {
    #[precedence(level="1")]
    Term,
   // ...
}

Term: Box<tree::CompExpr> = {
    <n: "int"> => Box::new(tree::CompExpr::Value(tree::Value::Integer(n))),
    "(" <CompExpr> ")",
    ! => { errors.push(<>); Box::new(tree::CompExpr::Error) },
}
```

And test with 
```rust
let mut errors = Vec::new();
let lexer = lexer::lexer::Lexer::new("1 + $"); // Invalid token '$'
let parser = CompExprParser::new();
assert_eq!(format!("{}", parser.parse(&mut errors, lexer).unwrap()), "1 + Error");
```

however it returns
```
---- tests::test_error_recovery stdout ----
thread 'tests::test_error_recovery' panicked at src/parser/src/lib.rs:86:38:
called `Result::unwrap()` on an `Err` value: User { error: UnknownToken }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
[...]

Issue #707, opened 2023-02-18
Don't require lexer and parser errors to be the same type


marcospb19:
Fallible actions allow for returning a custom error.

```rust
Num: i32 = {
    r"[0-9]+" =>? i32::from_str(<>)
        .map_err(|_| ParseError::User {
            error: YOUR_CUSTOM_ERROR_HERE
        })
};
```

However, if you use a custom lexer, LALRPOP requires that the error returned by your Lexer is the same as the one returned by your fallible actions.

I believe this requirement is unnecessary.

Fixing this would probably require re-designing https://docs.rs/lalrpop-util/0.19.8/lalrpop_util/enum.ParseError.html.

Pat-Lafon:
I'm kinda curious about this, do you have a full example? I'm wondering why this wouldn't be supported with something like `impl From<LexError> for CustomParseError`?

marcospb19:
I don't have a simple example but [this is the code](https://github.com/marcospb19/mini-c/blob/6c371fceb36a81a03d1a0f2b44d89713db8cde29/components/mini-c-lexer/src/lib.rs#L10-L11) I was messing with.

My _Parser_ crate depends on my _Lexer_ crate, so the _Lexer_ crate can't access `CustomParseError` to implement `impl From<LexError> for CustomParseError` and call it inside of its `impl Iterator`.

Sooo, to make this work, my parser error is actually declared in my lexer's lib.rs.

But even with this, `impl From` forces you to wrap your error type in some enum variant, making it a little more verbose in your `.lalrpop` file.

So yeah, ignoring the multi-crate issue, `impl From` in user code should work but it's just a small inconvenience.

---

I also think it would be cool if the error enum had a variant to tell us where the error was found by the parser.

```
pub enum ParseError<L, T, E> {
    ...
    UserParser {
        error: E,
    },
    UserLexer {
        error: E,
    },
}
```

Perhaps:

```
pub enum ParseError<L, T, PE, LE> {
    ...
    UserParser {
        error: PE,
    },
    UserLexer {
        error: LE,
    },
}
```

Pat-Lafon:
Maybe your code you linked is private? I couldn't follow the link.

marcospb19:
@Pat-Lafon  :O  oops, can you try again?

_(this comment will self-destruct soon)_

Pat-Lafon:
Ah, thanks for sharing. I'm a bit shaky on this usage of lalrpop but it's interesting to poke around.

[...]

PR #996, opened 2024-11-21
Document remaining undocumented lalrpop-util items

dburgener:
Enable `warn(missing_docs)` for lalrpop-util


lalrpop-util/src/lib.rs
```diff
@@ -108,14 +118,32 @@ impl<L, T, E> ParseError<L, T, E> {
         }
     }
 
+    /// Transform a ParseError by applying a function to the location field
```
Pat-Lafon, line 121: Should `ParseError` have backticks here as well?

lalrpop-util/src/lib.rs
```diff
@@ -48,11 +49,14 @@ pub mod state_machine;
 pub enum ParseError<L, T, E> {
     /// Generated by the parser when it encounters a token (or EOF) it did not
     /// expect.
-    InvalidToken { location: L },
+    InvalidToken {
+        /// The end of the invalid token
```
Pat-Lafon, line 53: Consistent period usage given that you added a period in the other case?

lalrpop-util/src/lib.rs
```diff
@@ -108,14 +118,32 @@ impl<L, T, E> ParseError<L, T, E> {
...
+    ///
+    /// This could be useful for example to obtain a `ParseError` where all
+    /// fields implement `Display` or `Debug`.  `ParseErrors` where all fields
+    /// implement `Display` also implement `Display` and `ParseErrors` where all
+    /// fields implement both `Display` and `Debug` implement `Error`
     pub fn map_location<LL>(self, op: impl FnMut(L) -> LL) -> ParseError<LL, T, E> {
         self.map_intern(op, |x| x, |x| x)
     }
 
+    /// Transform a ParseError by applying a function to the token field
```
Pat-Lafon, line 131: For the enum, each sentence ended with a period, maybe the first line here should as well for consistency

lalrpop-util/src/lib.rs
```diff
@@ -108,14 +118,32 @@ impl<L, T, E> ParseError<L, T, E> {
...
     pub fn map_token<TT>(self, op: impl FnOnce(T) -> TT) -> ParseError<L, TT, E> {
         self.map_intern(|x| x, op, |x| x)
     }
 
+    /// Transform a ParseError by applying a function to the error field
+    ///
+    /// This could be useful for example to obtain a `ParseError` where all
+    /// fields implement `Display` or `Debug`.  `ParseErrors` where all fields
+    /// implement `Display` also implement `Display` and `ParseErrors` where all
+    /// fields implement both `Display` and `Debug` implement `Error`
```
Pat-Lafon, line 146: I'm not sure that this example is clear. Basically what we are providing is https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err but for Lalrpop error construct.

Presumably these functions are more general. Like applying an offset value to a location or mapping a token character to it's escaped version.

lalrpop-util/src/lib.rs
```diff
@@ -108,14 +118,32 @@ impl<L, T, E> ParseError<L, T, E> {
         }
     }
 
+    /// Transform a ParseError by applying a function to the location field
+    ///
+    /// This could be useful for example to obtain a `ParseError` where all
+    /// fields implement `Display` or `Debug`.  `ParseErrors` where all fields
+    /// implement `Display` also implement `Display` and `ParseErrors` where all
+    /// fields implement both `Display` and `Debug` implement `Error`
     pub fn map_location<LL>(self, op: impl FnMut(L) -> LL) -> ParseError<LL, T, E> {
```
Pat-Lafon, line 127: One thing I find interesting is that location takes a FnMut compared to the other two functions(Because it is applied to both the start and ending locations).

Pat-Lafon (changes requested):
Don't want to bikeshed documentation too much but I think there are some quick changes here.

lalrpop-util/src/lib.rs
dburgener: I reworded these to try to make things more clear.  What do you think of the new version?

lalrpop-util/src/lib.rs
dburgener: Yeah, interesting.  I added a comment on the maptok declaration above, as well as a note here, since that's not immediate obvious when looking at the code.  I'm a little torn on what level of detail the user-facing docs need.  On the one hand I can imagine user confusion over the signature differences, but on the other hand, the reason it needs to be `FnMut` instead of `FnOnce` is arguable more of an implementation detail.

What do you think of the way it is now?

lalrpop-util/src/lib.rs
```diff
@@ -48,11 +49,14 @@ pub mod state_machine;
 pub enum ParseError<L, T, E> {
     /// Generated by the parser when it encounters a token (or EOF) it did not
```
dburgener, line 50: Just noticed while looking at #992.  This case is generated by the *lexer*, not the parser (which is the difference between this and UnrecognizedToken).  Let's get this and the UnrecognizedToken cleaned up as part of this PR.  I'll push a fix in a bit.

lalrpop-util/src/lib.rs
Pat-Lafon: Simpler but more clear in my opinion.

lalrpop-util/src/lib.rs
Pat-Lafon: This sounds related to https://github.com/lalrpop/lalrpop/issues/707. Maybe the built-in lexer errors should be better decoupled from the parser such that users and lalrpop can leverage the same lexer error interface.

Pat-Lafon (approved):
Thanks for fixing this up!

lalrpop-util/src/lib.rs
dburgener: Yeah, maybe?  I think my main thought on #707 is that I'm not really sure how ParseError::user doesn't get the job done?  But a LexerError variant would probably make sense.  I guess the question is what is the type of that?  A new generic in ParseError (as mentioned in #707)?  That's already a complicated signature (which has an error generic).  I'm also not clear on how that approach works for users of the built in lexer?

If instead we define LexerError as an enum ourselves, then we have to have a custom field in there, since we can't enumerate all user error types, so it's extra indirection.

I'm definitely in favor of improvements to ParseError, but neither of those paths sound initially promising to me...

I also wondered about extending lexer error coverage for #992 to let lexers return recoverable and non-recoverable errors, but I don't really see the benefit of a lexer returning `Err(RecoverableError(_))` rather than `Ok(Token::Error)` like in the workaround code.

PR #997, opened 2024-11-21
Convert map_or() to is_some_and()

dburgener:
Recommended by clippy 1.84.  This is slightly more concise and readable, since it omits the extra argument.


Pat-Lafon (approved):
Clean!

PR #998, opened 2024-11-21
Isolate allow(clippy::never_loop) to the part of the code that needs it

dburgener:
Recursive ascent parsers sometimes generate loops that return once. Preventing this would needlessly complicate the code generation logic. Previously, we blanket-allowed never_loop because our recursive_ascent tests triggered clippy warnings on that.  Limit this allow to only the part of the code that actually needs it.

This will enable users to see this clippy lint on their action code where applicable.

### DIFF
--- a/lalrpop/src/lr1/codegen/ascent.rs
+++ b/lalrpop/src/lr1/codegen/ascent.rs
@@ -452,6 +452,8 @@ impl<'ascent, 'grammar, W: Write>
 
         // finally, emit gotos (if relevant)
         if fallthrough && !this_state.gotos.is_empty() {
+            // Sometimes we write loops that unconditionally only loop once
+            rust!(self.out, "#[allow(clippy::never_loop)]");
             rust!(self.out, "loop {{");
 
             // In most states, we know precisely when the top stack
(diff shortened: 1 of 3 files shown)


Pat-Lafon (approved):
Less blanket clippy lints is great. I guess this case doesn't show up in the lalrpop grammar?

dburgener:
> I guess this case doesn't show up in the lalrpop grammar?

This case only shows up in recursive ascent codegen.  (Which is part of what makes this a nice win, since the blanket allows apply to both parse tables and recursive ascent.  Now it's not only isolated to just parts of the code, it's also isolated to not apply to parse tables at all).  The bootstrapped lrgrammar.rs is parse table generated, so it doesn't have the code clippy alerts on.

Somewhat relatedly, something I didn't know before is that generally our lalrpop-test examples are being built in TestAll configuration to test both recursive ascent generation and parse tables: https://github.com/lalrpop/lalrpop/blob/3d93f2db8c34799d3998368fc7a10e8d3dd9a79d/lalrpop/src/normalize/lower/mod.rs#L175.  (build.rs in lalrpop-test sets the `unit_test()` configuration).  So we see a lot of clippy hits in lalrpop-test for this lint, but nothing in lrgrammar.rs.

Pat-Lafon:
Ah interesting, I wasn't aware that we enabled testall like this but that is good to know.

PR #999, opened 2024-11-22
Remove blanket allow(clippy::needless_raw_string_hashes)

dburgener:
This lint was moved by clippy into the pedantic group, so it's allowed by default anyways.

For the sake of users who may warn on it anyways, include it locally on the list of terminals where it's actually relevant.

In our own code base, just drop it, since the lint is allowed.


Pat-Lafon (approved):
LGTM

PR #1002, opened 2024-12-02
Pin lexer logos dependency to 0.14.2 for now

dburgener:
0.14.3 bumps msrv to 1.74, so our CI test for 1.70 build is currently
  failing.

This can be removed once our MSRV moves to 1.74 or later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -275,6 +275,8 @@ dependencies = [
  "lalrpop",
  "lalrpop-util",
  "logos",
+ "logos-codegen",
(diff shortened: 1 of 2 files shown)


Pat-Lafon (approved):
Kind of surprising/annoying that dependencies of logos also need to be pinned to have the desired effect.

Seems like future cargo resolvers will better support this but it doesn't currently help with our workflow of using an older rust: https://github.com/rust-lang/cargo/issues/13540

PR #1003, opened 2024-12-02
Changes/Fixes for the 2024 edition

Pat-Lafon:
Since I mentioned the 2024 edition separately, I wanted to go through and see what changes would need to be made. There are two sides of this, some changes need to be made but are totally compatible with the current edition(replacing usage of `gen` as a variable since it is becoming a keyword) versus changes to be made that are backwards incompatible. There are also some updates that the edition fixer suggests that are too conservative. I'll try to create a list as I go.

- https://doc.rust-lang.org/nightly/edition-guide/rust-2024/gen-keyword.html The gen keyword will not be a valid variable name going forward
- https://doc.rust-lang.org/nightly/edition-guide/rust-2024/macro-fragment-specifiers.html For compatibility/conservativeness, the edition fixer replaces every usage of `expr` in a macro_rules macro with `expr_2021`. For our usage, this is unnecessary as we never try to separately match on const(and most of our macro_rules macros are simple anyways). See https://github.com/lalrpop/lalrpop/blob/35f1346afdb5b5b512d39d775f448ef1763e43b3/lalrpop/src/normalize/mod.rs#L17-L24 https://github.com/lalrpop/lalrpop/blob/35f1346afdb5b5b512d39d775f448ef1763e43b3/lalrpop/src/parser/mod.rs#L33-L45 as examples of our usage.
- https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-if-let-scope.html There has been a change to if-let and how long it keeps variables alive. Unfortunately, this is a very noise lint https://github.com/rust-lang/rust/issues/133167 linting in places like https://github.com/lalrpop/lalrpop/blob/35f1346afdb5b5b512d39d775f448ef1763e43b3/lalrpop/src/build/mod.rs#L382-L384 where there doesn't seem to be an issue(as there is no else in this case and I don't think the drop implementation has a lurking issue). In general, running `cargo clippy --fix --allow-dirty` cleaned some of these up. There are also cases with drop implementations like `RefCell` I think that have drop impls that cause the automation to insert matches. I don't believe any of this is necessary(see issue for others suggesting the conversion is very imprecise).
- https://doc.rust-lang.org/nightly/edition-guide/rust-2024/match-ergonomics.html Occurs here https://github.com/lalrpop/lalrpop/blob/35f1346afdb5b5b512d39d775f448ef1763e43b3/lalrpop/src/lr1/report/mod.rs#L47-L51 As apparently the match "resets the default binding mode" and needs the ref removed `Ok(states) => {`. 
- The lexer crate has issues from the attempted automated fix:
```sh
The following errors were reported:
error: expected one of `(`, `,`, `::`, or `=`, found `+`
  --> doc/lexer/src/tokens.rs:18:16
   |
18 | #[derive(Logos + use<'s>, Clone, Debug, PartialEq)]
   |                ^ expected one of `(`, `,`, `::`, or `=`
``` 
Not sure what is going on here `cargo +nightly check` seems to work... Hopefully something the logos crate will figure out?
- Updating to the new edition also requires updating the msrv(somewhat obviously), and I updated the resolver to `3`
- There are a large number of formatting changes


dburgener:
Thanks for putting this together!

I haven't looked at anything except the description yet, but regarding the Logos derive thing at the end, that sounds awfully similar to what I was reading about on their repo this afternoon looking at their msrv bump:

https://github.com/maciejhirsz/logos/issues/434
https://github.com/maciejhirsz/logos/pull/435

It sounds to me like they support this syntax on their most recent 0.14.3 release IF you have rust 1.82 or later.  That's probably the issue here?

(The msrv bump was just because they found a bug in their CI while working on that, and discovered that they actually needed a 1.74 msrv in some configurations)

dburgener:
Okay, took a somewhat more thorough look over this.  Thanks for the writeup and draft PR!

Where do you want to go with this in the short-term?  Are you intending to keep this PR semi-up-to-date until we're ready to migrate editions?

It looks to me like the majority of this is backwards-incompatible.  I don't see a reason to not take at least the `gen` change right now though.  Nothing else looks to me like something we both can and want to take in the short term (the match ergonomics and the reformat seem nice, but something we'll have to wait to officially migrate for).

Pat-Lafon:
> Where do you want to go with this in the short-term? Are you intending to keep this PR semi-up-to-date until we're ready to migrate editions?

Happy to keep this up to date/rebased on master. I don't expect too many changes tbf. Part of this is to have this ready for whenever lalrpop is ready and also to see what issues we might have run into.

A lot of this obviously still requires nightly to run, to limit changes, I'll wait to the edition hits stable to get the ci working. Then it can be merged when appropriate as we are discussing with the msrv.

> It looks to me like the majority of this is backwards-incompatible. I don't see a reason to not take at least the `gen` change right now though. Nothing else looks to me like something we both can and want to take in the short term (the match ergonomics and the reformat seem nice, but something we'll have to wait to officially migrate for).

yup

Pat-Lafon:
The Rust 2024 edition is out so I cleaned up a few things. This does bump the msrv to 1.85. I notice that `publish-dry-run` is failing... I'm pretty sure it's because we don't set up our rust version with a
```
      - uses: dtolnay/rust-toolchain@master
        with:
          toolchain: stable
```
and just rely on the github worker's cargo which hasn't been updated yet? Not 100% sure.

dburgener:
> and just rely on the github worker's cargo which hasn't been updated yet? Not 100% sure.

I imagine you're right.  Are you planning to submit a PR for that, or should I?

lalrpop-util/src/state_machine.rs
```diff
@@ -649,19 +649,11 @@ macro_rules! integral_indices {
     ($t:ty) => {
         impl<D: ParserDefinition<StateIndex = $t, ReduceIndex = $t>> ParserAction<D> for $t {
             fn as_shift(self) -> Option<D::StateIndex> {
-                if self > 0 {
-                    Some(self - 1)
-                } else {
-                    None
-                }
+                if self > 0 { Some(self - 1) } else { None }
```
dburgener, line 652: I find this reformat less readable, personally.  I think it's caused by [this](https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-formatting-fixes.html#format-the-last-expression-statement-as-an-expression) 2024 edition style guide change.  The change seems to be good in the abstract, but for an expression this complicated, it's a lot, IMO.

I found this configuration option that seems like it would possibly affect this: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#single_line_if_else_max_width.  It looks like the default max if/else width is 50 characters, and this is 44 characters, so it's just under the limit.  I wonder if dropping this setting to 40 for our repo would improve the code readability?  I haven't tried it, so I don't know if it would work or if it would have any other side effects.

dburgener: I just got to the doc/whitespace line in a later commit with `if negative { -value } else { value }` doing the same change.  That one seems very reasonable to me and is only 37 characters, so that reinforces 40 as a decent limit.

I do think that length is a mediocre proxy for the real issue, which is the complexity.  In the doc/whitespace example, the condition, if and else blocks are all just a value (possibly negated).  Here, both the condition and the if block are more complex expressions, and it's extra effort to parse it all out on one line.

dburgener:
This all seems fairly straightforward.  Everything looks good to me, except the one comment I noted.  I assume we're still planning to wait until 1.87 is stable per our N-2 MSRV policy.

lalrpop-util/src/state_machine.rs
Pat-Lafon: Hmm, I added a configuration file for rust format to set this to 40. It changed this back, though it did modify two other files. I'm a little ambivalent as long as we choose a consistent number.

Pat-Lafon:
> This all seems fairly straightforward. Everything looks good to me, except the one comment I noted. I assume we're still planning to wait until 1.87 is stable per our N-2 MSRV policy.

Sounds good to me, I just rebased this to address recent changes to the workflow file.

lalrpop-util/src/state_machine.rs
dburgener: Thanks for trying it...

I don't know... All the other examples it changed do feel subjectively worse to me...  I think length just doesn't really capture my intuition about readability at all for these.

I think seeing the full impact, my personal preference would be to drop that commit and leave it at the default of 50.  I don't think the readability downgrade for the other lines is worth the improvement in the ones I originally mentioned.

Pat-Lafon: Reverted

dburgener:
Just noting that 1.87 became stable last week.  So we should be good to do the MSRV bump for this now.

Pat-Lafon:
@dburgener Just want to check about `set_var` becoming unsafe in the new edition.

It is currently used to help in testing the different output directory behavior https://github.com/lalrpop/lalrpop/blob/dd942b1fa4aa99c123bc984cf723347e0a26926f/lalrpop/src/api/test.rs#L47 . `set_var` is not theadsafe on most platforms(except windows apparently) so it will be marked as unsafe going forward. https://doc.rust-lang.org/std/env/fn.set_var.html 

On one hand, lalrpop only uses this for testing so this shouldn't affect actual users. On the other hand, `cargo test` runs multi-threaded by default so our usage is probably not safe even with the mutex(It's actually not super clear to me that asking cargo test to run with only one thread would actually help https://github.com/rust-lang/rust/issues/104053).

I have a couple of ideas for fixes(with varying tradeoffs) that I'll enumerate:
- Only run these tests on windows via ci. The lowest effort safe solution but also makes working with those tests difficult.
- Use `cargo-nextest` instead of `cargo test`. They use a test-per-process approach which they argue makes set_var usage safe https://nexte.st/docs/configuration/env-vars/#altering-the-environment-within-tests
- Change the tests to provide the output directory by modifying `Configuration::new()` and not going through the `process_root`, `process_src` wrapper. Maybe a little sad that we aren't testing the same interface that users leverage, but atleast we are testing the actual functionality safely.
- Just wrap this call in `unsafe`, it's guarded by a mutex and hasn't caused us issues so far? I think users won't get affected by test unsafety, though I'm not sure of all the ways this could back-fire. Possibly sporadic/unexpected test failures?
- Maybe there are other crates that can be used for testing these couple of tests in isolation like https://github.com/oknozor/sealed-test . Didn't look too far into this to see if it's confirmed they avoid this unsafety.

Pat-Lafon:
I'm going to mention that `set_var` is also referenced in two other places. Once in the Changelog for Version 0.13(which I don't think people look back that far) and likewise in https://github.com/lalrpop/lalrpop/blob/master/doc/src/advanced_setup.md#using-the-legacy-lalr-parser which the Changelog is referencing. I'm not sure how much of an issue it is? Running `set_var` in build.rs is probably fine. Also there is #990 which suggests we are moving away from the legacy parser anyways.

dburgener:
Ouch, good catch on the `set_var` issue.  That's frustrating.  I spent some time reading up on it over the weekend, and it's a pretty nasty thread-unsafety issue...

Running through the options you enumerated.

* I really dislike the windows-only option.  I really want to be able to run as many tests locally as possible.
* I don't think setting it with `Configuration::new()` actually tests what we want to test.  The whole point is to test the interface users are using.  If we set `out_dir` manually, we're testing that part of the interface and ignoring the part that reads the environmental variable.  The "actually do something with it all" is supposed to be as minimal as possible.  We just want to know that it got set up correctly for these tests.

Those are the two I would strongly prefer to avoid off the bat.  The rest all sound viable to me.

* Just wrapping it in unsafe and running is probably okay?  My understanding of the issue is that `set_var` is thread-safe in rust code only. The issue is with libc or other C callouts.  In the C case, it's pretty nasty, since it's unsafe with `get_var` too, not just `set_var`.  We do link libc transitively in a few places, and the mutex guards we have only guard the API tests specifically, so they could in theory run simultaneously with some other test relying on libc.  We haven't seen any issues in practice yet, but we also haven't been running these new tests very long.
* `cargo-nextest` is really intriguing.  Their docs promise a lot of benefits besides just solving this issue (e.g. performance, handling the mutex case in the api tests natively).  The only immediate downsides I see there are 1. Nonstandard, extra effort for new contributors.  2. Minor risk that it someday becomes unmaintained (looks like a large, mature project, so low risk there) 3. retraining fingers to type `cargo nextest run` instead of `cargo test`.  Have you tried nextest at all?  I feel like both of us giving it a spin and playing with it seems like a good next step there.
* Sealed test is interesting.  Seems like a cool idea overall.  I have concerns about stablity though.  Only 2 contributors with 19 total commits and last commit 11 months ago.  May be a dead project.

In terms of the legacy parser - the decision to use `set_var` is really on anyone wanting to do that, and they can always just set the variable from the command line too.  But even if we're not pulling it out entirely yet, I don't think we want to support it.  Lane table is strictly better.

Thoughts on that overall approach?

Pat-Lafon:
This seems a little thorny. I was thinking about cargo-nextest and it seems to be almost too-good of a drop in replacement?

As in, what is the material difference between 1 and 2? In both I'll go wrap `set_var` in unsafe, I'll add a comment why there is unsafe code here(in 1 saying that we think this is fine because of the mutex guards plus the only code being run is what we use for testing so there is a limited confidence that nothing else is trying to access the environment variables, in 2 saying that tests should only be run per-process via cargo-nextest to ensure safety). In 1 we at this point we stop, in 2 I add cargo-nextest to the ci and add a note to the readme for contributors that we use cargo-nextest for testing?

Is there a code change such that `cargo test` doesn't work but `cargo nextest` does? Otherwise presumably people will accidentally use cargo test out of habit and we are basically at 1(Unless we remove the mutex guard because in 2 the per-process isolation makes this unneeded, then someone running `cargo test` has a reasonable chance of a race condition I imagine). There might still be good reasons to use cargo-nextest in ci anyways for what you mentioned.

dburgener:
> As in, what is the material difference between 1 and 2? In both I'll go wrap set_var in unsafe, I'll add a comment why there is unsafe code here(in 1 saying that we think this is fine because of the mutex guards plus the only code being run is what we use for testing so there is a limited confidence that nothing else is trying to access the environment variables, in 2 saying that tests should only be run per-process via cargo-nextest to ensure safety). In 1 we at this point we stop, in 2 I add cargo-nextest to the ci and add a note to the readme for contributors that we use cargo-nextest for testing?

Yes, that is my understanding as well.  We need `unsafe` either way (assuming we stick with setvar).  Switching to nextest makes that actually safe.  Staying on `cargo test` is theoretically thread unsafe.  We haven't observed crashes yet, but we've barely been running these tests so far.

> Is there a code change such that cargo test doesn't work but cargo nextest does? Otherwise presumably people will accidentally use cargo test out of habit and we are basically at 1(Unless we remove the mutex guard because in 2 the per-process isolation makes this unneeded, then someone running cargo test has a reasonable chance of a race condition I imagine). There might still be good reasons to use cargo-nextest in ci anyways for what you mentioned.

Yes, removing the mutex guard I would expect `cargo test` to fail fairly consistently, but not technically deterministically.  If we want to enforce "no cargo test", we could rely on the fact that nextest sets various environmental variables (https://nexte.st/docs/configuration/env-vars/) and write a test that just asserts that the `NEXTEST` variable is set to `1` or similar, with a panic message directing users to use nextest.

I think that if we want to mandate nextest, that's probably a good way to do it.  My slight leaning at the moment might be for now to recommend, but not require nextest.  We haven't ran these tests a lot, but we also haven't seen the set_var race hit in practice.  So for a casual contributor maybe accepting the race chance rather than needing to download nextest makes sense?  We could start with a recommendation approach, but if we end up relying on nextest more and more and don't want to support `cargo test` we could switch to mandating.

If we do merge #1069 all that does is make nextest possible to use, so that's a fairly non-committal step by itself in terms of this PR.  We could go ahead and merge that, and take some time with both options to test drive nextest more thoroughly before committing to a project policy.

Pat-Lafon:
At a glance, the lint phase is failing on #1075. If there are other failing lints that are uncovered after merging #1076 I'll go fix them

dburgener (approved):
LGTM!

I say let's hit it.

PR #1005, opened 2024-12-04
Drop extra space in generated code

dburgener:
The visibility Display impl will include a space (if needed - it may be the empty string).  Having an extra space in the trait declaration resulted in two spaces (or a line starting with a single space if the visibility displayed the empty string).


Pat-Lafon (approved):
Ah, I was wondering why remove the space here and not in the display impl but it the visibility is empty, then you don't want an unnecessary space here... cool

PR #1006, opened 2024-12-04
Propose a new release policy

dburgener:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,6 +56,39 @@ So this approach has been abandoned a while ago.
 
 ### Releasing LALRPOP
 
+#### Semver policy
+
+All releases should follow rust's semantic versioning guidelines, as described
+at https://doc.rust-lang.org/cargo/reference/semver.html. Breaking changes in
+our context include changes that break previously working lalrpop grammars.
+
+If a PR introduces a breaking change, the breaking-ness should not delay the
+merge of the PR.  Instead, master will become breaking and the next release
+from master will be a breaking change release.  However, to support continued
+non-breaking releases and reduce the frequency of breaking releases, we will
+create a branch off of the last commit before the breaking change and backport
+future non-breaking changes to it.  For example, if the latest release was 0.22
+and a breaking change is merged, we will create a branch named "0.22.x" from
+the commit prior to the breaking change.  We will backport to that branch as
+non-breaking changes are merged, and possible release 0.22.1, 0.22.2 etc from
+it.
+
+Except in rare and severe circumstances, the 0.22.x branch will only be
+maintained until the release of version 0.23.  A critical bug may warrant a
+later backport to the 0.22.x series, but in general, we will stop backporting
+and focus efforts on the 0.23 series as soon as 0.23 is released.
+
+Regarding msrv bumps, the guidance in the rust community is that msrv bumps are
+not necessarily breaking, but should be bundled with breaking changes if
+possible. Our msrv strategy follows this guidance, attempting to bundle msrv
+bumps with breaking releases.  We prefer to keep the msrv at least two versions
+behind stable (eg if the latest rust release is 1.83, we prefer our msrv to be
+no more than 1.81).  In the event that there is a prolonged time between
+breaking releases, it may be desirable to bump the msrv in a non-breaking
+release.  In this case, we prefer to keep at least eight releases behind stable
+(eg if the latest rust release is 1.83, we prefer to bump the msrv in a
+non-breaking release to at most 1.75).
+
 #### Prerequisites
 1. `version.sh` makes use of clog-cli.  Run `cargo install clog-cli` to get it
 


dburgener:
@Pat-Lafon and I had been discussing breaking change and msrv philosophy a bit on #1001, and I thought it warranted it's own forum for discussion.  I wrote up my proposal for how we should handle breaking changes and msrv bumps and opened this draft PR for discussion.  @lalrpop/maintainers what are your thoughts on the strategy documented in this PR?

CONTRIBUTING.md
```diff
@@ -56,6 +56,39 @@ So this approach has been abandoned a while ago.
...
+Regarding msrv bumps, the guidance in the rust community is that msrv bumps are
+not necessarily breaking, but should be bundled with breaking changes if
+possible. Our msrv strategy follows this guidance, attempting to bundle msrv
+bumps with breaking releases.  We prefer to keep the msrv at least two versions
+behind stable (eg if the latest rust release is 1.83, we prefer our msrv to be
+no more than 1.81).  In the event that there is a prolonged time between
+breaking releases, it may be desirable to bump the msrv in a non-breaking
+release.  In this case, we prefer to keep at least eight releases behind stable
+(eg if the latest rust release is 1.83, we prefer to bump the msrv in a
+non-breaking release to at most 1.75).
```
Pat-Lafon, line 90: Thinking a little bit about this part. Are the main reasons here because lalrpop want's to keep up to date with Rust features/best-practices? Another reason is what happened to us with logos, where a dependency pushed up our MSRV from under us(probably fixed in the new edition).

To clarify on eight releases, is this a magic number or approximating 1 years worth of releases(for what it's worth, the `regex` crate has msrv bumps after a year)?

Pat-Lafon:
This is well written!

CONTRIBUTING.md
dburgener: > Are the main reasons here because lalrpop want's to keep up to date with Rust features/best-practices? Another reason is what happened to us with logos, where a dependency pushed up our MSRV from under us(probably fixed in the new edition).

Pretty much, although I'll add two things:

* In addition to an existing dependency bumping the msrv, we may want to take on a new dependency with a more recent msrv (eg #981)
* The farther out of date we get, the more we have to pay attention to the version numbers in the official docs to avoid using functions that would bump the msrv.  When we do bump the msrv, we may want to remember to go forward and switch to the more modern functions where appropriate.

> To clarify on eight releases, is this a magic number or approximating 1 years worth of releases

Yes, magic number, existing basically as a strawman for discussion as much as as a real proposal.  Nine releases would approximate 1 year slightly better, but 8 is a power of 2, so I went with 8.  7+ is what would be needed to get us to 1.74 right now, which is desirable for a few open PRs.

> (for what it's worth, the regex crate has msrv bumps after a year)?

Interesting, I didn't know that.  It seems like "a year old" is probably about as old as we can reasonably hang out based on that, so reinforcement for the 8 releases threshold?

dburgener:
If we adopt this, we should probably also include the msrv policy somewhere more user-visible, like the lalrpop book or our docs.rs page so that users know what to expect about msrv bumps, since this is an explicitly contributor focused doc here.

CONTRIBUTING.md
Pat-Lafon: One thing that might affect our policy here is that there is a new msrv-aware solver in Rust 1.84. That is a few versions ahead of #1001 so it's not helpful yet, but it should help with reducing the number of breakages from package updates like in logos or pulling in new dependencies that have increased their msrv for new features and not api changes.

dburgener:
@lalrpop/maintainers Any more thoughts on this?

I'd kind of like to do a new point release (0.22.2) to grab #1047, but my preference would be to not bump the MSRV in a point release right now, since we do want to do a breaking release in a few months where we can bump the MSRV.  In order to do that, I'd like to execute the plan written here to create a 0.22.x branch off of eaecb9ade176e20e8af4c26438f4fdb3f2592d00, backport just #1047 (once it's merged to master of course), and then release 0.22.2 off of that.

If I'm going to do that, I'd love to have some explicit sign-off on the strategy, including merging this doc update :).

CONTRIBUTING.md
Pat-Lafon: That msrv-aware solver is now the default in the new edition

CONTRIBUTING.md
```diff
@@ -56,6 +56,39 @@ So this approach has been abandoned a while ago.
 
 ### Releasing LALRPOP
 
+#### Semver policy
+
+All releases should follow rust's semantic versioning guidelines, as described
+at https://doc.rust-lang.org/cargo/reference/semver.html. Breaking changes in
+our context include changes that break previously working lalrpop grammars.
```
Pat-Lafon, line 63: And breaking changes(though rare as they are) to the Configuration api used in build scripts?

CONTRIBUTING.md
```diff
@@ -56,6 +56,39 @@ So this approach has been abandoned a while ago.
...
+future non-breaking changes to it.  For example, if the latest release was 0.22
+and a breaking change is merged, we will create a branch named "0.22.x" from
+the commit prior to the breaking change.  We will backport to that branch as
+non-breaking changes are merged, and possible release 0.22.1, 0.22.2 etc from
+it.
+
+Except in rare and severe circumstances, the 0.22.x branch will only be
+maintained until the release of version 0.23.  A critical bug may warrant a
+later backport to the 0.22.x series, but in general, we will stop backporting
+and focus efforts on the 0.23 series as soon as 0.23 is released.
```
Pat-Lafon, line 79: Just to clarify there is only 1 0.22.x branch being maintained and not a 0.22.1, 0.22.2, etc

Pat-Lafon (approved):
In general I'm fully on board with this. I'm not actually sure how often the "eight-releases behind stable" situation will occur. The edition change may be a big bump per-se that we might be waiting on and seeing how quickly it's adopted.

Pat-Lafon:
> I'd kind of like to do a new point release (0.22.2) to grab #1047, but my preference would be to not bump the MSRV in a point release right now, since we do want to do a breaking release in a few months where we can bump the MSRV. In order to do that, I'd like to execute the plan written here to create a 0.22.x branch off of [eaecb9a](https://github.com/lalrpop/lalrpop/commit/eaecb9ade176e20e8af4c26438f4fdb3f2592d00), backport just #1047 (once it's merged to master of course), and then release 0.22.2 off of that.

👍

CONTRIBUTING.md
```diff
@@ -56,6 +56,39 @@ So this approach has been abandoned a while ago.
...
+If a PR introduces a breaking change, the breaking-ness should not delay the
+merge of the PR.  Instead, master will become breaking and the next release
+from master will be a breaking change release.  However, to support continued
+non-breaking releases and reduce the frequency of breaking releases, we will
+create a branch off of the last commit before the breaking change and backport
+future non-breaking changes to it.  For example, if the latest release was 0.22
+and a breaking change is merged, we will create a branch named "0.22.x" from
+the commit prior to the breaking change.  We will backport to that branch as
+non-breaking changes are merged, and possible release 0.22.1, 0.22.2 etc from
+it.
```
yannham, line 74: Note that this is not always easy to do in practice, because following changes might need to be based on breaking changes, even if they appear unrelated for users. I don't know how much we can promise to do that; maybe just mentioning that it's best-effort could a be a vague but safe strategy.

CONTRIBUTING.md
```diff
@@ -56,6 +56,39 @@ So this approach has been abandoned a while ago.
...
+
+Except in rare and severe circumstances, the 0.22.x branch will only be
+maintained until the release of version 0.23.  A critical bug may warrant a
+later backport to the 0.22.x series, but in general, we will stop backporting
+and focus efforts on the 0.23 series as soon as 0.23 is released.
+
+Regarding msrv bumps, the guidance in the rust community is that msrv bumps are
```
yannham, line 81: Nitpick: MSRV is introduced for the first time and not capitalized, so its definition might not be obvious for all occasional readers. Maybe we should spell it out entirely the first time?

yannham (approved):
FWIW, I find this base approach sensible.

CONTRIBUTING.md
dburgener: Yeah, the msrv aware resolver should definitely help.  One downside is that it essentially does the workaround of pinning the old version for you without needing manual effort.  While that's nice, we do want to get the new version at some point, which could still motivate an msrv bump.

The other reason we may want to bump msrv is that we directly want new features/fixes in the toolchain.  The examples I'm aware of are the `core::Error` thing we carried a workaround for for a little while, and the bug that was fixed in 1.71 that hit #978.

So hopefully msrv aware resolver will make our lives easier as maintainers, but I think we'll still want some sort of escape hatch to bump msrv in point releases after a while, assuming we can manage to go an extended period of time without breaking changes at some point.

CONTRIBUTING.md
dburgener: Yeah, my intent was to assume those were breaking and spell out the additional case that isn't breaking, but I'll push a change to this to make that more explicit in a moment here.

CONTRIBUTING.md
dburgener: I agree.  I added an additional paragraph on that (will push in a minute).  Let me know what you think.

CONTRIBUTING.md
dburgener: Will push fix with this suggestion in a minute.

dburgener:
Thanks guys, I just pushed an update that I think addresses all the comments.  Let me know if you have any other suggestions.  If I don't hear anything I'll assume we're good from prior approvals and merge sometime this week, as well as aim to make the 0.22.x branch for a backport release of #1047 (and #1048).  I'll do that through the PR process, and update the github actions configuration to run CI against 0.22.x once the branch is created as well.

yannham (approved):
You can consider my review comments to be fixed by the latest version :+1:

PR #1007, opened 2024-12-04
Replace `gen` keyword with `generator` instead of `r#gen`

Pat-Lafon:


dburgener:
LGTM

Just for my awareness, when you say "instead of `r#gen`" in the subject line, is `r#gen` what `cargo fix --edition` does?

Pat-Lafon:
> Just for my awareness, when you say "instead of `r#gen`" in the subject line, is `r#gen` what `cargo fix --edition` does?

Yup, that is the automated fix

PR #1010, opened 2024-12-06
Add code coverage write-up to contributing doc

dburgener:
It took a little while to get this working, so I wanted to make sure it's documented for future me or anyone else who may attempt it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,3 +77,25 @@ If using [cargo-release](https://github.com/crate-ci/cargo-release):
 4. `cargo release --execute --package lalrpop --no-tag`
 
 Drop the `--execute` flag to do a dry run and check for any issues first.
+
+### Running code coverage tools
+
+There are a few gotchas when running code coverage tooling on the lalrpop code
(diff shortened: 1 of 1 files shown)


Pat-Lafon (approved):
LGTM. Out of curiosity, what is our current coverage? Do the doc examples also count?(Maybe they don't provide any unique coverage though)

dburgener:
After #1011 we're at 72% function coverage and 63.5% line coverage (compared to 71% and 63% before).  The massive caveat though is that around half our functions and over half our lines for the whole project are in the generated lrgrammar.rs, which has 52% function coverage and 47% line coverage.  If you look at our overall coverage after #1011 and exclude lrgrammar.rs, we're at 91% function coverage and 90% line coverage.

In terms of docs, I assume you mean the `Example:`  field in doc comments?  `cargo llvm-cov` docs say that they can support it, but I didn't try out that functionality, so these numbers don't include those.  According to the docs it requires nightly.  In terms of the `doc/` directory examples, as is presumably clear from the PR, it builds those and runs any tests.

dburgener:
> as is presumably clear from the PR

Forgot which PR I was in.  By "the PR" I meant #1011.

PR #1011, opened 2024-12-06
Add some extra tests for untested areas of the codebase.

dburgener:

### DIFF
--- a/lalrpop/src/lr1/report/test.rs
+++ b/lalrpop/src/lr1/report/test.rs
@@ -1,0 +1,40 @@
+use std::rc::Rc;
+
+use crate::file_text::FileText;
+use crate::lr1;
+use crate::lr1::tls::Lr1Tls;
+use crate::session::Session;
+use crate::test_util::normalized_grammar;
+use crate::tls::Tls;
+
+use super::*;
+
+const GRAMMAR_TEXT: &str = "grammar;
+pub A: () = {
+    B,
+    C
+}
+
+B: () = \"b\";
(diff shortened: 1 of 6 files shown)


lalrpop/src/lr1/report/test.rs
```diff
@@ -0,0 +1,40 @@
...
+    let grammar = normalized_grammar(GRAMMAR_TEXT);
+    let _tls = Tls::install(
+        Rc::new(Session::new()),
+        Rc::new(FileText::new("".into(), "".into())),
+    );
+    let _lr1_tls = Lr1Tls::install(grammar.terminals.clone());
+    let start_nt = grammar.start_nonterminals.keys().next().unwrap().clone();
+    generate_report(&mut output, &lr1::build_states(&grammar, start_nt)).unwrap();
+
+    let report = String::from_utf8(output).unwrap();
```
Pat-Lafon, line 34: Nit: Could output just have been a string to begin with instead of a Vec<u8> that gets converted to a string?

Pat-Lafon (approved):
More tests are always good!

lalrpop/src/lr1/report/test.rs
dburgener: It could not, because Strings don't implement `io::Write`, and `generate_report()`s output buffer must do so.

Pat-Lafon: Ah, because it needs to be valid UTF-8 for strings and io::Write lets you write bytes. Whoops

PR #1014, opened 2024-12-17
Unbreak documentation tests

dburgener:
The value proposition here is a little sketchy - it only runs one test as of right now.  It's in theory helpful for future doc test expansion, although the rust part of our public API is pretty small.  Testing lalrpop code snippets would be much higher value.  As far as I can tell, that would require us to write our own driver to browse source looking for ```` ```lalrpop ```` blocks and funneling them into lalrpop.


Pat-Lafon:
Since this was mentioned, could be worth pushing this through. I think it's worth setting up in case we want to add more doc tests in the future.

PR #1017, opened 2024-12-17
Don't run cargo update on msrv or feature-powerset tests

dburgener:
This reduces the risk of breakage when a dependency bumps msrv and is consistent with the guidance at https://doc.rust-lang.org/nightly/cargo/reference/rust-version.html


dburgener:
I'm going to go ahead and merge this without review to unbreak the CI for now.  We can always modify or revert later.

Issue #1013, opened 2024-12-15
LSP Integration Demo


LighghtEeloo:
Hi LALRPOPers,

I've been working on integrating an LSP server for LALRPOP using tower-lsp, and I’d love to get your feedback. In this [repository](https://github.com/LighghtEeloo/lalrpop-lsp), I’ve implemented the following features (albeit in a somewhat hacky manner):

- Go to Definition
- Find References
- Hover (showing type information)
- Error Reporting

I've also provided a demo [VSCode extension](https://marketplace.visualstudio.com/items?itemName=LitiaEeloo.lalrpop-language-server), which should be easy to adapt to other editors that support LSP.

### Current Challenges

The implementation, while functional, has a few limitations and potential pitfalls:

1. Forking LALRPOP: To reuse the existing LALRPOP parser and internal data structures, and to avoid making invasive changes to the existing codebase, I [forked LALRPOP](https://github.com/LighghtEeloo/lalrpop/tree/lsp) and put all new code in a separate module `pub mod lsp`, which is only enabled when the `lsp` feature flag is set.
    1. Error Display: To produce LSP-friendly errors (without text highlights and printed location information), I had to duplicate code for displaying error messages for now. A sight refactor to how errors are displayed in LALRPOP can greatly help avoiding the duplication.
    2. Symbol Collection: All symbols are collected by traversing `crate::grammar::parse_tree::Grammar`. I'm not sure if this is feasible if I use LALRPOP as a library instead.
2. Equational Theory for Names: Since LALRPOP rules are nominal, name comparisons currently rely on direct string matching, which may not be ideal.

### Missing Features

Two useful features are currently difficult to implement:

1. Go to Definition for Macro Parameters
2. Semantic Tokenization

Both features would rely on an enriched version of [lrgrammar.lalrpop](https://github.com/LighghtEeloo/lalrpop/blob/lsp/lalrpop/src/parser/lrgrammar.lalrpop) with more detailed span information.

### Request for Feedback

I’m reaching out to the maintainers and the community to seek guidance on the following:

- Is there a better way to integrate LSP functionality with LALRPOP to minimize maintenance burden? For example, would it be better if we refactor error reporting a little bit?
- How feasible is it to enhance `lrgrammar.lalrpop` with additional span information for improved IDE features?

Any feedback, advice, or pointers would be greatly appreciated!

dburgener:
[...]

PR #1021, opened 2025-01-05
Refactor `report_parse_error` in `crate::build`

LighghtEeloo:
As noted in #1013, this PR aims to abstract error handling. The only file modified in this PR is `src/build/mod.rs`, where parse error reporting has been refactored out of the function `parse_and_normalize_grammar` into a new public function, `report_parse_error`.

The changes are straightforward: the new `report_parse_error` function is implemented in continuation-passing style, as it accepts a callback of type `impl FnMut(&FileText, pt::Span, &str, io::Error) -> E`. This design allows the caller to produce any desired error type.

That said, I wonder if error handling could be better organized. While this approach works well for all the use cases I've encountered, I suspect it’s far from optimal. Any feedback or suggestions would be greatly appreciated.

### DIFF
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -279,9 +279,9 @@ fn parse_and_normalize_grammar(session: &Session, file_text: &FileText) -> io::R
     }
 }
 
-pub fn report_parse_error<'input, E>(
+pub fn report_parse_error<E>(
     file_text: &FileText,
-    error: parser::ParseError<'input>,
+    error: parser::ParseError<'_>,
     mut reporter: impl FnMut(&FileText, pt::Span, &str, io::Error) -> E,
 ) -> E {
     match error {
--- a/lalrpop/src/build/mod.rs
+++ b/lalrpop/src/build/mod.rs
@@ -15,7 +15,6 @@ use crate::tls::Tls;
 use crate::tok;
 use crate::util::Sep;
 use itertools::Itertools;
-use lalrpop_util::ParseError;
 use sha3::{Digest, Sha3_256};
 use walkdir::WalkDir;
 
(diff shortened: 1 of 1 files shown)


LighghtEeloo:
Yes, your observation is correct. `report_parse_error()` is called with a different callback in the LSP code, and it's intended to be exposed in the public API of lalrpop_internals. I'll fix the clippy lints in the following commit.

lalrpop/src/build/mod.rs
```diff
@@ -266,53 +265,68 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
...
+            &error.message,
+            io::Error::from(io::ErrorKind::InvalidData),
+        ))?,
+    }
+}
+
+pub fn report_parse_error<E>(
+    file_text: &FileText,
+    error: parser::ParseError<'_>,
+    mut reporter: impl FnMut(&FileText, pt::Span, &str, io::Error) -> E,
```
dburgener, line 285: This implementation where the callback takes in an io::Error and then turns around and returns that same error (but now it has the generic type E) feels weird.  It looks like in the LSP implementation it just re-wraps up the io::Error.

io::Error was always a sort of weird return here.  The only reason the block that got factored out returned an io::Error is because that's the error type that goes up the stack.  I think rather than do this whole "pass the io::Error around" dance, it would make more sense to just have the internal implementation use io::ErrorKind::InvalidData unconditionally.  I don't care too much to preserve the one UnexpectedEof case, which is only arguably important in the first place.

lalrpop/src/build/mod.rs
```diff
@@ -266,53 +265,68 @@ fn lalrpop_files<P: AsRef<Path>>(root_dir: P) -> io::Result<Vec<PathBuf>> {
...
+        Err(error) => Err(report_error(
+            file_text,
+            error.span,
+            &error.message,
+            io::Error::from(io::ErrorKind::InvalidData),
+        ))?,
+    }
+}
+
+pub fn report_parse_error<E>(
```
dburgener, line 282: This function needs a doc comment.  One because the whole callback dance is a little weird, and should be explained, but 2, the ultimate goal is for this to become a public export in lalrpop_internals, and we'd like to have it documented for that.

dburgener (changes requested):

lalrpop/src/build/mod.rs
LighghtEeloo: Good point. If we don't need to keep the io::Error around, then we may consider removing it entirely.

lalrpop/src/build/mod.rs
LighghtEeloo: Indeed. I'll add the docs.

LighghtEeloo:
Sorry for the late response due to personal reasons. I've updated the git branch to add documentations and removed `io::Error` from the callback parameter.

dburgener (approved):
This is so much cleaner without the dance passing the error to the callback!

Thanks for the update.  This LGTM.  I'll give another day for any other maintainer reviews before merging.

EDIT: Just saw CI.  This will need a rustfmt run before merge as well.

Pat-Lafon (approved):
👍

dburgener (approved):

PR #1023, opened 2025-01-06
Update itertools to 0.14 and petgraph to 0.7

dburgener:


Pat-Lafon (approved):
Removes two transitive dependencies, nice

PR #1024, opened 2025-01-06
Silence new clippy lint literal_string_with_formatting_args in codegen

dburgener:
This new lint checks if we use format args in string literals that aren't passed to function that handle them like println!().  This makes sense normally, but in code generation contexts, we may be generating code like `rust!(rust, "println!("some format arg {}", arg))` which is a false positive hit for this lint.  There is only one such lint right now, but just blanket silence it in codegen and leave it on for the rest of the codebase, because it doesn't make sense in code generation.

### DIFF
--- a/lalrpop/src/lr1/codegen/mod.rs
+++ b/lalrpop/src/lr1/codegen/mod.rs
@@ -1,3 +1,5 @@
+// In code generation we may output format strings that will be used by the generated code
+#![allow(clippy::literal_string_with_formatting_args)]
 pub mod ascent;
 mod base;
 pub mod parse_table;


dburgener:
CI failure is because this lint name doesn't exist yet on clippy 1.83.  I guess the cleanest fix it to wait until 1.85 is stable and merge this then.

lalrpop/src/lr1/codegen/mod.rs
```diff
@@ -1,3 +1,5 @@
+// In code generation we may output format strings that will be used by the generated code
+#![allow(clippy::literal_string_with_formatting_args)]
```
Pat-Lafon, line 2: Thoughts on using `#![expect(...)]` over allow here in case they end up fixing this false positive?

Pat-Lafon (approved):

lalrpop/src/lr1/codegen/mod.rs
dburgener: I supposed I'm not opposed to `expect`, although I'm not really sure what clippy can do in this case.  There was [some discussion](https://github.com/rust-lang/rust-clippy/pull/13410#issuecomment-2372387787) in the PR adding this lint about false positives in crates that manipulate format strings, and the suggestion was that those crates should allow the lint globally, which makes sense.

We're in a pretty similar boat doing code generation I think.  I'm not sure how clippy would ever know that our use is legitimate.  We write a string out to a file and it contains format characters.  The fact that that file happens to contain valid rust and is later compiled is far outside of what clippy can know.

PR #1032, opened 2025-01-21
Release 0.22.1

dburgener:


dburgener:
We don't have a ton of new user-facing features, but it's been a few months, so it might be nice to get a point release out.  The primary value add in my book is publishing our updates to docs.rs, but there's a handful of user-facing improvements noted in the release notes.

Cargo.toml
```diff
@@ -17,7 +17,7 @@ resolver = "2"
 repository = "https://github.com/lalrpop/lalrpop"
 authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 license = "Apache-2.0 OR MIT"
-version = "0.22.0" # LALRPOP
+version = "0.22.1" # LALRPOP"
```
Pat-Lafon, line 20: Nit: seems an extra `"` showed up at the end here.

Pat-Lafon (approved):
I looked through the commit log and it is largely as you described. I don't see any non-trivial features that were missed. Looks good to go.

While looking through I was reminded of https://github.com/lalrpop/lalrpop/commit/1e72fd66946fa3a18dcb2b2fa5bdd19ae1ac44b2. Maybe this deprecation is something we will make louder in some way in the future.

Cargo.toml
dburgener: Fixed both this and the version.sh bug that introduced the extra `"`

dburgener:
> While looking through I was reminded of https://github.com/lalrpop/lalrpop/commit/1e72fd66946fa3a18dcb2b2fa5bdd19ae1ac44b2. Maybe this deprecation is something we will make louder in some way in the future.

Yeah, your suggestion about displaying a warning at runtime is probably the way to go.  It bugs me a little because the warning is a little disconnected from the actual problem location, but I don't have a better suggestion.  I do think it would be nice to package that in a breaking release though.  Not because it's really "breaking" per-say, just to be user-friendly and give people a good clean chance to handle it.  And since we'll need a breaking release for #979 at some point, we can probably do it whenever we're ready for 0.23?

Pat-Lafon:
Yep, totally agree. Sorry, not so related as to hold up this pr.

PR #1033, opened 2025-01-21
Cleanup README include in include_str

dburgener:
When running cargo doc, the include_str command works fine, however, cargo publish relocates the package into target/package, so the relative path to the README is different, causing cargo publish to fail.

External discussion at:

https://github.com/rust-lang/cargo/issues/13309
https://users.rust-lang.org/t/include-str-does-not-work-when-releasing-because-of-changed-pathes/15551/14

The solution suggested in those links is to symlink the top level README inside your package, and then consistently refer to the README in the package.  That README will be copied into the right place during cargo publish.

### DIFF
--- a/lalrpop/Cargo.toml
+++ b/lalrpop/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "lalrpop"
 description = "convenient LR(1) parser generator"
-readme = "../README.md"
+readme = "README.md"
 workspace = ".."
 default-run = "lalrpop"
 
--- a/lalrpop/README.md
+++ b/lalrpop/README.md
@@ -1,0 +1,1 @@
+../README.md
--- a/lalrpop/src/lib.rs
+++ b/lalrpop/src/lib.rs
@@ -1,4 +1,4 @@
-#![doc = include_str!("../../README.md")]
+#![doc = include_str!("../README.md")]
 #![warn(missing_docs)]
 // Need this for rusty_peg
 #![recursion_limit = "256"]


dburgener:
@Pat-Lafon I'd appreciate it if you'd take a look at this ASAP.

I published lalrpop-util 0.22.1, but then hit the failure described here in lalrpop 0.22.1.  My plan is to merge this, and publish lalrpop 0.22.1 off of this, while leaving the existing lalrpop-util 0.22.1 off the current HEAD alone.  Then I'll tag this commit as 0.22.1 in the log.  Technically lalrpop and lalrpop-util will have been off of different commits, but this PR only touches lalrpop, so it's fine in practice IMO.

Pat-Lafon (approved):
A fun little papercut. LGTM

Pat-Lafon:
> I published lalrpop-util 0.22.1, but then hit the failure described here in lalrpop 0.22.1. My plan is to merge this, and publish lalrpop 0.22.1 off of this, while leaving the existing lalrpop-util 0.22.1 off the current HEAD alone. Then I'll tag this commit as 0.22.1 in the log. Technically lalrpop and lalrpop-util will have been off of different commits, but this PR only touches lalrpop, so it's fine in practice IMO.

Yeah, that is fine with me. I don't think this should be an issue in practice.

dburgener:
> A fun little papercut. LGTM

Very much so.  Frustrating that "cargo doc" or a linter or something didn't catch this.  I don't know if adding a "cargo publish --dry-run" into our CI is worthwhile to avoid this sort of thing?

Pat-Lafon:
Looks like what we would want is https://github.com/rust-lang/rust-clippy/issues/11677, but otherwise yeah a dry-run can't hurt here.

dburgener:
Awesome.  Good find.  I was wondering about a clippy lint.  Unfortunately, it doesn't look like that's had activity in a year...

Pat-Lafon:
Yeah, in following the discussion links, it seems to be a bit of a dead-end at the moment. I also didn't find anything in cargo-release that seems to help.

PR #1034, opened 2025-01-21
Dry run cargo publish in CI

dburgener:
Try to catch publish issues in advance.

Cargo-release is not included in this PR because it has additional
pre-publish checks that I wasn't able to get working in CI.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -60,6 +60,15 @@ jobs:
           toolchain: nightly
       - name: Check the minimum possible crate versions
         run: rm Cargo.lock && cargo +nightly build -Zdirect-minimal-versions
+  publish_dry_run:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


Pat-Lafon:
> Cargo-release is not included in this PR because it has additional pre-publish checks that I wasn't able to get working in CI.

That should be fine because the main usage for cargo-release is be managing tags and other release artifacts.

I see in https://github.com/lalrpop/lalrpop/actions/runs/12895081371/job/35955206316?pr=1034#step:3:8 that the dry-run process raises a warning that the crate version already exists and yet the dry-run process and the action return success. This is nice for our ci to run outside of release pr's, but were you able to check that a dry-run would error in the case of our previous publish issue?

If so, then this pr works for me.

dburgener:
> were you able to check that a dry-run would error in the case of our previous publish issue?

Good suggestion!  Yes, it works: https://github.com/dburgener/lalrpop/pull/14

Pat-Lafon (approved):
👍

PR #1035, opened 2025-01-27
Update logos to 0.15

dburgener:
Now that our msrv is bumped to 1.81, the MSRV of logos is no longer a problem.


Pat-Lafon (approved):
LGTM

PR #1037, opened 2025-01-30
Update to rand 0.9

dburgener:
There is one breaking change that affects us.  gen_* functions have been deprecated in favor of new random_* functions, so update our usage of that.


Pat-Lafon (approved):
LGTM, thanks!

PR #1040, opened 2025-02-04
Supplement section Error recovery with custom lexer in doc

### LINKED ISSUE #1038: lalrpop + logos | returning Err instead of pushing to ErrorRecovery
We are experiencing a similar issue to #992 and https://stackoverflow.com/questions/77999387/lalrpop-doesnt-recover-after-finding-an-invalid-token-even-though-i-should-have
Our code is here:
https://github.com/osta-lang/osta-rs/commit/1cb7beff638db2ddc72374a87f85d884122cbfea
The issue can be replicated from `calculator7` replacing the grammar by:
```rs
use std::str::FromStr;
use crate::ast::{Expr, Opcode};
use lalrpop_util::ErrorRecovery;

grammar<'err>(errors: &'err mut Vec<ErrorRecovery<usize, Token<'input>, &'static str>>);

pub Exprs: Vec<Box<Expr>> = {
    (<Term>)+,
}

Term: Box<Expr> = {
    Num => Box::new(Expr::Number(<>)),
    ! => { errors.push(<>); Box::new(Expr::Error) },
};

Num: i32 = {
    r"[0-9]+" => i32::from_str(<>).unwrap()
};
```
In the `22 * + 3` example I expect to get `22 error error 3`, but instead I get
```
called `Result::unwrap()` on an `Err` value: InvalidToken { location: 3 }
```

chanbengz:
Fix #1038. Made it more clear to users that they need to also update the `extern` block in grammar for lexical error `Token::Error` from external lexer. Otherwise the error recovery may not work properly.

### DIFF
--- a/doc/src/lexer_tutorial/006_error_recovery_custom_lexer.md
+++ b/doc/src/lexer_tutorial/006_error_recovery_custom_lexer.md
@@ -52,13 +52,24 @@ impl<'input> Iterator for Lexer<'input> {
 and then put a new member into `Token` enum to represent the error token
 (or your custom error type).
 
-```rust
+```diff
 pub enum Token {
     //...
 
     // Dont forget the comma
 +   Error
 }
+```
+
+Also update the extern block in `grammar.lalrpop`.
+
+```diff
+extern {
+      type Location = usize;
+      type Error = LexicalError;
+
+      enum Token {
++         "error" => Token::Error,
 ```
 
 Like NaN is a number in JavaScript, we have now an `Token::Error` which is


dburgener (approved):

dburgener:
Thanks!

This change seems like a reasonable doc update.  I'll leave it up for a day or two before merging in case any other maintainers have comments.

I still want to look a little more at #1038.  I think that unlike `InvalidToken`, `UnrecognizedToken` errors should be recoverable, but the simple repro in that bug shows an `InvalidToken` error, and I'm also not currently sure exactly how this "forgetting extern" case breaks things.  So I need to find some time to play with it.  But that doesn't need to hold up this doc change.

chanbengz:
> but the simple repro in that bug shows an `InvalidToken` error

I reproduced it by tweaking example code at `doc/lexer`. The bug is triggered by external lexer rather than LALRPOP's lexer. I have the repro in hand but my laptop is away. Will be back in hours.

> and I'm also not currently sure exactly how this "forgetting extern" case breaks things. 

I guess I figured it out. `lalrpop` assigns index numbers to the tokens in `extern` and then during parsing, `grammar.rs` invokes `lalrpop-util/state_machine.rs` where it looks for the tokens' indices. If some token is "forgotten" and lexer returns it, state machine will receive a `None` value and return `NextToken::Done` that terminates the parsing. The lookup function generated by `lalrpop` is simply a match block.

Pat-Lafon (approved):

Pat-Lafon:
Hmm, I see it automatically closed #1038. Should that be reopened?

dburgener:
> Hmm, I see it automatically closed https://github.com/lalrpop/lalrpop/issues/1038. Should that be reopened?

Let's leave it open for now.  My inclination is that `UnrecognizedToken` should be recoverable, but I haven't had time to dig into it at all.  I'd like to at least poke at it and see if there's something besides docs that can improve the situation.  (No idea when I'll manage to get to it though)

chanbengz:
Here's the repro, plz take a look. https://github.com/lalrpop/lalrpop/compare/master...chanbengz:lalrpop:issue-1038

image

chanbengz:
`grammar.rs` generated

```rust
    fn __token_to_integer<
        'err,
    >(
        __token: &Token,
        _: core::marker::PhantomData<(&'err ())>,
    ) -> Option<usize>
    {
        #[warn(unused_variables)]
        match __token {
            Token::KeywordVar if true => Some(0),
            Token::KeywordPrint if true => Some(1),
            Token::Identifier(_) if true => Some(2),
            Token::Integer(_) if true => Some(3),
            Token::LParen if true => Some(4),
            Token::RParen if true => Some(5),
            Token::Assign if true => Some(6),
            Token::Semicolon if true => Some(7),
            Token::OperatorAdd if true => Some(8),
            Token::OperatorSub if true => Some(9),
            Token::OperatorMul if true => Some(10),
            Token::OperatorDiv if true => Some(11),
            _ => None,
        }
    }
```

Where it might be the trigger

https://github.com/lalrpop/lalrpop/blob/8a9cc50dd3907d775ea8720563b6e772740e6d35/lalrpop-util/src/state_machine.rs#L632-L639

Issue #924, opened 2024-07-19
lalrpop as a library in my project


xuanbachle:
I want to use lalrpop as a library in my project. Particularly, I want to use the grammar, parseTree etc directly from lalrpop. How can I do that? So far, I have added lalrpop as build dependencies and dependencies, but none would allow me to use the grammar module etc from the lalrpop code. 

If I take the lrgrammar.lalrpop and put that in my project, when generating parser for the lrgrammar.lalrpop, it has to use the grammar, parseTree etc from lalrpop and thus would not compile.

ANTLR allows to access the parseTree etc of ANTLR very easily, whereas LALRPOP makes it awfully difficult if I want to access the parseTree.

dburgener:
Short answer: we do not support this today, sorry.  That is by design, so that these internal components can change without impact to users.

Slightly longer answer: I am not wild about exposing all those internal details as public dependencies.  I'm (very slowly) working on a PR review of #658, which adds a crate in the lalrpop workspace needing similar functionality.  So far, the need for that functionality is the biggest concern I have with that PR.  If we did want to support this use case there are a few different ways we could handle it:

1. Just mark the parse tree, grammar and whatever else as pub at the crate level
2. Split the relevant bits into a lalrpop_internal library, which the main lalrpop crate depends on, so that that library can expose these bits while lalrpop does not.  If we did this, we would likely document a recommendation that users not link lalrpop_internal directly, but doing so anyways could support this use case.
3. Bring the functionality that needs access to lalrpop's internal bits directly into lalrpop, and expose a more limited public API to access it.

In my opinion, number 1, which is the easiest thing to implement is essentially a non-starter.  I don't think we want to expose all of these bits are part of our public API.  Number 2 is the most work, and adds to the long-term maintenance burden, but it is cleaner in terms of the API.  Number 3 is nice, but only makes sense for certain use cases.  And then of course the hidden number 4 option is the default, which is to simply not support this use case.

Can I ask why you need this functionality?  I'm entertaining the idea primarily because the use case in #658 is very appealing and seems to have a very similar need, but in general I wouldn't expect this to be something end users need very much.

xuanbachle:
Thank you for the very detailed response.

The use cases for this can be:
+ Synthesis of inputs belonging to an arbitrary grammar
+ Editor for LALRPOP grammars
and possibly many more.

[...]

PR #658, opened 2022-03-20
Draft: Documentation generator for LALRPOP ( markdown, railroad diagram svg, ebnf )

darach:
We use LALRPOP on the [tremor project](https://github.com/tremor-rs/tremor-runtime) for
defining the grammar of our embedded DSLs. Up until the current release, we have been maintaining
a proximate EBNF file and using online railroad diagram generators to assist with our language reference.

This has proven cumbersome, error prone and high maintenance!

This PR is a work in progress extension to LALRPOP that derives EBNF from the LALRPOP grammar, generates
railroad diagrams with support for LALRPOP macro rules visually, and generates markdown for the grammar
that is compatible with github markdown, pandoc, docusaurus.

The markdown can be annotated with human-maintained markdown files.

A few small changes to the core `lalrpop` package is needed to expose the parse tree such that
the grammar can be visited. If changes come to mind to align it better with the other LALRPOP crates
happy to make any such changes.

The same pic-args, session, config and api pattern was used with the `lalrpop-docgen` tool but this
tool hasn't yet been tested with or integrated with `build.rs` style usage.

Signed-off-by: Darach Ennis <darach@gmail.com>

PR #1045, opened 2025-02-19
RFC: Move the bulk of lalrpop into a new lalrpop-internals crate

dburgener:
This is so that we can support requests to access lalrpop internal functions and data structures without committing to bump semver on internal changes.

Example use cases include:
* IDE integration (#1013)
* benchmarking
* fuzzing (#924)
* program synthesis (#934)
* Documentation generation (#658)

As noted in the linked issues and PRs above, there has been interest in this from multiple users with multiple use cases.

The scheme here is that lalrpop-internals would expose whatever is needed for these use cases publically, and would bump semver liberally. The main lalrpop crate only holds the library and binary front-ends, and depends on lalrpop-internals for nearly all the work.

The public API of lalrpop-internals right now includes what is needed for the following use cases:

* lalrpop itself
    * build, log, session
* The documentation generator in #658
    * log, grammar, parser, tok
* The LSP demo at https://github.com/LighghtEeloo/lalrpop/tree/lsp
    * build, file_text, grammar, normalize, parser, session

I haven't gone through the public members of these modules.  It's likely that many of them should be marked pub(crate) rather than pub.  That is something that should be done before this is merged.


lalrpop-internals/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name = "lalrpop-internals"
+description = "Internal lalrpop structures.  If you want a parser generator, you should use the lalrpop crate"
+readme = "README.md"
+workspace = ".."
+# This crate is versioned independently - which is the whole point.  We want to
+# be able to bump semver here without bumping lalrpop/lalrpop-util
+version = "1.0.0"
```
Pat-Lafon, line 8: Should we just set this to "0.1.0" then?

dburgener: I don't have strong feelings one way or the other.  Why prefer 0.1.0 over 1.0.0?

lalrpop-internals/src/parser/lrgrammar.rs
```diff
@@ -1,4 +1,4 @@
-// auto-generated: "lalrpop 0.22.1"
+// auto-generated: "lalrpop-internals 1.0.0"
```
dburgener, line 1: Just noticed this change.  I feel like we probably want to keep "lalrpop 0.22.1" here.  I'll investigate.

Pat-Lafon: Hmm, that seems a little challenging because you are looking for the version of a user of this crate. In some sense, the most precise/accurate versioning is of the `lalrpop-internal` version.

👍 Interested to see what you find

Pat-Lafon (commented):
Hmm, what are the guidelines for which api's should public? Is it as api's are requested they are made public? Whatever lalrpop has as public?

Just to check, are we publishing this crate separately from lalrpop? Also double checking that adding this crate as a workspace member was all that was needed to keep the ci testing/checking as expected

dburgener:
> Hmm, I'm not sure how to evaluate this pr.

Yeah, it is tricky.  My hope with having the PR is that we can evaluate the overall *idea* in the context of actual code implementing it to have a concrete sense of what it looks like to architect this way.

> In some sense, the best description this is of what's not getting moved into lalrpop-internals which I think is the api used in the build script?

Yeah, that's probably a good way to think of it.  It's both the API and the binary build that are left behind in lalrpop.  I was originally trying to think of it as "what do we pull out into internals", but with so many interdependencies, I ultimately decided that this "leave the bare minimum" approach seemed cleaner.

> The expectation is that semver changes won't matter to lalrpop being in a mono-repo right?

I don't think I follow the question...  In terms of the lalrpop repo having a workspace with multiple crates, I don't think there is any semver implication of being in a workspace.  They're just separate crates.

Maybe enumerating a few use cases are helpful:

* Non-user facing bug fix in lalrpop-internals -> we can release a point release on lalrpop-internals and nothing else.  lalrpop will pick it up due to semver
* Breaking change (for lalrpop users) in lalrpop-internals -> We release a breaking lalrpop-internals release.  We take that new breaking release in lalrpop and release a breaking lalrpop with the updated dependency.  Lalrpop wouldn't take a breaking lalrpop-internals by default because of semver
* Breaking change for lalrpop-internals users (but not lalrpop users) (eg, adding a member to `struct parse_tree::Grammar`) -> We publish a breaking change of lalrpop-internals.  We bump the dependency in lalrpop (eg from 1.0.0 to 2.0.0), but we can release lalrpop as a point release rather than a breaking one, since there's no break for lalrpop users

> what is the public surface area we want to expose

Very good question.  I think the next step for this PR is to answer that question.  Right now (almost) everything that was marked `pub` in the mods that are now `pub` is public in lalrpop-internals.  That's likely not what we want.  Before merging this PR, we should pull some of that back to `pub(crate)` or similar.  More on this below.

> and are we worried about the more aggressive semver bumping with the other consumers of this api?

I think that's what they're signing up for getting at the internals.  In practice, I imagine it won't be that bad though.  How often are we actually changing the structs that they need access to.  We do need to narrow in the public API more, but if we can get a sensible public API, we might be able to strike a sensible balance.  Just thinking about what the various efforts linked in the description want access to, I'm not sure much of anything has changed that would cause a semver bump on those things in my time as a maintainer.  That could be an argument against even doing this PR (just make those things public and figure they're unlikely to change).  I just don't like the idea of wedding us to doing semver bumps for a niche use case like these that doesn't affect the vast majority of our users.

> Hmm, what are the guidelines for which api's should public? Is it as api's are requested they are made public? Whatever lalrpop has as public?

I don't immediately see a huge reason to have it be other than "what was requested"?

> Just to check, are we publishing this crate separately from lalrpop?

Yes

> Also double checking that adding this crate as a workspace member was all that was needed to keep the ci testing/checking as expected

I believe so.  The only caveat is that the publish-dry run test is currently red, because lalrpop-internals isn't an upstream package.  I don't think that's something we can really fix unless/until we take this PR and publish a lalrpop-internals.

Otherwise, the lalrpop-test tests are just using lalrpop as (sort of) an external library, so this is transparent to them (as it should be, since that's roughly how our users would see the change).  Adding lalrpop-internals to the workspace causes `cargo test`, `cargo clippy` etc pick up the new crate and run tests, lints, etc.

lalrpop-internals/Cargo.toml
Pat-Lafon: It's fair that in terms of version updates, they will act similar. I guess socially, one expects that something that is 1.0 will have less breaking changes/is more settled. I'm not sure if that's true for these internals?

Pat-Lafon:
> Maybe enumerating a few use cases are helpful:
> 
>     * Non-user facing bug fix in lalrpop-internals -> we can release a point release on lalrpop-internals and nothing else.  lalrpop will pick it up due to semver
> 
>     * Breaking change (for lalrpop users) in lalrpop-internals -> We release a breaking lalrpop-internals release.  We take that new breaking release in lalrpop and release a breaking lalrpop with the updated dependency.  Lalrpop wouldn't take a breaking lalrpop-internals by default because of semver
> 
>     * Breaking change for lalrpop-internals users (but not lalrpop users) (eg, adding a member to `struct parse_tree::Grammar`) -> We publish a breaking change of lalrpop-internals.  We bump the dependency in lalrpop (eg from 1.0.0 to 2.0.0), but we can release lalrpop as a point release rather than a breaking one, since there's no break for lalrpop users

Gotcha, thanks for the clarity. 

> > and are we worried about the more aggressive semver bumping with the other consumers of this api?
> 
> I think that's what they're signing up for getting at the internals. In practice, I imagine it won't be that bad though. How often are we actually changing the structs that they need access to. We do need to narrow in the public API more, but if we can get a sensible public API, we might be able to strike a sensible balance. Just thinking about what the various efforts linked in the description want access to, I'm not sure much of anything has changed that would cause a semver bump on those things in my time as a maintainer. That could be an argument against even doing this PR (just make those things public and figure they're unlikely to change). I just don't like the idea of wedding us to doing semver bumps for a niche use case like these that doesn't affect the vast majority of our users.

Cool. One thing I'm thinking about is that for a couple of the examples given, while not obligated, it's probably in our best interests to give them a heads up about the breaking changes. Hypothetically, an IDE extension not working might be a non-trivial QoL hit that makes an otherwise semver-compatible upgrade unappealing. (Coming from my experiences when needing to run a Haskell codebase and the Haskell language server was enough versions behind that it didn't support ARM-Mac's)

> > Also double checking that adding this crate as a workspace member was all that was needed to keep the ci testing/checking as expected
> 
> I believe so. The only caveat is that the publish-dry run test is currently red, because lalrpop-internals isn't an upstream package. I don't think that's something we can really fix unless/until we take this PR and publish a lalrpop-internals.

Hmm, that is certainly an interesting edge case. Like you said, not clear that there is something to be done.

PR #1047, opened 2025-02-22
Support ASCII escape codes in string literals

n8pjl:
This PR adds support for the `\0` and `\x##` escape sequences in standard string literals. Use the format described by the Rust reference (https://doc.rust-lang.org/stable/reference/tokens.html#character-literals); disallowing 8-bit characters and truncated escape sequences.

Supporting the `\x` escape code allows matching terminals that contain nonprintable ASCII characters without resorting to a regex string.


dburgener (approved):
LGTM, thanks for the PR!

Pat-Lafon (approved):
Thanks!

dburgener:
Just as an FYI, we're having some discussions right now around our release process in #1006.  My plan unless things change is to release 0.22.2 including this PR later this week or early next week.

PR #1049, opened 2025-02-25
Run checks on PRs to 0.22.x

dburgener:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: ["master"]
   pull_request:
-    branches: ["master", "0.22.x"]
+    branches: ["master", "0.[0-9]+.x"]
   merge_group:
 
 permissions:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: ["master"]
   pull_request:
-    branches: ["master"]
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: ["master"]
   pull_request:
-    branches: ["master"]
+    branches: ["master", "0.22.x"]
```
Pat-Lafon, line 7: Nit: just noting that you can use a filter here if we want to stick to this branch naming scheme... I think something like `"0.*.x"` or `0.**` if I understand this correctly https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-filters

Pat-Lafon (approved):
LGTM, I guess we won't know if it works until we try it!

.github/workflows/test.yaml
dburgener: Clicking through, I found [this cheat sheet](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) and tried to write a more precise filter.

dburgener:
> I guess we won't know if it works until we try it!

Yep.  I'm going to try to poke at #1050 until it runs checks.  Actions has some weird corner cases when merging changes to them that I tend to struggle to reason about, so I'm not sure if I'll have to merge a backport of this first or something.

PR #1050, opened 2025-02-25
0.22.2 backports

dburgener:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: ["master"]
   pull_request:
-    branches: ["master"]
+    branches: ["master", "0.[0-9]+.x"]
   merge_group:
 
 permissions:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -64,10 +64,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Dry run lalrpop-util (cargo publish)
-        run: cargo publish --dry-run -p lalrpop-util
-      - name: Dry run lalrpop (cargo publish)
-        run: cargo publish --dry-run -p lalrpop
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
+      - name: Dry run publish
+        run: cargo +nightly -Z package-workspace publish --dry-run -p lalrpop -p lalrpop-util
 
   feature_powerset:
     runs-on: ubuntu-latest
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,7 +70,7 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "calculator"
-version = "0.22.1"
+version = "0.22.2"
 dependencies = [
  "lalrpop",
  "lalrpop-util",
@@ -78,7 +78,7 @@ dependencies = [
 
 [[package]]
 name = "cfg"
-version = "0.22.1"
+version = "0.22.2"
 dependencies = [
  "lalrpop",
  "lalrpop-util",
@@ -224,7 +224,7 @@ dependencies = [
 
 [[package]]
 name = "lalrpop"
-version = "0.22.1"
+version = "0.22.2"
 dependencies = [
  "ascii-canvas",
  "bit-set",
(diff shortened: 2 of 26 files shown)


dburgener:
Switched back to draft.  I want to add release notes in this PR, and then I can make the release directly off it once it merges, and forward port the release notes back to master.

Pat-Lafon:
> This is our first attempt at a release with the new cargo publish check enabled. It's highlighting a new issue that when we dry-run lalrpop, it fails because lalrpop-util hasn't actually been published. This isn't an issue for day-in-day-out development, since typically it's dry-running the already published versions, but on a PR to actually do a version bump, lalrpop-util 0.22.2 isn't uploaded yet.
> 
> I don't immediately see a clean solution to this, besides just ignoring the failing check on version bump PRs.

I'm confused why this hasn't happened before? My vague recollection was trying out a dry-run before I did a publish for real(looking back at the git blame, maybe the difference is that the ci uses cargo publish instead of cargo release #1034).

There seems to be a nightly `--workspace` flag but I found that gave me an error I didn't expect. I did discover that on nightly one can do `cargo +nightly -Z package-workspace publish --dry-run -p lalrpop -p lalrpop-util` which if it's only for ci validation might be ok?

yannham:
> There seems to be a nightly --workspace flag but I found that gave me an error I didn't expect. I did discover that on nightly one can do cargo +nightly -Z package-workspace publish --dry-run -p lalrpop -p lalrpop-util which if it's only for ci validation might be ok?

Not directly related maybe, but what was the error? It happens that the implementers of the feature are my colleagues, so I'm curious.

Otherwise, yeah, this is an annoying issue when releasing interdependent crates. I guess running `cargo --workspace` in the CI is still better than nothing, although it has the shortcoming that the dry-run is different from the actual logic we use for releasing, so one could fail while the other would succeed. Still, I don't see any alternative either. Usually I just publish them in topological order and handle the mess if one of the crates fail in the middle of the release pipeline, but saying that it's far from ideal is an understatement :upside_down_face: .

Pat-Lafon:
> > There seems to be a nightly --workspace flag but I found that gave me an error I didn't expect. I did discover that on nightly one can do cargo +nightly -Z package-workspace publish --dry-run -p lalrpop -p lalrpop-util which if it's only for ci validation might be ok?
> 
> Not directly related maybe, but what was the error? It happens that the implementers of the feature are my colleagues, so I'm curious.

```txt
❯ cargo +nightly -Z package-workspace publish --dry-run --workspace
error: `calculator`, `nobol`, `pascal`, `whitespace`, `lexer`, `lexer-modes`, `cfg`, `lalrpop-test` cannot be published.
`package.publish` must be set to `true` or a non-empty list in Cargo.toml to publish.
```

It sees that some of our workspace members have `publish=false` in their Cargo.toml's and gives up?

> Otherwise, yeah, this is an annoying issue when releasing interdependent crates. I guess running `cargo --workspace` in the CI is still better than nothing, although it has the shortcoming that the dry-run is different from the actual logic we use for releasing, so one could fail while the other would succeed. Still, I don't see any alternative either. Usually I just publish them in topological order and handle the mess if one of the crates fail in the middle of the release pipeline, but saying that it's far from ideal is an understatement 🙃 .

That's fair. I guess as long as the validation it does is a subset of what happens during a publish that some checks are better than no checks? But yeah, worst case just drop/ignore the github action.

dburgener:
I tried `cargo +nightly -Z package-workspace publish --dry-run -p lalrpop -p lalrpop-util` on commit a4fd1d4df88c9293d3694bb1e0af69b4ca5ed7d1 which is the one that failed publish, prompting us to add the check, and it did correctly find the issue around the README location that we saw there.  I also tried it on this branch, and it passes.  In my opinion, if this new approach catches the motivating use case and avoids the false positive here, it seems good enough.  @Pat-Lafon do you want to submit a PR for that, since you suggested it?

dburgener:
@Pat-Lafon I'd like to get this released soon, so that we get #1063 into the hands of users before they start seeing clippy warnings on their generated code.  Can you hit approve before we go?  You shouldn't need to do an in depth code review - everything here is just a clean backport of stuff already on master.  But a once-over and a proofread of the release notes would be good.

Pat-Lafon (approved):
These are all as I expect

Pat-Lafon:
(nit: I have changed the release date to May though I'm not sure it's super relevant)

dburgener:
Lalrpop 0.22.2 published!

Pat-Lafon:
🎉

PR #1051, opened 2025-02-27
Run a workspace based dry-run

dburgener:
This avoids the failure where lalrpop-util isn't actually uploaded (because of dry-run) when we try dry-running lalrpop.


Pat-Lafon (approved):
Sorry, thanks for getting to this!

PR #1052, opened 2025-03-04
Make ParseError hashable

liferooter:
Motivation:
- Why not?
- Better to implement a trait from `std` than not.
- I will finally be able to use this type with `salsa-rs`


Pat-Lafon:
Yeah, makes sense. The only thing I was concerned about was to double check this doesn't end up forcing users to have their custom error type implement Hash(For example, floats aren't hashable by default).

The derive here does work somewhat nicely in only impl-ing Hash when all the generic parameters implement Hash https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f9b529bbf48727de5559c53a66f9f7cc

It is something that Lalrpop wouldn't want to accidentally turn into a hard requirement I think. On the other hand, because of the orphan rule, consumers of ParseError(which is in the lalrpop_util crate and publically exposed), can't implement Hash themselves here.

liferooter:
Yes, deriving `Hash` doesn't introduce any hard requirements. It only automatically implements `Hash` if all generic parameters are hashable too, so it can cause no problems at all, not even conflicting implementations (thanks to orphan rule).

dburgener (approved):

dburgener:
I had the same concern initially.  It seems as though it is safe though.  Based on that, this seems good to me.

The other question I have is if we want to take this for the 0.22.2 release.  According to the [api guidance](https://rust-lang.github.io/rfcs/1105-api-evolution.html#minor-change-implementing-any-non-fundamental-trait) adding a new trait is not considered a breaking change, however they highlight corner cases when it can break existing code.  Since we're looking to do a breaking 0.23 release in a few months anyways, my preference would be to error on the side of caution and save this for the breaking release.

Pat-Lafon:
> Yes, deriving `Hash` doesn't introduce any hard requirements.

Just to clarify, the point is not that this change itself does that, but because a hash implementation could be available, at some later point lalrpop_util accidentally starts to rely on a hash implementation always being available(an unintentional breaking change). Overall this seems rather vague so I think it is fine. I might go add a test case with a non-hashable user error if one doesn't already exist.

Pat-Lafon:
#1053 adds a test that uses a float as a user error. I believe that addresses the concerns mentioned for this pr.

dburgener:
Merged.  Thanks for the PR!

Issue #1054, opened 2025-03-23
Cannot add grammar to submodule


aetilley:
My package is structured as follows

```
lalrcase:
    build.rs
    src/
        lib.rs
        ast/
            my_grammar.lalrpop
            mod.rs
```
mod.rs contains 
```


use lalrpop_util::lalrpop_mod;
lalrpop_mod!(my_grammar);


```

lib.rs contains

```
pub mod ast;
```

build.rs contains
```
fn main() {
    lalrpop::process_root().unwrap();
}
```

When I `cargo build` I get 

```
error: couldn't read /Users/arthurtilley/Desktop/lalcase/target/debug/build/cool_rust-8c28549248c0c0d7/out/my_grammar.rs: No such file or directory (os error 2)
 --> src/ast/mod.rs:2:1
  |
2 | lalrpop_mod!(my_grammar);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `include` which comes from the expansion of the macro `lalrpop_mod` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Can we not add grammars to submodules?  The Quickstart guide says "(note that you can also add this (the `lalrpop_mod!`) line to a foo.rs module definition instead, which will then create a submodule foo::grammar):"

chanbengz:
How about check https://docs.rs/lalrpop/latest/lalrpop/struct.Configuration.html#method.process_dir or `set_in_dir`?

aetilley:
@chanbengz 

Looks like no.

I tried both adding
```
    let config = Configuration::new();
    let grammar_path = Path::new("./src/ast/");
    config.process_dir(grammar_path).unwrap();
```
and
```
    let mut config = Configuration::new();
    let grammar_path = Path::new("./src/ast");
    config.set_in_dir(grammar_path);
```
to the beginning of my build.rs main function and the problem persists.

chanbengz:
Weird. It works well in my machine. My setup is exactly as yours but with an extra `main.rs` for testing

```
.
├── Cargo.toml
├── build.rs
└── src
    ├── ast
    │   ├── calculator1.lalrpop
    │   └── mod.rs
    ├── lib.rs
    └── main.rs
```

I barely tweaking the example of calculator in doc.

dburgener:
[...]

PR #1055, opened 2025-03-29
Path sep

Pat-Lafon:
I didn't see tests for this path of `lalrpop_mod`. It also seems like adding the path separator(which is only marginally tricky because it needs to be a string literal at macro expansion time) is a reasonable paper-cut to fix?

### DIFF
--- a/lalrpop-util/src/lib.rs
+++ b/lalrpop-util/src/lib.rs
@@ -232,6 +232,20 @@ pub struct ErrorRecovery<L, T, E> {
     pub dropped_tokens: Vec<(L, T, L)>,
 }
 
+#[cfg(not(target_os = "windows"))]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! path_separator {
+    {} => { "/" }
+}
+
+#[cfg(target_os = "windows")]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! path_separator {
+    {} => { "\\" }
+}
+
 /// Define a module using the generated parse from a `.lalrpop` file.
 ///
 /// You have to specify the name of the module and the path of the file
@@ -264,7 +278,7 @@ macro_rules! lalrpop_mod {
         #[allow(clippy::needless_lifetimes)]
         #[allow(clippy::let_unit_value)]
         #[allow(clippy::just_underscores_and_digits)]
-        $(#[$attr])* $vis mod $modname { include!(concat!(env!("OUT_DIR"), $source)); }
+        $(#[$attr])* $vis mod $modname { include!(concat!(env!("OUT_DIR"), lalrpop_util::path_separator!(), $source)); }
     };
 }
 
--- a/lalrpop-test/src/lib.rs
+++ b/lalrpop-test/src/lib.rs
@@ -183,6 +183,18 @@ mod user_defined_error_visibility;
 
 lalrpop_mod_test!(zero_length_match);
 
+// Using lalrpop mod here to supply the filename path
+lalrpop_mod!(
+    nest_with_path_sep,
+    "nested_directory_grammars/error_issue_1054.rs"
+);
+
+// Using lalrpop mod here to supply the filename path
+lalrpop_mod!(
+    nest_with_path_sep2,
+    "/nested_directory_grammars/error_issue_1054.rs"
+);
+
 /// This constant is here so that some of the generator parsers can
 /// refer to it in order to test `super::` handling in action code.
 const ZERO: i32 = 0;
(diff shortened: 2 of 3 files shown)


lalrpop-util/src/lib.rs
```diff
@@ -232,6 +232,20 @@ pub struct ErrorRecovery<L, T, E> {
...
 }
 
+#[cfg(not(target_os = "windows"))]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! path_separator {
+    {} => { "/" }
+}
+
+#[cfg(target_os = "windows")]
```
dburgener, line 242: I wonder if we might want to add a windows test run into our testing matrix, just for sanity checking.  As far as I see this is our first `cfg(target_os = "windows")`, although I don't have a particular amount of confidence to say that we don't have some hidden non-portability somewhere, except that I assume someone would have complained if we didn't work on windows.

lalrpop-util/src/lib.rs
```diff
@@ -232,6 +232,20 @@ pub struct ErrorRecovery<L, T, E> {
     pub dropped_tokens: Vec<(L, T, L)>,
 }
 
+#[cfg(not(target_os = "windows"))]
+#[doc(hidden)]
+#[macro_export]
```
dburgener, line 237: I don't have a ton of knowledge on macro scoping rules, but does this really need `macro_export`?  I thought if it was only used farther down in the same file it wouldn't?  That would allow removing of `doc(hidden)` as well (although I assume I'm just wrong and there's some reason it needs `macro_export`).

dburgener (commented):
The overall idea seems great, just two small questions.

lalrpop-util/src/lib.rs
Pat-Lafon: https://github.com/lalrpop/lalrpop/pull/1056

lalrpop-util/src/lib.rs
Pat-Lafon: Yeah, the tricky part here is that this macro doesn't get expanded in the definition of lalrpop_mod in this file, it gets expanded where lalrpop_mod is used which is in the users file. This is why it is exported.

Alternatively, we could create 2 versions of lalrpop_mod, one for the windows case and one other-wise. This would let us hard-code the path separator instead of choosing it via a macro.

dburgener (approved):

dburgener:
I think this is good to land as is.  I wouldn't mind landing #1056 first just for the extra confidence and then rebasing this to get a CI run with the new windows test case.

Pat-Lafon:
I've rebased on the recent in_dir process changes. It seems like what that changed from this pr's perspective is that in the out directory, it is now flat so all grammars are at `.../out/*grammar_name*.rs` where as before this pr expected them to be at `.../out/nested_directory_grammars/*grammar_name*.rs`. Just double checking my understanding.

The fix for that will be to make the test look like the following. Note that someone using the setup in #1054 would need to drop the `/ast/` path part in the their lalrpop_mod when they upgrade I think.

```rs
lalrpop_mod!(
    nest_with_path_sep,
    "error_issue_1054.rs"
);

// Using lalrpop mod here to supply the filename path
lalrpop_mod!(
    nest_with_path_sep2,
    "/error_issue_1054.rs"
);
```

(My understanding from trying to debug this by comparing the nice testcase in https://github.com/lalrpop/lalrpop/blob/4ed193d19c9fa54486cd45f9324a051bed7886f2/lalrpop/src/api/test.rs#L105-L115 with some code to construct the out_dir structure)

dburgener:
Interesting.  That wasn't an intentional part of the change.  I'm looking at the code from #979 and I think I see why it happened.  Let me poke at it a bit to see if there's a clean way to restore the previous behavior in terms of the structure in OUT_DIR, without unintentionally dropping files in `in_dir`.  It *might* be as simple as just reverting this part:

```
-         .join(lalrpop_file.strip_prefix(in_dir).unwrap_or(lalrpop_file))
+        .join(
+            lalrpop_file
+                .file_name()
+                .ok_or_else(|| io::Error::from(io::ErrorKind::InvalidInput))?,
+        )
```

I've got a super busy week here, but once I get some time, I'll poke at it and see about adding some test cases to the new api tests for this scenario as well.  I haven't really thought through it deeply yet, so there are likely some extra subtleties.

Pat-Lafon:
Similarly this pr is waiting on the clippy lint changes to see if there is anything hiding underneath the current clippy errors.

dburgener:
Looks great, merged!  Thanks!

PR #1056, opened 2025-03-31
Windows workflow in github actions

Pat-Lafon:
Adding a windows runner took a few more attempts than I would have expected. This change adds the os as part of the matrix of choices for the build job, using both ubuntu-latest and windows-latest. Generally, we can probably keep the rest of the workflows defaulting to ubuntu-latest

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,8 +29,10 @@ jobs:
           - true
         include:
           - rust: stable
+            os: ubuntu
             cargo-update: false
           - rust: $MSRV
+            os: ubuntu
             cargo-update: false
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -43,7 +43,7 @@ jobs:
         if: matrix.cargo-update
         run: cargo update
       - name: Run tests
-        run: ./tools/ci.sh
+        run: bash ./tools/ci.sh
   lint:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,9 +15,12 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}-latest
     strategy:
       matrix:
+        os:
+          - ubuntu
+          - windows
         rust:
           - stable
           - beta
--- a/lalrpop-test/src/lib.rs
+++ b/lalrpop-test/src/lib.rs
@@ -1098,8 +1098,11 @@ fn verify_lalrpop_generates_itself() {
         .expect("lalrpop run failed")
         .success());
 
-    let actual = fs::read_to_string(grammar_file.with_extension("rs")).unwrap();
-    let expected = fs::read_to_string(copied_grammar_file.with_extension("rs")).unwrap();
+    let mut actual = fs::read_to_string(grammar_file.with_extension("rs")).unwrap();
+    let mut expected = fs::read_to_string(copied_grammar_file.with_extension("rs")).unwrap();
+    // Make the equality check whitespace agnostic(specifically for different line endings)
+    actual.retain(|c| !c.is_whitespace());
+    expected.retain(|c| !c.is_whitespace());
     assert!(
         actual == expected,
         "The snapshot does not match what lalrpop generates now.\n\
--- a/lalrpop-test/src/lib.rs
+++ b/lalrpop-test/src/lib.rs
@@ -1075,6 +1075,7 @@ fn generics_issue_417() {
 }
 
 #[test]
+#[cfg(not(target_os = "windows"))] // Windows uses different line endings
 fn verify_lalrpop_generates_itself() {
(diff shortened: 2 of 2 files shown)


Pat-Lafon:
Specifically, this adds the jobs: build - windows - (stable/beta/nightly)

dburgener:
I was poking at this a little this weekend.  I think that the windows runs are not actually running tests (https://github.com/lalrpop/lalrpop/actions/runs/14176976000/job/39714116560?pr=1056).  Here's my version on top of your path_sep PR: https://github.com/dburgener/lalrpop/pull/15

I had to switch to "bash tools/ci.sh" to get the tests to run out of powershell.  Once they ran, the verify lalrpop generates itself test fails.  I haven't looked into that, but I assume that it's (at least) because of windows line endings.  We probably want to just skip that test on windows.

Pat-Lafon:
> I was poking at this a little this weekend. I think that the windows runs are not actually running tests (https://github.com/lalrpop/lalrpop/actions/runs/14176976000/job/39714116560?pr=1056). Here's my version on top of your path_sep PR: [dburgener#15](https://github.com/dburgener/lalrpop/pull/15)

I didn't realize you had been testing this out, we can take your changes instead/convert my changes to align with yours.

> I had to switch to "bash tools/ci.sh" to get the tests to run out of powershell. Once they ran, the verify lalrpop generates itself test fails. I haven't looked into that, but I assume that it's (at least) because of windows line endings. We probably want to just skip that test on windows.

Ugh, yeah, this change has been annoying because of all of these subtle/silent issues that aren't well surfaced. We could also make the test whitespace agnostic since we aren't printing the diff anyways? Something like `actual.retain(|c| !c.is_whitespace());` I guess

dburgener:
> I didn't realize you had been testing this out, we can take your changes instead/convert my changes to align with yours.

Yeah, sorry, I had started poking at it, but hadn't gotten the windows test passing and got pulled off into something else and didn't get back to it to let you know.

> Ugh, yeah, this change has been annoying because of all of these subtle/silent issues that aren't well surfaced. We could also make the test whitespace agnostic since we aren't printing the diff anyways? Something like actual.retain(|c| !c.is_whitespace()); I guess

Yeah, I guess that would work...  I'm struggling with the use case though.  To some extent the test is a regression test, but the bigger point is to make sure that you remembered to run `./update_lalrpop.sh` if you needed to so that the bootstrapping works.  But developing directly on Windows and changing the lalrpop grammar isn't great, since it will introduce windows line endings.  I think the point of adding Windows tests is really about lalrpop users running Windows rather than lalrpop devs, so maybe this test isn't worth adding value to?

Because a cost of dropping whitespace is not noticing whitespace changes that we'd want to notice (including window-line ending conversion).

Pat-Lafon:
> > Ugh, yeah, this change has been annoying because of all of these subtle/silent issues that aren't well surfaced. We could also make the test whitespace agnostic since we aren't printing the diff anyways? Something like actual.retain(|c| !c.is_whitespace()); I guess
> 
> Yeah, I guess that would work... I'm struggling with the use case though. To some extent the test is a regression test, but the bigger point is to make sure that you remembered to run `./update_lalrpop.sh` if you needed to so that the bootstrapping works. But developing directly on Windows and changing the lalrpop grammar isn't great, since it will introduce windows line endings. I think the point of adding Windows tests is really about lalrpop users running Windows rather than lalrpop devs, so maybe this test isn't worth adding value to?
> 
> Because a cost of dropping whitespace is not noticing whitespace changes that we'd want to notice (including window-line ending conversion).

So the issue is that: someone develops on windows, they generate a new grammar with windows line endings, this passes tests(both the ones which parse lalrpop grammars and this whitespace-agnostic verify grammar test), a non-windows dev comes along and this breaks their workflow in some way?

I think the simplest change then is to just cfg guard this test with `#[cfg(not(target_os = "windows"))]` so at the testcase level it's clear that this isn't intended to work on windows/all of the scripts can remain unchanged.

dburgener:
> So the issue is that: someone develops on windows, they generate a new grammar with windows line endings, this passes tests(both the ones which parse lalrpop grammars and this whitespace-agnostic verify grammar test), a non-windows dev comes along and this breaks their workflow in some way?

I think my concern is more about unnecessary flipping back and forth.  If we have a windows and non-windows contributor regularly updating the grammar, then we're constantly flipping line endings in different commits. 

> I think the simplest change then is to just cfg guard this test with #[cfg(not(target_os = "windows"))] so at the testcase level it's clear that this isn't intended to work on windows/all of the scripts can remain unchanged.

I agree with this approach.  If you'd be okay updating here to get this across the finish line, we can just drop the branch I was working on.

Pat-Lafon:
> > I think the simplest change then is to just cfg guard this test with #[cfg(not(target_os = "windows"))] so at the testcase level it's clear that this isn't intended to work on windows/all of the scripts can remain unchanged.
> 
> I agree with this approach. If you'd be okay updating here to get this across the finish line, we can just drop the branch I was working on.

I have feature-gated this test and reverted the white-space removing commit

dburgener (approved):

dburgener:
Thanks!  This LGTM.  I've approved and merged it.

PR #1057, opened 2025-04-18
docs: fix incorrect use of string literal

mdqst:


dburgener:
I think the confusion may be that the docs mark this as a "rust" code block.  It is in fact a lalrpop block.  This grammar is valid in lalrpop, but not in pure rust.

The mislabeling of "lalrpop" blocks as "rust" blocks seems prevalent at least in this file.  I didn't look at other files though.  I can submit a PR to relabel the blocks as lalrpop blocks.  I don't know if there's additional changes needed to make the doc cleaner to point users to put these changes in the lalrpop grammar?  It seems clear enough to me, but I have the curse of knowledge when it comes to lalrpop, so I don't know.

dburgener:
I opened #1083.  I checked the rest of the lexer-tutorial and found one other instance of a mislabeled lalrpop block in 006, but the rest of the files in that tutorial looked fine to me.

Pat-Lafon:
Ooooh, I see. I didn't understand that initially. I think that is what is happening though it might be more of a visual/documentation confusion. The blocks are all syntax highlighted the same and the jump from one section to the other is not clear that they are separate files.

Pat-Lafon:
@mdqst did we address the issue correctly(Being a confusion over what is Lalrpop code versus what is Rust code)? If not, we can discuss improvements to the docs here otherwise maybe we can close this pr.

PR #228, opened 2017-04-24
Issue226 ltgt construct should work with named input

wieczyk:
Solution to #226.

Lalrpop validates also usage of the `<>` expression with struct constructors -- it requires to assign field names to values.

* Example

```
pub Foo: Foo = {
    <Num> Num <Num> => Foo{<>}
};
```

Provokes error message:
```
wieczyk-> lalrpop -f src/example.lalrpop 
processing file `src/example.lalrpop`
src/example.lalrpop:7:5: 8:0 error: the `<>` expression requires to explicitly assign fields' names to values
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~+
|     <Num> Num <Num> => Foo{<>} |
| };
+~
```

PR #1058, opened 2025-04-25
Inline format args

dburgener:
Nightly clippy has moved the uninlined-format-args lint from pedantic to style, so it's now triggering on our code.

https://github.com/rust-lang/rust-clippy/pull/14160

### DIFF
--- a/lalrpop/src/normalize/token_check/test.rs
+++ b/lalrpop/src/normalize/token_check/test.rs
@@ -19,18 +19,17 @@ fn check_err(expected_err: &str, grammar: &str, span: &str) {
 fn check_intern_token(grammar: &str, expected_tokens: Vec<(&'static str, &'static str)>) {
     let parsed_grammar = validate_grammar(grammar).expect("validate");
     let intern_token = parsed_grammar.intern_token().expect("intern_token");
-    println!("intern_token: {:?}", intern_token);
+    println!("intern_token: {intern_token:?}");
     for (input, expected_user_name) in expected_tokens {
         let actual_user_name =
             interpret::interpret(&intern_token.dfa, input).map(|(index, text)| {
                 let user_name = &intern_token.match_entries[index.index()].user_name;
                 (user_name.clone(), text)
             });
-        let actual_user_name = format!("{:?}", actual_user_name);
+        let actual_user_name = format!("{actual_user_name:?}");
         if expected_user_name != actual_user_name {
             panic!(
-                "input `{}` matched `{}` but we expected `{}`",
-                input, actual_user_name, expected_user_name
+                "input `{input}` matched `{actual_user_name}` but we expected `{expected_user_name}`"
             );
         }
     }
--- a/doc/calculator/src/ast.rs
+++ b/doc/calculator/src/ast.rs
@@ -24,8 +24,8 @@ impl Debug for Expr {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
         use self::Expr::*;
         match *self {
-            Number(n) => write!(fmt, "{:?}", n),
-            Op(ref l, op, ref r) => write!(fmt, "({:?} {:?} {:?})", l, op, r),
+            Number(n) => write!(fmt, "{n:?}"),
+            Op(ref l, op, ref r) => write!(fmt, "({l:?} {op:?} {r:?})"),
             Error => write!(fmt, "error"),
         }
     }
@@ -35,8 +35,8 @@ impl Debug for ExprSymbol<'_> {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
         use self::ExprSymbol::*;
         match *self {
-            NumSymbol(n) => write!(fmt, "{:?}", n),
-            Op(ref l, op, ref r) => write!(fmt, "({:?} {:?} {:?})", l, op, r),
+            NumSymbol(n) => write!(fmt, "{n:?}"),
+            Op(ref l, op, ref r) => write!(fmt, "({l:?} {op:?} {r:?})"),
             Error => write!(fmt, "error"),
         }
     }
--- a/doc/calculator/src/main.rs
+++ b/doc/calculator/src/main.rs
@@ -76,7 +76,7 @@ fn calculator4() {
     let expr = calculator4::ExprParser::new()
         .parse("22 * 44 + 66")
         .unwrap();
-    assert_eq!(&format!("{:?}", expr), "((22 * 44) + 66)");
+    assert_eq!(&format!("{expr:?}"), "((22 * 44) + 66)");
 }
 
 lalrpop_mod_doc!(pub calculator5);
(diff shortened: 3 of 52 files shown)


dburgener:
IMO this is a little debatable.  I think overall it's a readability improvement, but there are a few cases that are debatable.  It's also a lot of churn.  I'm a little back and forth on this one, but I think my opinion is that it's a net plus.  The other option of course would be to silence the lint.

lalrpop/src/normalize/token_check/test.rs
```diff
@@ -19,18 +19,17 @@ fn check_err(expected_err: &str, grammar: &str, span: &str) {
...
                 (user_name.clone(), text)
             });
-        let actual_user_name = format!("{:?}", actual_user_name);
+        let actual_user_name = format!("{actual_user_name:?}");
         if expected_user_name != actual_user_name {
             panic!(
-                "input `{}` matched `{}` but we expected `{}`",
-                input, actual_user_name, expected_user_name
+                "input `{input}` matched `{actual_user_name}` but we expected `{expected_user_name}`"
             );
```
Pat-Lafon, line 33: An example of a change that I view as having improved readability

Pat-Lafon (approved):
I'm in favor. I find this more readable so I appreciate clippy enforcing this for future code changes.

I'm not sure about the churn issue since I think tools impacted by this(like git blame) have methods to ignore style commits. If this would impact your workflow then we can ignore the lint but otherwise lets merge.

dburgener:
@Pat-Lafon

So I tested this again after merge and realized that I missed a lot.  Notably this PR didn't cover examples in generated code.  I also haven't ran with `--test` yet to find examples in tests.  I have a PR in process to clean up generated code examples, but I've hit a problem.  We don't always support the cleanest syntax for this.  Take a look at this change to satisfy the lint for example:

```
diff --git a/lalrpop-test/src/error_recovery_span.lalrpop b/lalrpop-test/src/error_recovery_span.lalrpop
index a5f8dcf7..561e5e6b 100644
--- a/lalrpop-test/src/error_recovery_span.lalrpop
+++ b/lalrpop-test/src/error_recovery_span.lalrpop
@@ -44,5 +44,5 @@ Expr1: String = {

 Expr2: String = {
     Num => format!("{:?}", <>),
-    "(" <Expr> ")" => format!("({})", <>),
+    "(" <Expr> ")" => format!("({<>})"),
 };
```

That fails to generate a parser with the following error:

```
/home/dburgener/lalrpop/lalrpop-test/src/error_recovery_span.lalrpop:47:5: 47:39 error: Using `<>` between curly braces (e.g., `{<>}`) only works when your parsed values have been given names (e.g., `<x:Foo>`, not just `<Foo>`)

  --- stderr
        "(" <Expr> ")" => format!("({<>})"),
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  thread 'main' panicked at lalrpop-test/build.rs:8:10:
```

That error comes from https://github.com/lalrpop/lalrpop/blob/6521858249733d4fad668df4780b22d984c4f15d/lalrpop/src/normalize/prevalidate/mod.rs#L288.  Obviously the workaround suggested by the error message would work, but it's definitely an ergonomic regression to force name usage here.

I don't know why prevalidate is enforcing that rule.  I assume there's a reason, but I don't know yet what it is.  The options I see are:

* Figure out if that rule can be relaxed
* Silence the lint on user supplied action code
* Update all of our code to use named bindings in that situation (and advise our users to do the same)

In terms of number one, I'm not sure how that would support this interesting example from our tests:

```
pub Query: String = {
    <Keyword> <Table> => format!("{} {}", <>)
};
```

I didn't know you could do that, but it looks like it can cleanly take both anonymous bindings and supply them to format jointly, which is kind of cool.  But even if we supported `{<>}`, I think we'd **have** to name them in at least this case if we want to satisfy the lint.

Thoughts?

Pat-Lafon:
> Thoughts?

Ok, I got this setup on my end to experiment. I see some of the lint changes related to `panic!` and some of the doc grammars that your pr in process probably already covers. I also see a similar the pattern you referenced as `<Expr> "+" <Expr1> => format!("{} + {}", <>),` which looks nifty.

I would like to note that the pre-validate step seems to miss a variation of this which it reports with no errors
```diff
diff --git a/lalrpop-test/src/error_recovery_span.lalrpop b/lalrpop-test/src/error_recovery_span.lalrpop
index 9030fac..22d6083 100644
--- a/lalrpop-test/src/error_recovery_span.lalrpop
+++ b/lalrpop-test/src/error_recovery_span.lalrpop
@@ -43,6 +43,6 @@ Expr1: String = {
 };
 
 Expr2: String = {
-    Num => format!("{:?}", <>),
+    Num => format!("{<>:?}"),
     "(" <Expr> ")" => format!("({})", <>),
 };
```
Which seems to be because of the limited detection in https://github.com/lalrpop/lalrpop/blob/6521858249733d4fad668df4780b22d984c4f15d/lalrpop/src/normalize/norm_util.rs#L70 

I was looking at the git blame for this error message and I noticed it comes from https://github.com/lalrpop/lalrpop/pull/228 ... I wonder, is lalrpop incorrectly identifying the format string braces as the braces of a struct(which needs specific field names)?

I explored option 1(under the assumption above). The change is somewhat straight forward I think. In your example, the `<>` gets converted into a comma-separated string. We can instead detect the multiple `<>` in something like
```
pub Query: String = {
    <Keyword> <Table> => format!("{<>} {<>}")
};
```
and replace each with it's name individually. Modulo a more accurate pre-validate, I tried this out here https://github.com/lalrpop/lalrpop/compare/master...Pat-Lafon:lalrpop:inline_format . 

If this is something interesting, the validation step could be relaxed by trying to determine if the `{<>}` is inside of a string or not.

Otherwise, I lean towards 2? It's still just a nightly lint for now so not sure about getting users to change. We can replace our grammar examples with named bindings since that seems like a slightly less magical thing to encourage in the future.

dburgener:
Thanks for the investigation!

That is an interesting approach.  I think it seems like possibly the best option to handle that situation as far as I'm concerned.

I'll submit a PR for what I think is everything except the multiple `<>` case that you've got code for.  If you want to either submit that separately or on top of my PR I'm good either way.

dburgener:
Looking at #1061 just alerted me to the fact that `<>.0` is also apparently supported.  So that would be a different option for the "multiple `<>`" problem, which might be more readable?

dburgener:
@Pat-Lafon I reran nightly clippy now that #1063 merged and that reminded me that we still have a handful of cases for the "`<>` expands to multiple format arguments" situation.  Thoughts on the discussion above?  I feel like either your approach or the `<>.0` I mentioned in my last comment seem like reasonable options?  Any preference?

Pat-Lafon:
> > Looking at #1061 just alerted me to the fact that <>.0 is also apparently supported. So that would be a different option for the "multiple <>" problem, which might be more readable?
> 
> I tried this, and it doesn't work in format strings. I'm not sure the exact details of format string requirements, but in match_alternatives.lalrpop, I tried changing the format strings to both `{<>.0}` and `{(<>).1}`, and those expand to `{__0, __1.0}` and `{(__0, __1).0}`. The first is just obviously wrong, but either way both of them choke when compiling the generated code, because neither `,` or `(` is expected in a format string.
> 
> I think the approach you outlined above seems like the only way to go here. Are you planning to submit a PR for that?

Yeah, I'm not sure you can have any expressions inside of a format string expansion like that. I think I only see variable names used? It would need to be in the form I listed for inline above

PR #1059, opened 2025-05-01
Fix log typo

A4-Tacks:


dburgener:
I think the original is preferable.  Keeping the past tense in line with the past tense "completed" at the end makes sense, and "begin" makes this sort of a sentence fragment.

Issue #1060, opened 2025-05-02
Clippy warns about uninlined_format_args in the generate code


asomers:
Using lalrpop triggers `uninlined_format_args` warnings in the generated code.  It's a problem, because it's hard to suppress those warnings downstream.  Better to fix or suppress them here.

To reproduce, just run `cargo clippy --all-features --all-targets` in lalrpop's own repository.  A recent nightly compiler will produce warnings like this:
```
warning: variables can be used directly in the `format!` string
   --> /usr/home/somers/src/rust/lalrpop/target/debug/build/lalrpop-test-a3885c91ff55565d/out/zero_length_match.rs:546:26
    |
546 |                     _ => panic!("invalid reduction index {}", __reduce_index)
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
help: change this to
    |
546 -                     _ => panic!("invalid reduction index {}", __reduce_index)
546 +                     _ => panic!("invalid reduction index {__reduce_index}")
```

PR #1063, opened 2025-05-09
Fix uninlined format strings in generated code

dburgener:


dburgener:
Fixes #1060.

Since this can cause problems for users, and will grow in impact once this nightly clippy check hits stable, I'd like to backport this PR onto #1050 once merged and then go ahead and actually publish 0.22.2.

Pat-Lafon (approved):

PR #1065, opened 2025-05-17
Fixes: Enforcing Clone in parser for custom Location

### LINKED ISSUE #1064: should allow location that only implements `Clone`
Using 0.22.1.

It seems that lalrpop assumes `Copy` of `Location` but clearly this does not have to be the case. From the generated code:
```rust
        let __sym0 = __pop_Variant10(__symbols);
        let __start = __sym0.0;
        let __end = __sym0.2;
        let __nt = super::__action159::<>(__sym0);
``` 
Here `__sym0` is a triple, and `__sym0.0` and `.2` are my `Location` struct. It only implements `Clone`, but then this code moves both `Location`s away, so the fourth line issues a life time error. 

Proposed fix: lalrpop should check whether `Location` implements `Copy`. If not, then call `.clone()`.

Jaredanwolfgang:
After viewing the Issue #1064, it mentions that the MyLoc only implements Clone not Copy trait and suggests that `.clone()` should be added.  I added tests (see `lalrpop-test/src/loc_clone_issue1064.rs` and `lalrpop-test/src/loc_clone.lalrpop`) and made changes in `lalrpop/src/lr1/codegen/ascent.rs` and `lalrpop/src/lr1/codegen/parse_table.rs`. By applying these changes, I assume the `.clone()` issue might be solved.

### DIFF
--- a/lalrpop-test/src/lib.rs
+++ b/lalrpop-test/src/lib.rs
@@ -180,6 +180,9 @@ mod user_defined_error_visibility;
 
 lalrpop_mod_test!(zero_length_match);
 
+lalrpop_mod_test!(loc_clone);
+mod loc_clone_issue1064;
+
 /// This constant is here so that some of the generator parsers can
 /// refer to it in order to test `super::` handling in action code.
 const ZERO: i32 = 0;
--- a/lalrpop-test/src/loc_clone.lalrpop
+++ b/lalrpop-test/src/loc_clone.lalrpop
@@ -1,0 +1,31 @@
+use crate::loc_clone_issue1064::{Tok, MyLoc, ExprSymbol, Opcode};
+
+grammar<'input>;
+
+pub Expr: Box<ExprSymbol<'input>> = {
+    Expr "ExprOp" Factor => Box::new(ExprSymbol::Op(<>)),
+    Factor,
+};
+
+Factor: Box<ExprSymbol<'input>> = {
+    Factor "FactorOp" Term => Box::new(ExprSymbol::Op(<>)),
+    Term,
+};
+
+Term: Box<ExprSymbol<'input>> = {
+    "num" => Box::new(ExprSymbol::NumSymbol(<>)),
+    "(" <Expr> ")"
+};
+
+extern {
+    type Location = MyLoc;
+    type Error = ();
+
+    enum Tok<'input> {
+        "num" => Tok::NumSymbol(<&'input str>),
+        "FactorOp" => Tok::FactorOp(<Opcode>),
+        "ExprOp" => Tok::ExprOp(<Opcode>),
+        "(" => Tok::ParenOpen,
+        ")" => Tok::ParenClose,
+    }
+}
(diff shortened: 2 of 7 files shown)


dburgener (approved):
This seems really clean to me.  And the new test code is good to avoid future problems.  Thanks for putting this together!  I just triggered PR tests, so we'll see if those hit any issues, but from a review perspective, I think this looks fine to merge.

dburgener:
Looks like the linting failed.  Just a quick `cargo fmt` should address that.  The `cargo clippy` check runs after the `cargo fmt` one, so if you haven't already, you'll want to check clippy too.

Jaredanwolfgang:
Roger that, I will check the `cargo fmt` and `cargo clippy`.

Jaredanwolfgang:
I have checked the Fmt issue. By `cargo clippy`, it generates warning regarding `Clone on Copy` and `Unit Argument` issue. I try to solve it with local `rust!(self.out, "#[allow(clippy::clone_on_copy, clippy::unit_arg)]");` but not sure whether it is a good fix, but at least `cargo fmt` and `cargo clippy` should pass smoothly.

Pat-Lafon (approved):

Jaredanwolfgang:
Yeah, I assume that global allow might be a better option and have already force-pushed this commit. Thanks for the advice!

dburgener:
This looks good to me.  I think the global allow is a good starting place for now.  We can always revisit later if we want (we have narrowed the scope of a few of those in the past).  Certainly the `clone_on_copy` lint seems reasonable to have global.  I'm a little confused by the `unit_arg` lint.  Do we know what it's complaining about?  Just looking at the updated generated grammar, I don't see why it would hit that lint.

dburgener:
I grabbed the PR and ran clippy without the `allow(clippy::unit_arg)` myself:

```
warning: passing a unit value to a function
    --> /home/dburgener/lalrpop/target/debug/build/lalrpop-test-af0b9e61f9ef194a/out/nested.rs:1837:29
     |
1837 |                 let __end = __sym1.2.clone();
     |                             ^^^^^^^^^^^^^^^^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call and replace it with the unit literal `()`
     |
1837 ~                 let __end = {
1838 +                     __sym1.2;
1839 +                     ().clone()
1840 ~                 };
     |
```

I guess this is just a side effect of the fact that sometimes things like `__sym1.2` are actually `()`, so we're doing `().clone()`?

This all feels weird, but I still think silencing it seems like the right thing to do?

Jaredanwolfgang:
Yes this is what I get when I run the clippy. Therefore I use allow(clippy::unit_arg) to silence the warning.

Pat-Lafon:
I think the question is why/when is `__sym1.{0, 2}` just `()` and it seems to be when an `extern { enum Tok... }` is defined in the lalrpop grammar but a location type is left unmentioned. I figure Lalrpop defaults to `()` in this case. I tried this out and adding`type Location = usize;` to a bunch of test grammars was all that was needed. This seems like the kind of internal lalrpop details/changes that we want to suppress from showing up in clippy lints.

Issue #1061, opened 2025-05-04
Multiple variable pattern support


A4-Tacks:
```rust
grammar;

Or<A, B> = { A B, B A }
pub Foo: String = <x:Or<"a", "b">> => {
    let (a, b) = x;
    a.to_string() + b
};
```

Rewritten to:

```rust
grammar;

Or<A, B> = { A B, B A }
pub Foo: String = <(a, b):Or<"a", "b">> => {
    a.to_string() + b
};
```

And support such as `<(mut a, b):Bar>`, `<(a, (b, c)):Baz>`

chanbengz:
LALRPOP supports a shorthand `<>`. See https://lalrpop.github.io/lalrpop/tutorial/003_type_inference.html.

I would suggest using that

```rust
grammar;

Or<A, B> = { A B, B A }
pub Foo: String = Or<"a", "b"> => {
    format!("{}{}", <>.0, <>.1)
};
```

> And support such as `<(mut a, b):Bar>`, `<(a, (b, c)):Baz>`

I see your point, but IMO this is weird because `<>` simply solves the problem. The only problem is that it might not be so "readable".

A4-Tacks:
When containing deep nesting, it is very difficult to read without using patterns to untangle it

dburgener:
I think that having support for this added syntax would be a value-add, although as @chanbengz points out, there are existing ways to express this situation pretty well.

I would say that I'm certainly open to accepting PRs that add this syntax if someone wants to submit one.

chanbengz:
I can be working on it but I'm waiting for #1065 to merge to avoid conflicts, since there will be some grammar changes.

Pat-Lafon:
> I can be working on it but I'm waiting for [#1065](https://github.com/lalrpop/lalrpop/pull/1065) to merge to avoid conflicts, since there will be some grammar changes.

Ah, just as I was merging it

chanbengz:
> Ah, just as I was merging it

Thanks. BTW, can you come up with a case where there's a nested tuple in grammar? Because it's uncommon to me.

PR #1067, opened 2025-05-22
Support pattern of arguments in grammar

chanbengz:
Implements #1061. This is a breaking change and I think we should do a careful review because I'm not sure whether I did it right, but I wrote a test for it and other tests were passed. Syntax supports and code generation works as expected in the `arg_pattern` test. I created a new AST node `Tuple` to holds the tuples who contains `TupleItem` with variant `Name` and `Tuple` (nested). However, I might not check the semantics or something for `Tuple` as `SymbolKind::Name` does. In codegen, `Name` is replaced with a enum `ArgPattern` which is the same as `TupleItem` (for `Name` and the new `Tuple`), because I think we should distinguish them in `arg_patterns`. And there are some chores remain undone

- [ ] Ensure semantic check works well for `Tuple`
- [ ] Document for this new feature

And #1065 reminds me that some of the allow could be global, like `too_many_arguments`.


chanbengz:
> Thanks for looking at this!  I'll aim to give it a code review as soon as a get a chance, but that likely won't be until early next week.

No rushes, as it definitely needs a careful review.

> Just a note on the breaking change - it's good timing for that, since we have 1 definitely breaking change and 2 arguably breaking changes already merged, so the next release off of master (as opposed to our backport 0.22.x branch) will be breaking anyways.  So the more needed breaking changes we get in prior to that the better.

Glad to know that. And there could be more changes since I haven't checked whether `<(mut a, b):Foo>` works properly. But before that, I'm gonna document this feature (1st item on TODOs) at first.

doc/src/tutorial/003_type_inference.md
```diff
@@ -86,5 +86,18 @@ The `<>` expressions also works with struct constructors (like `Foo
 {...}` in examples above). This works out well if the names of your
 parsed values match the names of your struct fields.
 
+However, `<>` would be confusing if you're dealing with complicated
```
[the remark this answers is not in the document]
chanbengz: Aha, as I'm not a native speaker. Thanks for this tip.

lalrpop-test/src/arg_pattern.lalrpop
```diff
@@ -0,0 +1,15 @@
+grammar;
+
+Or<A, B> = { A B, B A }
+DeepOr<A, B, C> = {
+    A Or<B, C>,
+}
+
+pub Foo: String = {
```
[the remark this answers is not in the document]
chanbengz: Yea, why not?

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -208,6 +211,18 @@ Symbol0: Symbol = {
                     SymbolKind::Repeat(Box::new(RepeatSymbol { symbol: lhs, op }))),
 };
 
+Tuples: Tuple =
```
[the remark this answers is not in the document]
chanbengz: They have incompatible types... I will see if I can resolve that.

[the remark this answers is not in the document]
dburgener: Yes, all I'm talking about is ordering in this file.  I'd like this block to be inserted lower in the file (after the Symbol1 definition) so that the defintions of Symbol0 and Symbol1 stay closer together.

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -988,6 +1027,30 @@ impl Display for Name {
...
+impl Display for TupleItem {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
+        match *self {
+            TupleItem::Name(ref n) => Display::fmt(&n, fmt),
+            TupleItem::Tuple(ref t) => Display::fmt(&t, fmt),
+        }
+    }
+}
+
+impl Display for ArgPattern {
```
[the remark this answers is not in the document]
chanbengz: `ArgPattern` is truncated.

chanbengz:
Done. Please have a look. I put the check for the match of tuple and type inside `infer_type` right after all nonterminals are inferred. I guess it should be refactored a little bit but I have no idea where to put it right now. RFC...

chanbengz:
Fix #1071. Its weird to have an another PR fixing that...

dburgener:
> Fix https://github.com/lalrpop/lalrpop/issues/1071. Its weird to have an another PR fixing that...

Thanks for the fix!  Would you mind opening a separate PR? Since we typically squash-merge, that will help keep the git history cleaner.  And it will be extremely quick and easy to just merge that without waiting on another round of review on this.

Which, speaking of - I haven't forgotten about this PR.  I've just been super busy the last few weeks.  Hoping to get back around to another review on this later this week.

lalrpop/src/normalize/tyinfer/mod.rs
```diff
@@ -370,6 +390,34 @@ impl<'grammar> TypeInferencer<'grammar> {
             }
         }
     }
+
+    #[allow(clippy::only_used_in_recursion)]
```
dburgener, line 394: This confuses me... I'm looking at the code, and I don't see why this lint would fire.  Are you sure it's actually needed, and not inherited from a previous version of the PR?  If it is still needed, what is it actually complaining about?

lalrpop/src/normalize/tyinfer/mod.rs
```diff
@@ -147,9 +147,29 @@ impl<'grammar> TypeInferencer<'grammar> {
...
 
-        for id in ids {
-            self.nonterminal_type(&id)?;
-            debug_assert!(self.types.lookup_nonterminal_type(&id).is_some());
+        for id in &ids {
+            self.nonterminal_type(id)?;
+            debug_assert!(self.types.lookup_nonterminal_type(id).is_some());
+        }
+
+        for id in &ids {
```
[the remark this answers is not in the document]
chanbengz: Yeah I agree. I was thinking that we can only check the tuple after everything is inferred but actually we don't. But IMO it will makes `nonterminal_type()` a bit more complicated and I'd keep it as is.

chanbengz:
Thanks for looking it again. I guess we just need one final round to get it done.

lalrpop/src/normalize/tyinfer/mod.rs
chanbengz: I see that it's not a false positive. `self` is not used in `validate_tuple()` and it's just being passed around and around. And surprisingly, when I moved the snippet inside (see conversation above), compiler warns that `nonterminal_type()` takes a mutable self and so, `validate_tuple()` cannot be invoked. So I moved it outside, and it solves both errors.

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -243,6 +246,18 @@ SymbolKind1: SymbolKind = {
         SymbolKind::Error,
 };
 
+Tuples: Tuple =
+    <lo:@L> "(" <tuples:(<TupleItem> ",")+> <e:TupleItem> ")" <hi:@R> => {
+        Tuple::new(tuples.into_iter().chain(once(e)).collect())
+    };
```
Pat-Lafon, line 252: I was comparing this with https://doc.rust-lang.org/reference/types/tuple.html. While it may be a little silly, do you think there is a reason to not support a 1-arity tuple in the same style that rust does? This would be like `(a,)`

Pat-Lafon: Maybe a better reference would be https://doc.rust-lang.org/reference/patterns.html#tuple-patterns . Interesting to note that you can match on unit, I imagine this would only really be done for readability reasons

lalrpop-test/src/arg_pattern.lalrpop
```diff
@@ -0,0 +1,19 @@
...
+    <(a, b):Barm<"a", "b">> => {
+        a.to_string() + b
+    },
+    <(a, (b, c)):DeepBar<"A", "B", "C">> => {
+        a.to_string() + b + c
+    },
+    <(a,b):Bar> => a + &b
```
Pat-Lafon, line 18: One super simple addition to add is a `mut` somewhere here. Just as a sanity check

lalrpop/src/normalize/tyinfer/mod.rs
```diff
@@ -226,6 +226,23 @@ impl<'grammar> TypeInferencer<'grammar> {
             Ok(alternative_types.pop().unwrap())
         })?;
 
+        for alt in nt.alternatives {
+            let symbols = &alt.expr.symbols;
+            for (t, s) in symbols.iter().filter_map(Symbol::as_tuple) {
+                let ty = if let SymbolKind::Nonterminal(ref id) = s.kind {
+                    self.nonterminal_type(id).unwrap()
```
Pat-Lafon, line 233: Could this `.unwrap()` be a `?`? I assume it's this NormResult type

doc/src/tutorial/003_type_inference.md
```diff
@@ -88,5 +88,18 @@ The `<>` expressions also works with struct constructors (like `Foo
...
+However, `<>` may be confusing if you're dealing with complicated
+tuples. In that case, LALRPOP provides a pattern matching to unwrap
+the tuple. For example,
+
+```lalrpop
+Foo = { A B }
+Bar = { C Foo }
+pub FooBar: String = {
+    <(a, b):Foo> => format!("{}{}", a, b),
+    <(c, (a, b)):Bar> => format!("{}{}{}", c, a, b)
```
Pat-Lafon, line 100: nit: we can encourage people to inline their format arguments

lalrpop/src/parser/lrgrammar.lalrpop
Pat-Lafon: I see you mentioned having conflicts with `ExprSymbol`, I assume that was because the macro wouldn't have the trailing comma on the single element case

Pat-Lafon:
(I'm hopeful that most of my review comments are higher level and easy to address so we don't need too much more time to get this through)

lalrpop/src/parser/lrgrammar.lalrpop
chanbengz: Not silly at all. Yeah we could do that but it takes to refactor this gammar and resolve the comma conflict.

Adding more pattern is possible too, but in my mind, it's effort-taking because we have to parse that again in LALRPOP's way and format the elements back. Have no idea if it can be done another way.

doc/src/tutorial/003_type_inference.md
chanbengz: Aha, I'm not completely accustomed to 2024 edition. Thanks!

lalrpop/src/normalize/tyinfer/mod.rs
```diff
@@ -389,3 +406,27 @@ fn maybe_tuple(v: Vec<TypeRepr>) -> TypeRepr {
...
+                    span,
+                    "expected a tuple of length {}, but found a tuple of length {}",
+                    items.len(),
+                    tuple.tuples.len()
+                );
+            }
+
+            for (item, tuple_item) in items.iter().zip(&tuple.tuples) {
+                if let TupleItem::Tuple(t) = tuple_item {
+                    validate_tuple(span, t, item).unwrap();
```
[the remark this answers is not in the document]
chanbengz: It will. `validate_tuple()` checks for length matching at the beginning and recursively.

lalrpop/src/parser/lrgrammar.lalrpop
chanbengz: > Maybe a better reference would be https://doc.rust-lang.org/reference/patterns.html#tuple-patterns . Interesting to note that you can match on unit, I imagine this would only really be done for readability reasons

Are you suggesting that we should follow this grammar? If we'd extend to Rust's pattern matching, I wonder that why we don't just fall back to `match` expression inside action body? And if we actually planed to implement, I think we have to follow a similar practice than this one, which may require refactoring our grammar and parse tree.

lalrpop/src/normalize/tyinfer/mod.rs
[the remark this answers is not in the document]
Pat-Lafon: Panic-ing doesn't seem like a good thing when we could just propagate the error message instead?

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -130,7 +130,7 @@ AttributeInner: Attribute =
         Attribute { id_span: Span(lo, hi), id, arg: arg.unwrap_or_default() }
 };
 
-AttributeArg: (AttributeArg) = {
```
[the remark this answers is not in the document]
chanbengz: `AttributeArg: (AttributeArg)` will generate `-> (AttributeArg)` in return type and arguments type annotation of action function. This is fine because `(AttributeArg)` is the same as `AttributeArg`, but a unary pattern would enforce that `let (a, ) = (1, );` with trailing comma. So you can see that I tweak the format in `repr.rs` for that formatting, which cause this part crashed.

lalrpop/src/grammar/parse_tree.rs
```diff
@@ -1052,7 +1109,13 @@ impl Display for TypeParameter {
...
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
         match *self {
-            TypeRef::Tuple(ref types) => write!(fmt, "({})", Sep(", ", types)),
+            TypeRef::Tuple(ref types) => {
+                if types.len() == 1 {
+                    write!(fmt, "({}, )", types[0])
+                } else {
+                    write!(fmt, "({})", Sep(", ", types))
+                }
+            }
```
Pat-Lafon, line 1118: Does this touch any of the code you changed? Or are you just changing it because it's talking about tuples? You don't change this part of the grammar

lalrpop/src/parser/lrgrammar.lalrpop
```diff
@@ -243,6 +246,18 @@ SymbolKind1: SymbolKind = {
...
+TupleItem: ArgPattern = {
+    <m:"mut"?> <l:Id> =>
+        ArgPattern::Name(Name::new(m.is_some(), l)),
+
+    <t:Tuples> => ArgPattern::Tuple(t),
+};
+
 TypeRef: TypeRef = {
     "(" <Comma<TypeRef>> ")" =>
         TypeRef::Tuple(<>),
```
Pat-Lafon, line 263: Is this the key part? That here we have a comma separated list for the type which gets converted to a tuple? Here there is no trailing comma to disambiguate the 1-arity case form the other?

lalrpop/src/normalize/tyinfer/mod.rs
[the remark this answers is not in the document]
chanbengz: Sorry for the confusion, and `validate_tuple()` actually returns `NormError` instead of `panic!`. Is that propagation of error message that you expect?

```rust
                return_err!(
                    span,
                    "expected a tuple of length {}, but found a tuple of length {}",
                    items.len(),
                    tuple.tuples.len()
                );
```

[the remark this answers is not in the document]
Pat-Lafon: Or likewise what would happen if one of the elements of the tuple was ill-typed? Like (int, b, c) when it expects (string, b, c)?

lalrpop/src/parser/lrgrammar.lalrpop
chanbengz: Exactly. I'm wavering whether we should refactor here because Rust takes it with warning while my implementation thinks `(a)` is a tuple.

lalrpop/src/normalize/tyinfer/mod.rs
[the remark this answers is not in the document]
chanbengz: > Note that the code i highlighted here specifically calls unwrap. This was my point about "inner tuples". So for `(a, b, (c, d))` if it should have been `(a, b, (c,))` I'm trying to figure out if your code will panic. It will recurse on the inner tuple, call validate on it, return and error, but then unwrap the error right?

My mistake on the `unwarp`. I've fixed and added a test for it.

> Or likewise what would happen if one of the elements of the tuple was ill-typed? Like (int, b, c) when it expects (string, b, c)?

Ill-typed will be checked by compiler. Should we check that here?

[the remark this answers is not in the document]
Pat-Lafon: That's true, I don't fully remember if we do any validation like that on our end in the non-tuple cases. Maybe we can just leave that as a compiler error which is how I assume we currently handle this

Pat-Lafon:
> No it makes no change. I'm trying to make it consistent with other tuples.

Ok, sounds reasonable to me. I think it's worth elevating for @dburgener that there is the tuple type and then the tuple pattern. This pr is concerned with the latter, however there is a question of syntax and consistency between the two

The tuple type is a comma separated list, the tuple pattern right now is a 1 or more list of names which doesn't need to end in a comma in the >1 case. How much do we try to align the two?

Currently the breakage comes from adding a comma to the end of the tuple type which changes how we output the type from `(A)` to `(A,)` in the rust file. We currently still accept `(A)` in the grammar file which we handle internally as a tuple type but isn't really. This would change that.

There are other ways we could poke at this, maybe keeping the current typeref behavior but adding an explicit `(A,)` case at the grammar level. `(A)` would have to become something other than a TypeRef::Tuple internally then. We could keep Typeref behaviour as it is and accept some inconsistency(which maybe means we don't currently support 1-arity tuples at the type level? but we do support the unit type which we don't support in the pattern).

lalrpop/src/parser/lrgrammar.lalrpop
Pat-Lafon: > > Is there a reason not to? I think the only difference between that and what we have now is that this currently allows for trailing commas which rust only has in the 1-arity case.
> 
> Actually rust allows trailing commas in any tuples. The difference between tuple type and pattern is the [RestPattern](https://doc.rust-lang.org/reference/patterns.html#grammar-RestPattern) which we don't have.

Oh interesting, somehow I missed that and thought it was only for the single arity case.

> > I'm a little less sure about your comment about changing this to be more like pattern matching. Is that something that's more relevant with a wider variety of patterns? Right now I kinda think of this set of patterns like `let pattern = ...` where the pattern allows for destructuring.
> 
> Yeah it's a sugar for destructuring. What I mean is that a better way to implement this feature and more patterns is to take user's input of pattern without parsing and re-formatting inside like user action code (of course ideally). Otherwise, we'll have to re-implement Rust's grammar (which is a huge workload). The cons is that validation is disabled and that we may not able to parse the patterns as a whole.

Ah ok. That could be interesting in that it would be more extensible I guess? I'm not sure that we would want to do this as part of this pr since you've gotten this to a good spot and that's an internal detail we can change/improve in a different pr. Would be very interested if this is something you want to work on. I would hope in that case we can still have some validation. I understand we can always just get compiler errors at the end of the day but it makes me a little sad because the error messages point to the generated code, not the grammar file so hard for users to figure out what the actual problem is.

PR #1068, opened 2025-05-22
Forward port 0.22.2 release notes to main branch

dburgener:
One note is that this also brings forward the version numbers across the board to 0.22.2 instead of 0.22.1.  I think that's fine, since 0.22.2 is a strict subset of master right now.  Just highlighting it, since it's a little weird given that 0.22.2 exists on the 0.22.x branch, but not the master branch.


Pat-Lafon (approved):

Pat-Lafon:
Yeah, or alternatively, master should be a strict superset of the backported branch.

PR #1069, opened 2025-05-28
Enable running tests with cargo-nextest

dburgener:
* Clean up CUSTOM_TEST_DIR appropriately after failures
* Set up nextest config to run api tests in serial

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,6 @@
+[test-groups]
+api = { max-threads = 1 }
+
+[[profile.default.overrides]]
+filter = 'test(api::)'
+test-group = 'api'
--- a/lalrpop/src/api/test.rs
+++ b/lalrpop/src/api/test.rs
@@ -66,6 +66,10 @@ fn remove_local_generated_files() {
             }
         }
     }
+    let custom_dir = temp_dir().join(CUSTOM_TEST_DIR);
+    if fs::exists(&custom_dir).unwrap() {
+        fs::remove_dir_all(&custom_dir).unwrap();
+    }
 }
 
 // This is maybe a little nonintuitive at first.  We verify that the file exists where we expect
@@ -132,9 +136,6 @@ fn test_explicit_in_out() {
     let (orig_dir, _lock) = setup();
 
     let custom_dir = temp_dir().join(CUSTOM_TEST_DIR);
-    if fs::exists(&custom_dir).unwrap() {
-        fs::remove_dir_all(&custom_dir).unwrap();
-    }
     fs::create_dir(&custom_dir).unwrap();
 
     Configuration::new()


Pat-Lafon:
As you set up nextest, did you notice a speedup or some other interesting benefit?

dburgener:
> As you set up nextest, did you notice a speedup or some other interesting benefit?

time cargo test:
real    0m2.500s
user    0m2.901s
sys     0m0.476s

time cargo nextest run:
real    0m1.154s
user    0m4.072s
sys     0m1.847s

So about twice as fast wall clock time (these are both after building everything, so should be just time to actually run the tests).

The nextest output also looks subjectively a little nicer to me?

I think the biggest benefit would just be all the configurability - things like running the API test in serial would have been a lot easier in nextest.  I left the mutex in for this PR, so that `cargo test` would continue to work, but the mutex is useless for nextest, since they are running in separate processes, so the mutex isn't actually shared.  The test group setup does the mutual exclusion there, and it's pretty trivial.  I don't have specific other use cases in mind, and this PR doesn't make much use of anything, but all the configuration and UI knobs seem like they could in theory be helpful.

Pat-Lafon:
I've tried cargo-nextest out and it's pretty slick. I'm a little sad I can't find a way to force cargo-nextest on `cargo test` invocation but that isn't too big of an issue. The output is a bit nicer which I appreciate. I saw something about nextest not supporting doc tests but I don't think we have any. In general, our ci testing times are probably more dependent on compilation than testing itself but the speed improvements are neat regardless.

The main reason to do this is still to enforce environment variable isolation.

dburgener:
I updated this to remove the interdependency with the `in_dir` code (and that fix is now part of #1082, where it belongs).  I also added a note in our CONTRIBUTING doc to clarify that both test runners are supported for now, but we are considering switching to exclusive `cargo-nextest` which is the plan I suggested [here](https://github.com/lalrpop/lalrpop/pull/1003#issuecomment-2931216331).

Pat-Lafon (approved):
LGTM

dburgener:
@Pat-Lafon just FYI, another issue I just noticed is that with regular cargo test, if one of the api tests fails, that means that it panics at that point and never calls the `teardown()` function.  That can cause other tests to failure because the state wasn't properly cleaned up (I was seeing it not change directory back to the original cwd).  Running these tests in `cargo nextest` seems to avoid that problem entirely.

I wouldn't be surprised if this is fixable in regular `cargo test`, but since we're talking about maybe mandating nextest anyways, we might not bother and consider it another point in favor of nextest.

Pat-Lafon:
That is a point in favor for sure. Would the suggestion here be to create wrap the Lock/State in a struct that calls teardown when it is dropped? I guess in the nextest case it doesn't matter as much because of the isolation and that each test creates a clean slate when it starts

PR #1079, opened 2025-07-13
Revert "Revert "Stop using in_dir once we get to process_* family functions (…"

Pat-Lafon:
Reverts lalrpop/lalrpop#1077

To have this as an open pr so it isn't lost. @dburgener can close this in favor of his own pr.

Either way, the main change that needs to be made is that the generated lalrpop file should be in the correct location.

### DIFF
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -5,8 +5,12 @@ use std::default::Default;
 use std::env;
 use std::env::current_dir;
 use std::error::Error;
+use std::io;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
+
+#[cfg(test)]
+mod test;
 
 /// Configure various aspects of how LALRPOP works.
 /// Intended for use within a `build.rs` script.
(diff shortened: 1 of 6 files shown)


Pat-Lafon:
Note in the new edition `set_var` should be marked as unsafe. I'm rebasing the edition changes on the current master and would like this pr to be merged after the edition update.

The following is what I wrote as the safety description for our use of `set_var`. Feel free to change as needed.

```rust
/// Safety
///
/// Warning: This function contains unsafe code to modify the testing
/// environment(setting environment variables).
///
/// This is used to properly simulate different OUT_DIR configurations. In
/// general the modification of environment variables is not thread-safe. The
/// setup of these tests is guarded by a mutex to ensure that no two tests
/// concurrently attempt to access the environement.
///
/// `cargo-nextest` is a test-per-process test runner which ensures that two
/// tests cannot interfere with each other and is thus the recommended way to
/// execute these api tests.
```

dburgener:
Closing in favor of #1082.  Thanks!

PR #1080, opened 2025-07-14
Make variable name clearer with anon symbols

dburgener:
The previous variable name really only refered to the first field of the tuple.  As a recent commit started using the second field of the tuple, this has become confusing.  Name the variable to be more clear.


Pat-Lafon (approved):
Works for me!

PR #1082, opened 2025-07-14
Reintroduce in dir fix

dburgener:

### DIFF
--- a/lalrpop/src/api/test.rs
+++ b/lalrpop/src/api/test.rs
@@ -1,0 +1,168 @@
+use std::env::{current_dir, set_current_dir, set_var, temp_dir};
+use std::fs;
+use std::path;
+use std::sync::{LockResult, Mutex, MutexGuard};
+
+use super::*;
+
+// tests may be run in parallel and these tests affect global state, so lock
+static API_TEST_MUTEX: Mutex<i32> = Mutex::new(0);
+
+const TEST_DIR: &str = "lalrpop-test";
+const CUSTOM_TEST_DIR: &str = "lalrpop-test2";
+
+#[derive(Debug, PartialEq)]
+enum GenFileLoc {
+    Src,
+    Other,
+    Root,
+    OutDir,
+    CustomOut,
+    DoesntExist,
+}
+
+// Set up for API tests.  The directory structure in test_files
+// is:
+//
+// outer.lalrpop
+// other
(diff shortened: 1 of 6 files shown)


Pat-Lafon:
We might not need it, but I had the suggestion of implementing a wrapper struct that calls teardown on drop. The goal would be to unsure that teardown is still called on panic but might not matter with nextest test isolation.

lalrpop/src/api/test.rs
```diff
@@ -0,0 +1,215 @@
...
+    OutDir,
+    OutDirSlashOther,
+    CustomOut,
+    DoesntExist,
+}
+
+// This holds state during the test to be cleaned up on drop
+struct TestState {
+    orig_dir: path::PathBuf,
+    lock: LockResult<MutexGuard<'static, i32>>,
```
Pat-Lafon, line 28: Maybe this is a silly question since i don't work with Mutex's often but why do we hold the LockResult and not the MutexGuard here?

Pat-Lafon:
>     1. I took the entirety of your patch at [Path sep in lalrpop_mod #1055 (comment)](https://github.com/lalrpop/lalrpop/pull/1055#issuecomment-2900061189) and put it in here as is.  That seemed to be working well to me, and just needed cleanup in other places.  I agree that what you wrote there is the right approach to fix this.  I credited you in the commit message with a Co-authored-by tag.  Let me know if that approach is okay with you.

Sure, works for me 👍 

>     2. I did change update_lrgrammar.sh.  I feel like the way it was working is just wrong.  I get that we want to reproduce the directory structure when we search in a path.  But if we're taking a specific file, we didn't search down some path.  I think putting that file at the out_dir root makes sense to me.  So I updated the script to call it that way, which seems to work in my testing.

Gotcha. I should double check this test as well when this pr is ready to go but that sounds reasonable to me.

lalrpop/src/api/test.rs
dburgener: No, good question.  The direct answer is "I don't really remember from when I did earlier rounds of this months ago".

But at the end of the day, with the new drop() implementation for teardown, it shouldn't matter much.  We return an error if a different test failed.  That would matter if they didn't have cleaned up state, but they should always clean up state on drop() now.  I added some handling for defense in depth.

PR #1083, opened 2025-07-15
Correctly label lalrpop code blocks in the lexer tutorial

dburgener:
Blocks mislabeled as rust code could cause user confusion.


Pat-Lafon:
Ah I see, that's fair. Partially accurate syntax highlighting is probably not worth the downsides you mention here.

PR #1088, opened 2025-08-21
Note the traits that custom lexer locations must implement

ISSOtm:
This caught me off-guard, and I would have preferred knowing about it ahead of time so I could design my `Location` type accordingly. So, add a note.

### DIFF
--- a/doc/src/lexer_tutorial/003_writing_custom_lexer.md
+++ b/doc/src/lexer_tutorial/003_writing_custom_lexer.md
@@ -50,7 +50,9 @@ pub type Spanned<Tok, Loc, Error> = Result<(Loc, Tok, Loc), Error>;
 ```
 
 `Loc` is typically just a `usize`, representing a byte offset into the input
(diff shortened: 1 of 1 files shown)


ISSOtm:
Wait. It looks like the `start_location` function, which requires `Default`, is unused?

dburgener:
Good call outs.  Digging some more on the semantics of PartialEq, it turns out that that requirement is only introduced by Lalrpop test code.  It's not actually needed for user lexer implementations - it's just needed for how we verify our custom lexer implementations in our tests (and technically workaroundable there).

So I think we can drop the mention of PartialEq - that's up to the end user.  In the "runs as a lalrpop test case" usecase, we need it, but other users may not.

Yes, Default should be the initial parsing location.  I would stay away from terms like "beginning of file" though, just since with a custom lexer who knows what you're actually parsing.  It may not be an actual file.  You might define "beginning" differently in your use case.

dburgener:
Merged.  Thanks for the contribution!

PR #1099, opened 2025-10-27
Update string_cache dependency to 0.9

dburgener:
It's not clear what changed in this version of string_cache, but anything semver breaking about it does not appear to affect us.


Pat-Lafon (approved):
Thanks. I saw in the lockfile that there was a duplicate dependency caused by petgraph. I've opened a pr: https://github.com/petgraph/petgraph/pull/915 and it's otherwise out of our hands.

PR #1101, opened 2025-10-31
Reproducer of breaking example from benchmark pr

Pat-Lafon:
I was poking around #981 and noticed an odd panic which I sourced to https://github.com/lalrpop/lalrpop/blob/dd7196feb26cfa494438f2943bd3ee9ee2dbb9ad/lalrpop-test/benches/src/compile_benches.rs . I'm starting this pr off with a simple reproducer. I'm not sure of the clean solution yet.

Note the error is that the file path is : `../lalrpop/src/parser/lrgrammar.lalrpop` but the in_dir is `src`(which in this case is wrong, in the reproducer is actually correct, but then a small prefix is added which breaks the assumption that in_dir is a prefix of the file path). This leads to an unwrap on None panic.

### DIFF
--- a/lalrpop/src/api/test.rs
+++ b/lalrpop/src/api/test.rs
@@ -192,6 +192,20 @@ fn test_process_src() {
 }
 
 #[test]
+fn test_process_file() {
+    let _state = setup();
+
+    // This test is noting that with cargo_dir_conventions, "src/src.lalrpop"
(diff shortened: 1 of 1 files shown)


dburgener:
See #1103.  I grabbed the test case from here to include in there to demonstrate how it's addressed.  Note that as we agreed above, it now returns `Err` instead of panicing.

PR #1124, opened 2026-04-16
Clippy lint 1.95

Pat-Lafon:
I updated rand to 0.10.1 based on https://github.com/rust-random/rand/releases/tag/0.10.1 and a security warning from dependabot(though we don't use the log feature so I don't think this is an issue for lalrpop?)

### DIFF
--- a/lalrpop/src/tok/test.rs
+++ b/lalrpop/src/tok/test.rs
@@ -15,7 +15,7 @@ fn gen_test(input: &str, expected: Vec<(&str, Expectation<'_>)>) {
 
     let tokenizer = Tokenizer::new(&input, 0);
     let len = expected.len();
-    for (token, (expected_span, expectation)) in tokenizer.zip(expected.into_iter()) {
+    for (token, (expected_span, expectation)) in tokenizer.zip(expected) {
         let expected_start = expected_span.find('~').unwrap();
         let expected_end = expected_span.rfind('~').unwrap() + 1;
         println!("token: {token:?}");


dburgener:
I don't understand the motivation here.  Is it just about satisfying dependabot?  0.10.1 is semver compatible with 0.10, so in practice we'll nearly always get the new version anyways.  And the issue in the PR says it requires `log` and `thread_rng` features to be enabled, while we enable neither, so the issue doesn't seem to apply to us.

lalrpop/src/lr1/error/mod.rs
dburgener, line 534: It looks like your branch is a few commits behind master.  These two were already added in https://github.com/lalrpop/lalrpop/commit/5a436469eb6e1209df58de42caf4c7d758ffef9e

Pat-Lafon:
> I don't understand the motivation here. Is it just about satisfying dependabot? 0.10.1 is semver compatible with 0.10, so in practice we'll nearly always get the new version anyways. And the issue in the PR says it requires `log` and `thread_rng` features to be enabled, while we enable neither, so the issue doesn't seem to apply to us.

Sure, in the same way that Clippy/Rustformat bring new lints to the code base that usually take very minimal effort to fix(though probably don't have an obvious benefit for any given lint fix). Since this is just a patch release, I don't think there is any downside to setting the cargo.lock to the new release.

If you would prefer, I won't be bothered if we just close this PR and dismiss the alert.

lalrpop/src/lr1/error/mod.rs
Pat-Lafon: My bad, I hadn't fetched the remote before checking out remote/master. It does seem that https://github.com/lalrpop/lalrpop/blob/f67f8741aeb5b78b0d4dc884ee4e6c97b3ac03ed/lalrpop/src/tok/test.rs#L18 is still linted by clippy

dburgener: Yeah, the other one makes sense.  I was surprised to see clippy lints, because I had been running locally.  The first one is explained by the branch delta, but the second one is real and explained by my consistent forgetting to run with `--tests`.  We should definitely take that one.

dburgener:
My preference is to drop the change and dismiss the alert.  It's not a big deal, but in the unlikely event that someone wants to build lalrpop in an offline environment without 0.10.1 available, I don't see any reason to break them over this (if there was a real correctness or security issue, sure, but this doesn't seem to be that).

Pat-Lafon:
I've removed that commit

dburgener (approved):

Issue #94, opened 2016-02-25
Automatically generate parse tree


nikomatsakis:
It'd be nice if one could easily generate variants of the parser which automatically generate a parse tree. This would avoid the need to write action code for demonstrations and tests, but also maybe some use cases in practice. It would also be nice to have a special mode for result type `()` -- i.e., for just checking whether parsing is *possible*.

nixpulvis:
This would be awesome if `()` had the meaning of building a parse tree for debugging. I currently have a convention when I'm playing around of just making a string, and wrapping everything in parens to see what things look like but a much better job could be done than that.

mattbettcher:
Firstly, excellence work! After much toying around and learning how Lalrpop works, I've started using a single enum to represent all the types that are parsed. This is leading to a vary nice tree that's easy to read and test with. I suspect it could be auto generated by someone of your skill. Let me know if you want to take a look and I'll send you the code via email. Not ready for public release.

ahmedcharles:
I assume that parse tree here means an AST which implements `Debug` and could be used to verify the details of the result of the parse?

I'm mildly confused about `()` being a special mode, doesn't it already allow checking whether parsing is possible?

PR #1125, opened 2026-05-19
Expose Parse Tree for processed LALRPOP grammar files

Scanavium21:
# Expose Parse Tree (Grammar AST) Through the Public API

Exposes the AST for parsed `.lalrpop` files as part of the public API.

## Motivation

LALRPOP is primarily used as a parser generator — it consumes `.lalrpop` grammar
files and emits Rust source code. However, the parsed grammar AST (the
`parse_tree::Grammar`) contains all the structural information about the
grammar such as nonterminals, productions, alternatives, token definitions,
action code, attributes, etc. This information is valuable beyond code
generation.

**Use cases include:**

- **Programmatic grammar analysis** — A `build.rs` script or standalone tool can
  inspect the grammar's structure to generate additional artifacts (e.g.,
  `tmLanguage.json` syntax highlighting files for editors, documentation, type
  definitions for other languages, etc.).
- **Tooling integration** — Grammar authors can validate conventions, extract
  metadata, or generate cross-references by accessing the AST directly.
- **Downstream processing** — The raw parse tree can be consumed by other
  transformers or generators without having to re-parse the `.lalrpop` file
  themselves.

Previously, the internal `process_file_into` function
parsed the grammar, passed it through `normalize`, and
then discarded the `Grammar`. The public API methods on
`Configuration` and the free functions
(`process_root`, `process_src`) all returned
`Result<()>`, losing the AST.

### Core refactor: `build/mod.rs`

The internal processing pipeline in `build/mod.rs` was refactored
to return the `Grammar` value instead of discarding it:

| Function            | Before           | After                                   |
|---------------------|------------------|-----------------------------------------|
| `process_file_into` | `io::Result<()>` | `io::Result<pt::Grammar>`               |
| `process_file`      | `io::Result<()>` | `io::Result<Option<pt::Grammar>>`       |
| `process_dir`       | `io::Result<()>` | `io::Result<Map<PathBuf, pt::Grammar>>` |

**Specific changes:**

1. **`process_file_into`** — Now returns the parsed `pt::Grammar`. The
   `force_build` / `needs_rebuild` check was moved up to the caller
   (`process_file`), so `process_file_into` always parses and returns an AST
   if invoked.

2. **`process_file`** — Now returns `Option<pt::Grammar>`:
   - `Some(grammar)` when the file was actually processed (parsed + code
     generated).
   - `None` when the cached output (`.rs` file) was up-to-date and no
     processing was necessary.

   The `needs_rebuild` / `force_build` check is performed here; if the file is
   already up-to-date, the function returns `None` without parsing. This
   preserves the incremental build optimization: if a caller does not need the
   AST, they pay nothing extra.

3. **`process_dir`** — Now returns `Map<PathBuf, pt::Grammar>`.
   Each key is the full path of the processed
   `.lalrpop` file, and the value is its corresponding AST. Files that did not
   need rebuilding are omitted from the map (since `process_file` returned
   `None` for them).

4. **`process_dir` collects results** — Iterates over `process_file` results
   and only inserts entries where a grammar was returned.

5. **`parse_and_normalize_grammar`** was split into two distinct helpers:
   - **`parse_grammar`** — Parses raw text into `pt::Grammar`.
   - **`normalize_grammar`** — Takes ownership of a `pt::Grammar` and normalizes
     it into `r::Grammar`.

   This allows for the intermediary AST to be cloned for use elsewhere before being
   normalizes.

### Public API additions: `api/mod.rs`

New methods were added to `Configuration` and as free
functions. All existing method signatures remain **unchanged** for full backward
compatibility.

#### Private helpers

Aims to reduce duplicated code in existing and new functions.

- **`Configuration::get_in_dir(&self) -> &Path`** — Extracted from `process()`.
- **`Configuration::prepare_session<P>(&self, path: P) -> Result<Session, Box<dyn Error>>`** —
  Extracted from `process_dir`.
  
### `lib.rs`

Makes the `parse_tree` module public for use in libraries to consume the AST.

### `grammar/parse_tree.rs`

Added `#![allow(missing_docs)]` to suppress warning about missing docs on public types.
What an AST does should be pretty self-explanatory

#### New `Configuration` methods

| Method | Return Type | Description |
|---|---|---|
| `process_with_grammar(&self)` | `Result<Map<PathBuf, Grammar>>` | Like `process()`, returns all processed grammars keyed by path. |
| `process_current_dir_with_grammar(&self)` | `Result<Map<PathBuf, Grammar>>` | Like `process_current_dir()`, returns grammars keyed by path. |
| `process_dir_with_grammar<P>(&self, path)` | `Result<Map<PathBuf, Grammar>>` | Like `process_dir()`, returns grammars keyed by path. |
| `process_file_with_grammar<P>(&self, path)` | `Result<Option<Grammar>>` | Like `process_file()`, returns `None` if cached. |

#### New free functions

| Function | Return Type | Description |
|---|---|---|
| `process_root_with_grammar()` | `Result<Map<PathBuf, Grammar>>` | Free-function version of `process_current_dir_with_grammar`. |
| `process_src_with_grammar()` | `Result<Map<PathBuf, Grammar>>` | Free-function version of `process_with_grammar` with in_dir set to `./src`. |

### In a `build.rs` to generate a `tmLanguage.json`

```rust
// build.rs
use lalrpop::Configuration;
use std::env;
use std::fs;
use std::path::PathBuf;

fn main() {
    // Process all grammar files and get their ASTs
    let grammars = Configuration::new()
        .use_cargo_dir_conventions()
        .process_with_grammar()
        .unwrap();

    for (path, grammar) in &grammars {
        // Extract terminal/token information from the grammar
        let keywords: Vec<_> = grammar
            .items
            .iter()
            .filter_map(|item| item.as_match_token())
            .flat_map(|mt| &mt.contents)
            .flat_map(|mc| &mc.items)
            .filter_map(|mi| /* extract terminal literals */ None)
            .collect();

        // Generate a TextMate grammar from the parsed grammar
        let tm_language = generate_tmlanguage(&grammar);
        let out_path = PathBuf::from(env::var("OUT_DIR").unwrap())
            .join(path.file_stem().unwrap())
            .with_extension("tmLanguage.json");
        fs::write(&out_path, &tm_language).unwrap();
    }
}
```

### Selective processing with caching awareness

```rust
// Only process a single file, respecting incremental build
if let Some(grammar) = Configuration::new()
    .process_file_with_grammar("src/parser.lalrpop")
    .unwrap()
{
    // File was processed; grammar is fresh
    analyze_grammar(&grammar);
} else {
    // File was already up-to-date; no grammar returned
    println!("No rebuild necessary");
}
```

## Design Decisions

1. **Backward compatibility** — All existing public functions retain their exact
   signatures, return types, and behavior. No breaking changes to the public
   API.

2. **`Option<Grammar>` for `process_file`** — Since `process_file_into` only
   processes when needed, the API honestly reflects that no grammar is available
   when a rebuild was skipped. Callers who always need the grammar can use
   `process_file_with_grammar` and handle `None`, or call
   `force_build(true)` before processing.

3. **`Map<PathBuf, Grammar>` for directory methods** — Using the full
   `PathBuf` as key avoids ambiguity (multiple files may have the
   same filename stem in different subdirectories) and makes it trivial to map
   results back to input files.

5. **Split `parse_and_normalize_grammar`** — Separating parse from normalization
   gives the caller control over whether to normalize at all.

---

**Disclaimer:** This summary was written with the help of AI (LLM)
after the code was edited by hand. None of the contributed code has been
generated by AI.

### DIFF
--- a/lalrpop/src/api/mod.rs
+++ b/lalrpop/src/api/mod.rs
@@ -1,4 +1,6 @@
 use crate::build;
+use crate::collections::Map;
+use crate::grammar::parse_tree::Grammar;
 use crate::log::Level;
 use crate::session::{ColorConfig, Session};
 use std::default::Default;
@@ -211,10 +213,26 @@ impl Configuration {
         self.process_dir(self.get_in_dir())
     }
 
+    /// Process all files according to the `set_in_dir` and
+    /// `set_out_dir` configuration.
+    ///
+    /// Returns a map of processed grammar files with their corresponding AST.
+    pub fn process_with_grammar(&self) -> Result<Map<PathBuf, Grammar>, Box<dyn Error>> {
+        self.process_dir_with_grammar(self.get_in_dir())
+    }
+
     /// Process all files in the current directory, which -- unless you
     /// have changed it -- is typically the root of the crate being compiled.
     pub fn process_current_dir(&self) -> Result<(), Box<dyn Error>> {
         self.process_dir(current_dir()?)
+    }
+
+    /// Process all files in the current directory, which -- unless you
+    /// have changed it -- is typically the root of the crate being compiled.
+    ///
+    /// Returns a map of processed grammar files with their corresponding AST.
+    pub fn process_current_dir_with_grammar(&self) -> Result<Map<PathBuf, Grammar>, Box<dyn Error>> {
+        self.process_dir_with_grammar(current_dir()?)
     }
 
     // The user should only use set_in_dir() with process(), not process_*() functions which
@@ -268,6 +286,15 @@ impl Configuration {
         Ok(())
     }
 
+    /// Process all `.lalrpop` files in `path`.
+    ///
+    /// Returns a map of processed grammar files
+    /// with their corresponding AST.
+    pub fn process_dir_with_grammar<P: AsRef<Path>>(&self, path: P) -> Result<Map<PathBuf, Grammar>, Box<dyn Error>> {
+        let session = Rc::new(self.prepare_session(&path)?);
+        Ok(build::process_dir(session, path)?)
+    }
+
     /// Process the given `.lalrpop` file.
     pub fn process_file<P: AsRef<Path>>(&self, path: P) -> Result<(), Box<dyn Error>> {
         let session = Rc::new(self.session.clone());
@@ -276,6 +303,16 @@ impl Configuration {
         self.verify_no_in_dir_conflict(None)?;
         build::process_file(session, path)?;
         Ok(())
+    }
+
+    /// Process the given `.lalrpop` file.
+    ///
+    /// Returns the corresponding AST for
+    /// the given file if processed.
+    pub fn process_file_with_grammar<P: AsRef<Path>>(&self, path: P) -> Result<Option<Grammar>, Box<dyn Error>> {
+        let session = Rc::new(self.session.clone());
+        self.verify_no_in_dir_conflict(None)?;
+        Ok(build::process_file(session, path)?)
     }
 }
 
@@ -290,6 +327,14 @@ pub fn process_root() -> Result<(), Box<dyn Error>> {
     Configuration::new().process_current_dir()
 }
 
+/// See [`process_root`] for details.
+///
+/// Returns a map of processed grammar
+/// files with their corresponding AST.
+pub fn process_root_with_grammar() -> Result<Map<PathBuf, Grammar>, Box<dyn Error>> {
+    Configuration::new().process_current_dir_with_grammar()
+}
+
 /// Process all files in ./src.
 ///
 /// In many cargo projects which build only one crate, this is the normal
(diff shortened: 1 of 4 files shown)


Scanavium21:
And while it is possible to integrate the AST (like #94 suggests) or `tmLanguage.json` generation directly into LALRPOP, having a public API for users allows for custom behaviors and generation beyond what LALRPOP provides.

dburgener:
See substantial prior discussion on this topic linked in this PR: https://github.com/lalrpop/lalrpop/pull/1045

My main concern with this sort of approach is that we expand the potentially breaking surface.  Any changes to the parse tree would now become semver breaking if we merge this PR.

In the case of the parse tree specifically, that may not be a big deal.  But I think the various usecases listed in the linked PR are worth supporting, so I'm not wild about merging a partial solution that hits only a subset.  I'm not immediate clear on the impact of this refactor.  It does seem like the parse_tree should be sufficient for most use cases, and the lalrpop-internals approach in the linked PR has definite drawbacks too.

What would be helpful to me in thinking about this PR is if you could engage with the prior work.  Could the various usecases enumerated in the linked PR be handled with this approach?  I can dig into that more at some point, but no promises that it will be particularly soon.  If we can meet most/all of the usecases with just the parse tree exposed, that's probably better than the lalrpop-internals approach IMO, but we'll need to do the homework of diving back in to all those use cases, which I haven't looked at in over a year, so it's pretty stale in my mind.

Scanavium21:
I can understand your concern and agree, exposing more internal structures results in less freedom to change anything without breaking those that depend on it.
I should've looked into the Pull Request you mentioned before submitting this change, though thank you for the detailed response on a topic that is already being covered.
I will evaluate this change's impact and usefulness, as well as potential alternative solutions to include other use-cases with minimal fragility, and get back to this, or the mentioned Pull Request.

PR #1128, opened 2026-05-25
Fix remaining issues for #1064

Colecf:
https://github.com/lalrpop/lalrpop/pull/1065 attempted to fix issue #1064, but missed several cases. This fixes some remaining ones. I'm not 100% confident I got them all, but it is enough to make my project compile on the latest LALRPOP.

### DIFF
--- a/lalrpop-test/src/loc_clone.lalrpop
+++ b/lalrpop-test/src/loc_clone.lalrpop
@@ -7,6 +7,8 @@ pub Expr: Box<ExprSymbol<'input>> = {
     Factor,
 };
 
+pub MaybeExpr: Option<Box<ExprSymbol<'input>>> = Expr?;
+
 Factor: Box<ExprSymbol<'input>> = {
     Factor "FactorOp" Term => Box::new(ExprSymbol::Op(<>)),
     Term,
--- a/lalrpop-test/src/loc_clone_issue1064.rs
+++ b/lalrpop-test/src/loc_clone_issue1064.rs
@@ -143,7 +143,7 @@ impl Debug for Opcode {
     }
 }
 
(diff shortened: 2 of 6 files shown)


Colecf:
Fixed clippy issues.

dburgener:
Thanks for the PR!

As a data point, I ran our benchmarking tests on this, and this is what I got:

```
Build lalrpop grammar   time:   [58.043 ms 58.863 ms 59.698 ms]
                        change: [+1.4893% +3.2245% +5.0721%] (p = 0.00 < 0.05)
                        Performance has regressed.

Parse JSON              time:   [3.1714 ms 3.2146 ms 3.2597 ms]
                        change: [−5.3545% −3.1963% −1.1355%] (p = 0.00 < 0.05)
                        Performance has improved.

Parse JSON with Referenced Input
                        time:   [2.4398 ms 2.4748 ms 2.5115 ms]
                        change: [+1.2948% +2.9651% +4.6715%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
```

I'm not sure how much to trust that, since the Parse JSON performance time actually improved.  It could just be my system wasn't quiescent enough at test time.

From a performance perspective, I would think this really shouldn't matter despite adding a bunch of `clone()`s, since those objects are being cloned anyways, the question is just about whether we need to be explicit since they aren't marked copy.

My inclination is to say that this is fine and merge it, but I'd like to leave it open for a bit first and see if @Pat-Lafon has any thoughts to add.

dburgener (approved):

Pat-Lafon (approved):
This all looks reasonable to me, thanks!

Colecf:
Thanks for reviewing!

PR #1131, opened 2026-06-22
Bump lalrpop-util version in tutorial to match lalrpop version

speicherwerk:
In the previous commit f67f874, only the version of lalrpop and not lalrpop-util was bumped in the tutorial.

### DIFF
--- a/doc/src/tutorial/001_adding_lalrpop.md
+++ b/doc/src/tutorial/001_adding_lalrpop.md
@@ -34,7 +34,7 @@ edition = "2024"
 lalrpop = "0.23.1"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


yannham (approved):

dburgener:
In case anyone else is curious, there's no need for a fix to version.sh for this.  There was a previous bug that resulted in non-update in this case, which was fixed by commit 0c418095f47626ebe9d03ac7d0522b3308e1109d.  Since version.sh checks the previous version, once this case got out of date, it wasn't updated by future runs of version.sh.  I verified that after this PR, version.sh picks this line up for future updates, so it will get automatically updated going forward.

PR #1139, opened 2026-09-21
Rebuild grammars when the feature set changes

### LINKED ISSUE #1136: CLI: regenerate the grammar when compiling with diffferent features
When recompiling the grammar with a different set of features (`lalrpop --features foo,bar grammar.lalrpop`) I have noticed that it doesn't regenerate the parser if the grammar was unchanged. I guess it's only checking the sha3 hash in the generated parser.

I haven't tried with the lib, but perhaps it suffers from the same bug.

dburgener:
We store a hash and the lalrpop version in generated grammar files, so that if the .lalrpop input file and the lalrpop version are identical, we don't need to rebuild.  However, the expected output could still change if we change features.  Add a third line showing which features it was built with, triggering a rebuild if those have changed.

Closes #1136


lalrpop/src/build/mod.rs
```diff
@@ -225,6 +237,16 @@ fn needs_rebuild(lalrpop_file: &Path, rs_file: &Path) -> io::Result<bool> {
...
 
+fn features_to_string(features: &Option<BTreeSet<String>>) -> String {
+    Some("// ".to_string())
+        .into_iter()
+        .chain(itertools::Itertools::intersperse(
+            features.iter().flatten().cloned(),
+            ",".to_string(),
+        ))
+        .collect()
+}
```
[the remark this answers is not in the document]
dburgener: The justification here seems bogus.  `String` does implement `FromIterator<String>` (https://doc.rust-lang.org/std/string/struct.String.html#impl-FromIterator%3CString%3E-for-Box%3Cstr%3E), and the code does of course compile as evidenced by CI.

That said, the code suggestion does seem like a much simpler and cleaner option, so I'll go ahead and accept that.

Buried in its output it also claims that due to trimming no feature builds will regenerate unconditionally, which seems like the much more substantial finding to me.  I'll investigate that.

lalrpop/src/build/mod.rs
```diff
@@ -185,6 +189,7 @@ fn process_file_into(
             let mut output_file = fs::File::create(rs_file)?;
             writeln!(output_file, "{LALRPOP_VERSION_HEADER}")?;
             writeln!(output_file, "{}", hash_file(lalrpop_file)?)?;
+            writeln!(output_file, "{}", features_to_string(&session.features))?;
```
Pat-Lafon, line 192: As is shown in the compiled grammar, some/most grammars don't rely on features. Do you think we could make this an optional output/input(so first check that session.features is non-empty)?

Pat-Lafon: Hmm, maybe this wouldn't work because it would complicate the needs rebuild logic versus just being able to always assume there is a feature line to read in.

lalrpop/src/parser/lrgrammar.rs
```diff
@@ -1,5 +1,6 @@
 // auto-generated: "lalrpop 0.23.1"
 // sha3: 5e8ffc3aef6c16a84fdea60b91ebe93a2cba830360325338d8aebb2427f18259
+//
```
Pat-Lafon: Nit: To elide initial confusion, maybe we could follow the same style as above and do a `// features:` prefix?

lalrpop/src/build/test.rs
```diff
@@ -0,0 +1,62 @@
+use super::*;
+
+use std::io::Write;
+use tempfile::NamedTempFile;
```
Pat-Lafon, line 4: I remember before you used temp_dir though this looks cleaner to me so maybe it's worth the extra dependency.

https://github.com/lalrpop/lalrpop/blob/8ab6b4104333c9d6501b07e94b62b61cc0efd681/lalrpop/src/api/test.rs#L1

Pat-Lafon (approved):
Thanks! I remember hitting this issue once myself so its nice to see this fixed.

lalrpop/src/build/mod.rs
dburgener: I don't feel super strongly here, but it seems to me as though there's some value in being explicit about "this grammar had no features enabled" as opposed to "for some reason that line wasn't generated".  Since these generated files aren't intended for human consumption anyways, I'm not sure that it's all that big a deal to have the empty line in most cases?

Is the concern here somewhat helped by adding `features:` as per the below comment, since that makes the line seem less like a pointless blank comment?

lalrpop/src/build/test.rs
dburgener: My thought was mostly that "dev dependencies barely matter".  I'm fine either way, but tempfile is pretty straightforward for this use case.

lalrpop/src/build/mod.rs
dburgener: > Hmm, maybe this wouldn't work because it would complicate the needs rebuild logic versus just being able to always assume there is a feature line to read in.

The funny thing about how the backwards compatibility works is that we end up reading the next blank line, and it's always not equal, which happens to be exactly what we want.  But I think having the line there unambiguously and actually reading the line we intend seems like a good thing to me.

Pat-Lafon: Gotcha, that makes sense to me

